PR #1, opened 2019-01-05
No sles11

rjschwei:


usr/sbin/registercloudguest
```diff
@@ -321,143 +321,90 @@ if cfg.has_section('instance'):
...
+
+# Register the base product first
+base_registered = False
+failed_smts = []
+while not base_registered:
+    cmd = [register_cmd, '--url']
+    cmd.append('https://%s' % registration_target.get_FQDN())
+    if os.path.exists(instance_data_filepath):
+        cmd.append('--instance-data')
+        cmd.append(instance_data_filepath)
```
smarlowucf: Can do this in one line with an extend but not sure it's any more performant.

rjschwei:
Also included further simplification w.r.t. https://github.com/SUSE/Enceladus/pull/224

PR #2, opened 2019-01-07
- Re-factor code to make instance data collection re-usable

rjschwei:
+ This will be needed for use in zypper plugins to transmit instance
    data during zypper requests

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -338,6 +338,50 @@ def get_current_smt():
 
 
 # ----------------------------------------------------------------------------
+def get_instance_data(config):
+    """Run the configured instance data collection command and return
+       the result or none."""
+    instance_data = None
+    if (
+            config.has_section('instance') and
+            config.has_option('instance', 'dataProvider')
+    ):
+        instance_data_cmd = config.get('instance', 'dataProvider')
+        cmd = instance_data_cmd.split()[0]
+        if cmd != 'none':
+            if not cmd.startswith('/'):
+                try:
+                    p = subprocess.Popen(
+                        ['which %s' % cmd],
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE,
+                        close_fds=True
+                    )
+                except:
+                    errMsg = 'Could not find configured dataProvider: %s' % cmd
+                    logging.error(errMsg)
+            if os.access(cmd, os.X_OK):
+                try:
+                    p = subprocess.Popen(
+                        instance_data_cmd.split(),
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE,
+                        close_fds=True
+                    )
+                    instance_data, errors = p.communicate()
+                except:
+                    errMsg = 'Error collecting instance data with "%s"'
+                    logging.error(errMsg % instance_data_cmd)
+                if errors:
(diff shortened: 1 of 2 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -337,6 +337,50 @@ def get_current_smt():
...
+                    p = subprocess.Popen(
+                        instance_data_cmd.split(),
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE,
+                        close_fds=True
+                    )
+                    instance_data, errors = p.communicate()
+                except:
+                    errMsg = 'Error collecting instance data with "%s"'
+                    logging.error(errMsg % instance_data_cmd)
```
42B:
```python
logging.error(err_msg, instance_data_cmd)  # ?
```

lib/cloudregister/registerutils.py
```diff
@@ -337,6 +337,50 @@ def get_current_smt():
...
+                if errors:
+                    errMsg = 'Data collected from stderr for instance '
+                    errMsg += 'data collection "%s"' % errors
+                    logging.error(errMsg)
```
42B:
```python
logging.error(err_msg, errors)  # ?
```

usr/sbin/registercloudguest
```diff
@@ -298,28 +298,11 @@ utils.set_as_current_smt(registration_target)
...
-                os.system("%s > %s" % (instance_data_cmd,
-                                       instance_data_filepath))
-            else:
-                msg = 'Configured dataProvider "%s" is not executable' % cmd
-                logging.error(msg)
+instance_data = utils.get_instance_data(cfg)
+if instance_data:
+    inst_data_out = open(instance_data_filepath, 'w')
+    inst_data_out.write(instance_data)
+    inst_data_out.close()
```
42B:
```python
with open(instance_data_filepath, 'w') as f:
    f.write(instance_data)
```
?

smarlowucf: Ditto, good to use context manager.

lib/cloudregister/registerutils.py
```diff
@@ -337,6 +337,50 @@ def get_current_smt():
...
+       the result or none."""
+    instance_data = None
+    if (
+            config.has_section('instance') and
+            config.has_option('instance', 'dataProvider')
+    ):
+        instance_data_cmd = config.get('instance', 'dataProvider')
+        cmd = instance_data_cmd.split()[0]
+        if cmd != 'none':
+            if cmd[0] != '/':
```
smarlowucf: Given the comparisons, cmd is assumed to be a string. Suggest using startswith which I think makes this more clear when reading.

```
if not cmd.startswith('/')
```

lib/cloudregister/registerutils.py
```diff
@@ -337,6 +337,50 @@ def get_current_smt():
...
+                        stderr=subprocess.PIPE,
+                        close_fds=True
+                    )
+                except:
+                    errMsg = 'Could not find configured dataProvider: %s' % cmd
+                    logging.error(errMsg)
+            if os.access(cmd, os.X_OK):
+                try:
+                    p = subprocess.Popen(
+                        instance_data_cmd.split(),
```
smarlowucf: Suggest using shlex.split which will respect quotes and generally be less bug prone when splitting shell commands.

PR #4, opened 2019-01-24
- Support RMT as target

rjschwei:
+ Use health API for RMT to figure out if the server is responsive
  + Find the RMT cert with the new name rmt.crt


lib/cloudregister/smt.py
```diff
@@ -179,22 +193,27 @@ def __is_cert_valid(self, cert):
...
-            except Exception:
-                # No response from server
-                logging.error('=' * 20)
-                logging.error('Attempt %s of %s' % (attempts, retries))
-                logging.error('Server %s is unreachable' % ip)
-            if cert_rq and cert_rq.status_code == 200:
-                attempts = retries
-
-        return cert_rq
+            for cert_name in ('smt.crt', 'rmt.crt'):
```
ikapelyukhin: Number of retries is still 3, and this will make twice as many requests, half of which are guaranteed to fail. Increase to 6?

lib/cloudregister/smt.py
```diff
@@ -179,22 +193,27 @@ def __is_cert_valid(self, cert):
...
+                    # No response from server
+                    logging.error('=' * 20)
+                    logging.error(
+                        'Attempt %s with %s of %s' % (
+                            attempts, cert_name, retries)
+                    )
+                    logging.error('Server %s is unreachable' % ip)
+                if cert_res and cert_res.status_code == 200:
+                    attempts = retries
+                    break
```
ikapelyukhin: `break` should be enough to break out of the loop, `attempts = retries` shouldn't be needed anymore.

lib/cloudregister/smt.py
```diff
@@ -179,22 +193,27 @@ def __is_cert_valid(self, cert):
...
-                logging.error('=' * 20)
-                logging.error('Attempt %s of %s' % (attempts, retries))
-                logging.error('Server %s is unreachable' % ip)
-            if cert_rq and cert_rq.status_code == 200:
-                attempts = retries
-
-        return cert_rq
+            for cert_name in ('smt.crt', 'rmt.crt'):
+                try:
+                    ip = self.get_ipv4()
```
smarlowucf: Could be `ip = self.get_ipv4() or self.get_ipv6()`

PR #5, opened 2019-02-06
- Support RMT as target

rjschwei:
+ RMT delivers https repos, we need to handle this when checking for
    the repo format for service switching
  + RMT has multiple services, each will link to the repo service script,
    for service switching we only want one entry in the trigger file per
    service
  + When switching from SMT to RMT we need to write the new credentials
    files
- zypper expects a string in the header to be sent along
- spawning a process in Python 3.6 is very ugly, but the nice older ways
  are no longer supported P_DETACH is gone
- need to lock the service switch trigger file to avoid having additional
  information added while the service plugin is still in play


rjschwei:
> `zypper creates a weird date stamped service file` -- I couldn't reproduce this (with code that removes extra services commented out). Curious why this was happening, maybe this isn't happening anymore.

Well, we both saw it at least once, thus I am going to leave the code in place, just in case.

PR #7, opened 2019-05-24
Module order

rjschwei:
Register modules in the proper order to avoid failure


usr/sbin/registercloudguest
```diff
@@ -90,6 +83,43 @@ def get_products():
...
     return products
 
+# ----------------------------------------------------------------------------
+def register_modules(extensions, products, registered=[]):
+    """Register modules obeying dependencies"""
+    for extension in extensions:
+        arch = extension.get('arch')
+        identifier = extension.get('identifier')
+        version = extension.get('version')
+        triplet = '%s/%s/%s' % (identifier, version, arch)
```
smarlowucf: Could be `'/'.join([identifier, version, arch])`.

usr/sbin/registercloudguest
```diff
@@ -90,6 +83,43 @@ def get_products():
...
+                cmd.append('--instance-data')
+                cmd.append(instance_data_filepath)
+
+            logging.info('Registration: %s' % ' '.join(cmd))
+            p = subprocess.Popen(
+                cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE
+            )
+            res = p.communicate()
```
smarlowucf: Consider more verbose naming, e.g. `resp` or `response` and `proc` instead of `p`.

usr/sbin/registercloudguest
```diff
@@ -361,36 +392,40 @@ while not base_registered:
...
-        entry = item.decode()
-        if 'Error:' in entry:
-            logging.error('\tRegistration failed: %s' % entry)
+base_prod_xml = open('/etc/products.d/baseproduct').read()
+prod_def_start = base_prod_xml.index('<product')
+product_tree = etree.fromstring(base_prod_xml[prod_def_start:])
+prod_identifier = product_tree.find('name').text.lower()
+version = product_tree.find('version').text
+arch = product_tree.find('register').find('target').text.split('-')[-1]
+headers = {'Accept':'application/vnd.scc.suse.com.v4+json'}
```
smarlowucf: Space after `:` for flake8.

usr/sbin/registercloudguest
```diff
@@ -361,36 +392,40 @@ while not base_registered:
...
-        if 'Error:' in entry:
-            logging.error('\tRegistration failed: %s' % entry)
+base_prod_xml = open('/etc/products.d/baseproduct').read()
+prod_def_start = base_prod_xml.index('<product')
+product_tree = etree.fromstring(base_prod_xml[prod_def_start:])
+prod_identifier = product_tree.find('name').text.lower()
+version = product_tree.find('version').text
+arch = product_tree.find('register').find('target').text.split('-')[-1]
+headers = {'Accept': 'application/vnd.scc.suse.com.v4+json'}
+query_args = 'identifier=%s&version=%s&arch=%s&release_type=ga' % (
```
ikapelyukhin: In the DB `release_type` is `NULL` for all products after SLES10, setting the filter to `release_type=ga` will result in an empty response.

rjschwei:
> Whoever wrote that `A registration code is not required for authentication by RMT or SMT` in SCC API documentation lied -- a valid registration code is required (by RMT, at least):
> 
> * https://github.com/SUSE/rmt/blob/6ee70a806ae43e81e8c77ae2462435dd48f879c7/app/controllers/api/connect/v4/subscriptions/products_controller.rb#L1-L3
> * https://github.com/SUSE/rmt/blob/4ad0f596b893e754de01d6138635fe5e48a11cc1/app/controllers/api/connect/base_controller.rb#L26-L35
> 
> `registercloudguest` would probably to register the base product first, get system credentials and use those to access `/connect/systems/products` endpoint with those credentials instead.

Well I tested the implementation against SMT, Based on your comments I conclude that the implementation for SMT and RMT are different :(

Against SMT in EC2:
`
>>> r = requests.get('https://smt-ec2.susecloud.net//connect/subscriptions/products?identifier=sles&version=15&arch=x86_64&release=ga', headers={'Accept': 'application/vnd.scc.suse.com.v4+json'})
>>> r.status_code
200
>>> data = json.loads(r.text)
>>> extensions = data[0].get('extensions')
>>> len(extensions)
1
>>> deps=extensions[0].get('extensions')
>>> len(deps)
6
`
The base product is registered, so yes we can form a basic auth object and pass it along but for SMT that is not required as shown above.

rjschwei:
@ikapelyukhin switched the API, ready for another review. Thanks for the help.

lib/cloudregister/registerutils.py
```diff
@@ -998,6 +995,7 @@ def __remove_repos(smt_server_name):
 def __remove_service(smt_server_name):
     """Remove the services pointing to the update infrastructure"""
     service_files = glob.glob('/etc/zypp/services.d/*')
+    service_files += glob.glob('usr/lib/zypp/plugins/services/*')
```
ikapelyukhin: The path probably has to be absolute (missing leading slash).

usr/sbin/registercloudguest
```diff
@@ -361,36 +393,45 @@ while not base_registered:
...
```
ikapelyukhin: `release_type` isn't an allowed parameter for `/systems/products` endpoint, so it could be dropped (it's ignored by the backend).

usr/sbin/registercloudguest
```diff
@@ -361,36 +393,45 @@ while not base_registered:
...
-    for item in res:
-        entry = item.decode()
-        if 'Error:' in entry:
-            logging.error('\tRegistration failed: %s' % entry)
+base_prod_xml = open('/etc/products.d/baseproduct').read()
+prod_def_start = base_prod_xml.index('<product')
+product_tree = etree.fromstring(base_prod_xml[prod_def_start:])
+prod_identifier = product_tree.find('name').text.lower()
+version = product_tree.find('version').text
+arch = product_tree.find('register').find('target').text.split('-')[-1]
```
ikapelyukhin: Can we just use the `arch` element instead? `<arch>x86_64</arch>`

ikapelyukhin: It seems to be present on both SLE12 and SLE15.

PR #10, opened 2019-06-05
- Handle errors if metadata retrieval fails

rjschwei:
+ It is possible that metadata retrieval from an instance may fail. In
    GCE the identity information we need to send along may be blocked.
    Handle the errors by providing warning, but let the registration process
    proceed as there is a fallback on the registration server to verify
    the instance identity.


usr/sbin/cloudguest-repo-service
```diff
@@ -101,7 +101,11 @@ user, password = utils.get_credentials(
     utils.get_credentials_file(update_server)
 )
 auth_creds = HTTPBasicAuth(user, password)
-instance_data = bytes(utils.get_instance_data(utils.get_config()), 'utf-8')
+instance_data = None
```
smarlowucf: Inconsistent initialization of instance_data probably good to be `b''` for all instances since `if instance_data` resolves the same for None and empty string.

lib/cloudregister/registerutils.py
```diff
@@ -290,7 +290,11 @@ def get_activations():
 
     auth_creds = HTTPBasicAuth(user, password)
 
-    instance_data = bytes(get_instance_data(get_config()), 'utf-8')
+    instance_data = None
```
smarlowucf: Ditto.

usr/sbin/cloudguest-repo-service
rjschwei: The initialization matches with how the `instance_data` variable is used in the context of this reference. Python, not C.

smarlowucf: If there isn't an exception they all get initialized by `instance_data = bytes(utils.get_instance_data(utils.get_config()), 'utf-8')` which is bytes?

ikapelyukhin:
@rjschwei why does fetching metadata fail, though? Could it be that now we are poking the endpoint too often (due to it being fetched by URL resolver every time) and are hitting the rate limit? Can we use client-side cache to not fetch metadata on every request?

lib/cloudregister/registerutils.py
```diff
@@ -481,8 +486,7 @@ def get_instance_data(config):
     if instance_data:
         # Marker for the server to not return https:// formated
         # service and repo information
-        instance_data += b'<repoformat>plugin:susecloud</repoformat>\n'
-        return instance_data.decode()
+        instance_data += '<repoformat>plugin:susecloud</repoformat>\n'
```
ikapelyukhin: The client probably should always send this, even if the metadata is not available. Otherwise if the client can't fetch metadata, the server would fall back to HTTPS URLs, even if the the client is recent.

usr/sbin/cloudguest-repo-service
rjschwei: The get_instance_data() function returns a string.

rjschwei:
> @rjschwei why does fetching metadata fail, though? Could it be that now we are poking the endpoint too often (due to it being fetched by URL resolver every time) and are hitting the rate limit? Can we use client-side cache to not fetch metadata on every request?

Discovered by the the QA team. When an instance is started without a service account in GCE then the `guest-attributes` API is not accessible. Behind this API is the JWT that we send as instance data. Other parts of the metadata are still accessible in that condition. It is not a rate limit issue it is a "hard" permission issue. I created #11 to consider the caching of the metadata.

lib/cloudregister/registerutils.py
rjschwei: OK, that makes sens and simplifies things. That means this function always has a return value and all the other if conditions go away :) Also centralized the logging of the warning message. Great :)

PR #13, opened 2019-06-07
- Eliminate incorrect log messages about inconsistent state

rjschwei:
+ The URL Resolver runs when the system is first being registered.
    However, at this time various consistency checks fail as the system
    is not yet registered and thus a loop begins to try and find other
    "responsive" servers. Switch the is_registered() logic to use
    the service file instead of repos. Update the is_responsive() logic
    to be more streamlined and prefer the new API.
  + Version bump to 9.0, the has_repos() method in utils has been removed


usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -31,34 +31,36 @@ utils.set_proxy()
...
+        if not utils.is_new_registration():
+            update_server = utils.get_smt()
+            server_name = ''
+            if not update_server:
+                # Something went seriously wrong, however try the name from
+                # the hosts file
+                server_name = utils.get_update_server_name_from_hosts()
+            else:
+                server_name = update_server.get_FQDN()
+            if server_name:
```
smarlowucf: Can dedent this if block? and move `server_name = ''` below `repo_url = None`?

lib/cloudregister/registerutils.py
```diff
@@ -741,6 +758,12 @@ def set_proxy():
     return True
 
 
+# ----------------------------------------------------------------------------
+def set_new_registration_flag():
+    """Set a marker that this is the beginning of the registration process"""
+    Path(REGISTRATION_DATA_DIR + NEW_REGISTRATION_MARKER).touch()
```
ikapelyukhin: Hmm, but is it actually required?
The call stack before the fix looks something like this:

```
susecloud.RESOLVEURL():
  if utils.get_smt()
    if get_current_smt():
      registered = is_registered(): # <--- False due to has_repos check down below
        if check_registration():
          if has_repos(smt_server_name) and __has_credentials(smt_server_name):
            return True
      
      if registered:
        if is_responsive():
          return current_smt
        else find_equivalent_smt_server():
          return equivalent()
  else utils.get_update_server_name_from_hosts()
    logging.warning('The system is in an inconsistent state')
```

`utils.get_smt` is failing due to nested `has_repos` call, and then it enters the `get_update_server_name_from_hosts` branch that produces the warning.

However, you've replaced `has_repos` with `has_services` -- shouldn't it work properly with just that change?
IMO, having the `NEW_REGISTRATION_MARKER` makes thing more convoluted, it'd be great to avoid extra complexity if possible.

usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -31,34 +31,36 @@ utils.set_proxy()
...
-            except TypeError:
-                logging.warning(
-                    '[URL-Resolver] Unable to retrieve instance data'
-                )
-            # zypper wants a string or bad things happen on the other side
-            inst_data_bin_str = base64.b64encode(instance_data).decode('utf-8')
-        else:
-            repo_url = None
+        repo_url = None
+        if not utils.is_new_registration():
```
ikapelyukhin: Also, what happens with URL resolver when `utils.is_new_registration()` returns `True`?
Wouldn't that call `self.answer` with `repo_url = None` and no instance data?
What does zypper do when `None` is returned?

rjschwei: Yes, if `utils.is_new_registration()` is true then `self.answer` is none. However, based on observation, and what zypper appears to be doing, when `utils.is_new_registration()` there is nothing to resolve.

usr/lib/zypp/plugins/urlresolver/susecloud
rjschwei: Yes `server_name` could be initialized outside the `if not utils.is_new_registration()` condition. Then the check for `if server_name` could move outside the `if not utils.is_new_registration()` check.

However conceptually we want to resolve the URL only when it is not a "first_time" invocation of the resolver. Thus conceptually that would hen imply 

```
if utils.is_new_registration():
    self.answer(
            'RESOLVEDURL',
            {'X-Instance-Data': inst_data_bin_str},
            repo_url
        )
else:
    EVERYTHING_ELSE
```

We can of course argue about which is easier to read, but that just a mounts to a style choice with no "right" or "wrong" way.

Anyway, thanks for raising the question it made be realize that `inst_data_bin_str` is not properly initialized.

PR #16, opened 2019-07-31
- Handle credentials issues

rjschwei:
+ All credentials files should have equal content, this is verified
  + Verify credentials files have the expected entries
  + Check that username and password are recognized on the connected update
    server. If not the user is instructed to re-register


usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -48,6 +63,39 @@ class SUSECloudPlugin(Plugin):
...
+                        )
+                        if not user or not password:
+                            self.error(
+                                {'CREDENTIAL_ERROR': 'NOCREDENTIALS'},
+                                ('Credentials required for "%s" but not found '
+                                 'in "%s"' % (repo_url, credentials_file_path))
+                            )
+                            return
+                        # Verify access to the server withthe credentials
+                        auth_creds = HTTPBasicAuth(user, password)
```
ikapelyukhin: Move everything related to checking the credentials also into a separate method, perhaps?

PR #18, opened 2019-09-06
- Fix logic of handling new registrations

rjschwei:
+ When the new registration marker was set the url resolution was skipped
    completely. This caused issues during repo setup and caused repos to go
    missing. The new code skips credential and service verification during
    the initial registration process only and resolves the url from the
    entry in /etc/hosts without triggering a log message.


usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -34,81 +34,89 @@ utils.set_proxy()
...
+        if verify_credentials and not utils.credentials_files_are_equal():
+            self.error(
+                {'CREDENTIAL_ERROR': 'INCONSISTENT'},
+                'Not all credentials files are eqivalent'
+            )
+            return
+        server_name = ''
+        if utils.is_new_registration():
+            server_name = utils.get_update_server_name_from_hosts(True)
+        else:
```
smarlowucf: Suggestion: Since there's already an else clause `server_name = ''` could move here.

usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -34,81 +34,89 @@ utils.set_proxy()
...
+            except TypeError:
+                logging.warning(
+                    '[URL-Resolver] Unable to retrieve instance data'
+                )
+            # zypper wants a string or bad things happen on the other side
+            inst_data_bin_str = base64.b64encode(
+                instance_data).decode('utf-8')
+        else:
+            self.error(
+                {'NO_TARGET_ERROR': 'NOTFOUNTD'},
```
smarlowucf: s/NOTFOUNTD/NOTFOUND

Issue #20, opened 2019-09-22
SUSE Azure update failures


jjedwa165:
We recently upgraded our SUSE SLES 12SP4 azure VMs. Afterwards all additional upgrades failed with the message:

```
Problem retrieving files from 'SLE-Module-Adv-Systems-Management12-Updates'.
Not all credentials files are eqivalent
Problem retrieving files from 'SLE-Module-Containers12-Updates'.
Not all credentials files are eqivalent
```

After reversing the issue via strace we discovered that the the ``susecloud`` zypper plugin as a apart of the ``cloud-regionsrv-client`` is comparing all files in ``/etc/zypp/credentials.d/`` to a base file as a precondition. While this normally works, our VMs **also** have other private repos with credential files in the same default [zypper credential location](https://doc.opensuse.org/projects/libzypp/12.3/group__ZyppConfig.html). 

Since these (private) credential files are different than the base SUSE ones all calls to resolve the the SUSE repo URLs ( e.g ``=plugin:/susecloud?`` ) via the ```urlresolver``` python code fail and none of our repos can refresh anymore. 

To try illustrate this I have suggested PR fix for the ``registerutils`` module. I'm hoping you will consider this as a mitigation for multiple private repos with different credentials using zypper.

bear454:
Thanks for reporting; the primary maintainer is out on vacation; we'll review your contribution over the next few weeks.

PR #21, opened 2019-09-22
intial suggestion

jjedwa165:
This is a suggested fix to issue #20 

Since the plugin is run on every url resolution the check can be made only for the specific credential file corresponding to the repo being validated. This avoids accidental checks of other private repo credential files

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -136,15 +136,14 @@ def clear_new_registration_flag():
 
 
 # ----------------------------------------------------------------------------
-def credentials_files_are_equal():
-    """Compare all the credentials files and make sure they have the same
+def credentials_files_are_equal(in_credential):
+    """Compare the base credentials files the the repo header and make sure they have the same
        values."""
-    credentials_files = glob.glob('/etc/zypp/credentials.d/*')
-    if credentials_files:
-        base = credentials_files[0]
-        for credential_file in credentials_files[1:]:
-            if not filecmp.cmp(base, credential_file):
-                return False
+    credentials_location = '/etc/zypp/credentials.d/'
+    credentials_base     = credentials_location + 'SCCcredentials'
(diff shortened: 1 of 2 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -136,15 +136,14 @@ def clear_new_registration_flag():
 
 
 # ----------------------------------------------------------------------------
-def credentials_files_are_equal():
-    """Compare all the credentials files and make sure they have the same
+def credentials_files_are_equal(in_credential):
```
rjschwei: Please rename the argument to "repo_credentials"

usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -42,10 +42,10 @@ class SUSECloudPlugin(Plugin):
         if zypper_pid != prev_zypper_pid:
             verify_credentials = True
         # Note this logic breaks when FATE#320882/PM-1251 gets implemented
-        if verify_credentials and not utils.credentials_files_are_equal():
+        if verify_credentials and not utils.credentials_files_are_equal(headers.get('credentials')):
```
rjschwei: Create a temporary `repo_credentials = headers.get('credentials')` and then also re-use this cached value on lines 67 and 70 and remove line 65.

i.e.:
`repo_url += '?credentials=' + repo_cerdentials`
'/etc/zypp/credentials.d/%s' % repo_credentials

lib/cloudregister/registerutils.py
jjedwa165: done

usr/lib/zypp/plugins/urlresolver/susecloud
jjedwa165: done

jjedwa165:
I've tested the script on my Azure SUSE 12 SP4 VM, the issue looks resolved from my end.

PR #24, opened 2019-09-30
Include python3-six in dependencies (bsc#1152567)

bear454:
See [bsc#1152567](https://bugzilla.suse.com/show_bug.cgi?id=1152567)

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -33,6 +33,7 @@ Requires:       python3
 Requires:       python3-lxml
 Requires:       python3-M2Crypto
 Requires:       python3-requests
+Requires:       python3-six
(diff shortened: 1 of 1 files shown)


rjschwei:
This is generally incorrect. The code does not depend on six

```
-> find . -name *.py | grep six
-> 
```
The traceback in the referenced bugzilla entry clearly shows that the dependency on "six" is a dependency of urllib3, which is covered by (bsc#1150895) and was submitted with mr#201080. The accepted standard is that packages do not include indirect dependencies. I have other changes for the code pending will remove this incorrect requirement at that time.

bear454:
Don't remove it until that MR is released - which is currently not planned until 10/29 or we'll be broken again, _regardless of where the root is_.

Issue #23, opened 2019-09-23
Registration client should be aware of region changes.


bear454:
If an instances is registered, but that image moves to a new region, the client should trigger re-registration in order to connect to region-local updates.

PR #26, opened 2019-10-18
- Add --clean option to the command

### LINKED ISSUE #22: Need to add a cleanup option for deprovisioning 'golden images'
Tools like Azure Image Builder run an instance in order to allow customers to simply customize. We need a way for customers or tooling to simply clean up when deprovisioning. This will allow such customized regions to properly register for region-local updates.

rjschwei:
+ This option provides users that create images from running instances
    with an easy way to properly remove any data connected to on-demand
    instance registration
- Implement command line argument handling with argparse
  + This provides proper support for "--help"
- Add man page
- Remove improper RPM package dependency on Python six
- Closes #22


rjschwei:
Also closes #23

rjschwei:
> I would suggest adding removal of symlinks in `/usr/lib/zypp/plugins/services` to `utils.`remove_registration_data`.

@ikapelyukhin ???

`remove_registration_data()` contains a call to `__remove_service()`, which in turn uses `__get_service_plugins()` to find the symlinks and then remove them.

I do not understand your suggestion.

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,100 @@ def register_modules(extensions, products, registered=[]):
...
+    action='store_true',
+    dest='force_new_registration',
+    default=False,
+    help='Force a new registration, for excaptional cases only',
+)
+help_msg='The update server cert fingerprint. Use '
+help_msg+='in exceptional cases only'
+argparse.add_argument(
+    '--smt-fp',
+    action='store_true',
```
ikapelyukhin: `s/store_true/store/` or needs to be removed completely (`store` is the default action). Otherwise works as a boolean flag.

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,100 @@ def register_modules(extensions, products, registered=[]):
...
+    dest='user_smt_fp',
+    default=False,
+    help=help_msg
+)
+help_msg='Use a specific FQDN to connect to the update server, use '
+help_msg+='in exceptional cases only'
+argparse.add_argument(
+    '--smt-fqdn',
+    action='store_true',
```
ikapelyukhin: `s/store_true/store/`

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,100 @@ def register_modules(extensions, products, registered=[]):
...
+    dest='user_smt_fqdn',
+    default=False,
+    help=help_msg
+)
+help_msg='Use a specific IP to connect to the update server, use '
+help_msg+='in exceptional cases only'
+argparse.add_argument(
+    '--smt-ip',
+    action='store_true',
```
ikapelyukhin: `s/store_true/store/`

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,100 @@ def register_modules(extensions, products, registered=[]):
...
+if args.clean_up and args.force_new_registration:
+    msg = '--clean and --force-new are incompatible, use one or the other'
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+time.sleep(int(args.delay_time))
 
+if args.clean_up:
+    utils.remove_registration_data()
+    os.exit(0)
```
ikapelyukhin: `s/os.exit/sys.exit/`

rjschwei:
> @rjschwei ah, sorry -- I was thinking about the issue that was related to symlinks; I didn't realize symlink removal is invoked from `remove_registration_data()`. I've tested `--clean` and the symlinks are indeed removed.
> 
> `--force-new`, however, doesn't remove the symlinks -- perhaps it should clean them up as well? There shouldn't be any reason to keep them when re-registering.

Hmm. I do not see how the described behavior happens. When `--force-new` is used we still end up calling `remove_registration_data()` which cleans up the symlinks. But in order to get there the system is assumed to be registered. So if the user cleaned the SMT object cache manually then yes the symlinks would be left behind, but so would be other data. I am not certain we can deal with all the complexities in a nice way if we make the assumption the user put the system into an inconsistent state.

cloud-regionsrv-client.spec
```diff
@@ -110,6 +109,9 @@ cp -r usr %{buildroot}
 python3 setup.py install --prefix=%{_prefix}  --root=%{buildroot}
 mkdir -p %{buildroot}/var/lib/regionService/certs
 mkdir -p %{buildroot}/var/lib/cloudregister
+install -d -m 755 %{buildroot}/%{_mandir}/man1
+install -m 644 man/man1/* %{buildroot}/%{_mandir}/man1
+gzip %{buildroot}/%{_mandir}/man1/*
```
bear454: Is it redundant to call `gzip` on man contents in both the specfile and the makefile?

Re: https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/26/files#diff-b67911656ef5d18c4ae36cb6741b7965R22

man/man1/registercloudguest.1
```diff
@@ -0,0 +1,58 @@
...
+.\" groff -man -Tascii registercloudguest.1
+.\"
+.TH registercloudguest.1
+.SH NAME
+registercloudguest \- Register a SUSE Linux Enterprise on-demand instance
+.SH SYNOPSIS
+.B registercloudguest [option]
+.SH DESCRIPTION
+.B registercloudguest
+Register a SUSE Linux Enterprise on-demand instance with the region local
```
bear454: The infrastructure is _not always_ region local... especially in Azure where they have explicitly asked that we not put infrastructure in every region.

man/man1/registercloudguest.1
```diff
@@ -0,0 +1,58 @@
...
+.SH OPTIONS
+.IP "--clean"
+Clean the running instance of any data related to repository setup and
+instance registration. Use this option when creating new images from
+running instances before taking the snapshot of the root volume.
+.IP "-d --delay"
+Delay the process by the given number of seconds.
+.IP "-f --config-file"
+Specify the location of the configuration file. By default the configuration
+file "/etc/regionserverclnt.cfg" is expected and used
```
bear454: Needs a period at the end of the sentence, for consistency.

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,94 @@ def register_modules(extensions, products, registered=[]):
...
+    '-f', '--config-file',
+    dest='config_file',
+    help='Path to config file, default: /etc/regionserverclnt.cfg',
+)
+argparse.add_argument(
+    '--force-new',
+    action='store_true',
+    dest='force_new_registration',
+    default=False,
+    help='Force a new registration, for excaptional cases only',
```
bear454: s/excaptional/exceptional/

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,94 @@ def register_modules(extensions, products, registered=[]):
...
+)
+help_msg='The update server cert fingerprint. Use '
```
bear454: The `--smt-fp` and `--smt-fqdn` arguments use the help message suffix ", use ..."; this one uses ". Use ..."

ikapelyukhin:
> I am not certain we can deal with all the complexities in a nice way if we make the assumption the user put the system into an inconsistent state.

Just execute the code executed by `--clean` automatically when `--force-new` is invoked?

usr/sbin/registercloudguest
rjschwei: ????

rjschwei:
> > I am not certain we can deal with all the complexities in a nice way if we make the assumption the user put the system into an inconsistent state.
> 
> Just execute the code executed by `--clean` automatically when `--force-new` is invoked?

Well if the system is not registered --force-new is superfluous and we should not go down that code path. If the system is registered than `remove_registration_data()` is called.

usr/sbin/registercloudguest
```diff
@@ -128,58 +129,94 @@ def register_modules(extensions, products, registered=[]):
...
+
+if args.clean_up and args.force_new_registration:
+    msg = '--clean and --force-new are incompatible, use one or the other'
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+time.sleep(int(args.delay_time))
 
+if args.clean_up:
+    utils.remove_registration_data()
```
ikapelyukhin: 1. I would suggest calling `utils.clean_smt_cache()` here (if the intended use is to create a snapshot of an instance, then there's no point to bring along SMT cache), `--force-new` calls it too:

https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/a938dc18613c1f82f626c9f2376ecd6c686b9bfb/usr/sbin/registercloudguest#L238-L240

2. And maybe even throw in a call to `utils.clear_new_registration_flag()`? (Otherwise the marker would be cleaned up next time `registercloudguest` is called, but it still might be a good idea to clean up everything)

usr/sbin/registercloudguest
bear454: The help message for `--smt-fp` says "The update server cert fingerprint. Use in exceptional cases only"
The help message for `--smt-fqdn` says "Use a specific FQDN to connect to the update server, use in exceptional cases only"
The help message for `--smt-ip` says "Use a specific FQDN to connect to the update server, use in exceptional cases only"

I would like to see the same statement about exceptional use cases be used consistently in all cases. Either put a period in front of it, or a comma in front of it, in all cases, instead of mixing syntax.

rjschwei:
Changes requested by @ikapelyukhin implemented, merging

PR #31, opened 2020-04-20
- Handle hosts file as binary file

rjschwei:
+ In Python 3.4 we may be unable to read the /etc/hosts file if it contains
    non ascii characters. Modify the reading and handling of entries in
    /etc/hosts to use bytes.


lib/cloudregister/registerutils.py
```diff
@@ -109,19 +109,23 @@ def check_registration(smt_server_name):
 # ----------------------------------------------------------------------------
 def clean_hosts_file(domain_name):
     """Remove the smt server entry from the /etc/hosts file"""
+    if isinstance(domain_name, str):
```
smarlowucf, line 112: Suggest using a pythonic try/except here. If this script runs on python2 and python3 you would need to handle different types (basestring for python2 and str for python3).

lib/cloudregister/registerutils.py
```diff
@@ -109,19 +109,23 @@ def check_registration(smt_server_name):
...
+        if b'# Added by SMT' in entry:
             smt_announce_found = True
             continue
         if smt_announce_found and domain_name in entry:
             smt_announce_found = False
             continue
         new_hosts_content.append(entry)
 
-    with open(HOSTSFILE_PATH, 'w') as hosts_file:
+    with open(HOSTSFILE_PATH, 'wb') as hosts_file:
```
smarlowucf, line 128: Using a context manager here but the previous usage is not (line: 117). Therefore the file could still be open for reading and now be open a second time for writing. Suggest using a context manager every time the file is open to ensure it's closed when processing is finished.

lib/cloudregister/registerutils.py
```diff
@@ -613,11 +623,11 @@ def get_update_server_name_from_hosts(ignore_inconsistent=False):
                         'definitions, cached update server data, and '
                         'credentials file do not match')
     servers = get_available_smt_servers()
-    hosts_content = open(HOSTSFILE_PATH).read()
+    hosts_content = open(HOSTSFILE_PATH, 'rb').read()
```
smarlowucf, line 626: Ditto

lib/cloudregister/registerutils.py
rjschwei: We no longer support Python 2 with the registration client. So the concern with the multiple string types is somewhat academic. Never the less. The problem is that using `.encode()` generates a very generic "AttributeError" if domain_name is not a string. Thus, what would be caught in a "try - except" implementation a very generic error would be caught and that IMHO is not very intuitive in this case. Since `[ ].encode` produces the same as `b' '.encode()`.  I decided that showing the intent that domain_name should be a str, or could by bytes, would be more explicit in this case.

lib/cloudregister/registerutils.py
rjschwei: Not the same. Python constructs a temporary file handle. The temporary goes out of scope at the end of the function and that will close the file.

lib/cloudregister/registerutils.py
smarlowucf: Fair, my assumption if I see something like:

```
try:
    something = something.encode()
except AttributeError:
    pass
```

Is the same though. That the code is handling both bytes like objects and string like objects but in a duck typed way rather than the strict typed way.

lib/cloudregister/registerutils.py
smarlowucf: Ack, but garbage collection in Python happens arbitrarily and not exactly when all references are removed. Chances are that it never causes an issue but the suggestion to always use context manager guarantees that the very slight chance of a confusing error can't happen.

lib/cloudregister/registerutils.py
```diff
@@ -109,19 +109,24 @@ def check_registration(smt_server_name):
 # ----------------------------------------------------------------------------
 def clean_hosts_file(domain_name):
     """Remove the smt server entry from the /etc/hosts file"""
+    if isinstance(domain_name, str):
+        domain_name = domain_name.encode()
     new_hosts_content = []
-    content = open(HOSTSFILE_PATH, 'r').readlines()
+    # Handle entries as bytes,
+    # Yes, users put non ascii characters into /etc/hosts
```
ikapelyukhin, line 116: Well, its not entirely unreasonable to expect unicode strings in 2020 :-)

PR #32, opened 2020-05-18
- Loop over region server access

rjschwei:
+ In the cloud framework it possible to have the network interface
    recognized as being online but external routing may not yet be working.
    This triggers a race condition between accessing the region servers
    and the framework routing configuration. Retrying the connections
    to the region servers allows the framework routing setup to win.


lib/cloudregister/registerutils.py
```diff
@@ -240,35 +240,77 @@ def fetch_smt_data(cfg, proxies):
...
+        # and we may not be able to immediately reach the update
+        # infrastructure.
+        # As time passes is mor likely that the frameowrk completes routing
+        # setup for outbound traffic. Therefore we wait less time the longer
+        # we are in the process of attempting to access the update
+        # infrastrcuture.
+        # Processing delays are handled at two levels; the individual request
+        # timeout and the time we wait between the retry of obtaining the
+        # update server information. For each timeout the starting number
+        # 15 and 20 seconds, respectiveky is an arbitrary choice. The maximum
```
smarlowucf, line 254: s/respectiveky/respectively/

lib/cloudregister/registerutils.py
```diff
@@ -240,35 +240,77 @@ def fetch_smt_data(cfg, proxies):
...
+                        if retry_cnt < max_attempts:
+                            log_msg = 'Waiting %d seconds before next attempt'
+                            logging.info(log_msg % retry_timeout)
+                            time.sleep(retry_timeout)
+                            attempt += 1
+                            continue
+                        else:
+                            logging.error('Exiting without registration')
+                            sys.exit(1)
         if (not response) or (not response.status_code == 200):
```
smarlowucf, line 314: I think this is dead code now as 311-313 would catch this state?

lib/cloudregister/registerutils.py
```diff
@@ -240,35 +240,77 @@ def fetch_smt_data(cfg, proxies):
...
+                    logging.error('\tNo response from: %s' % srvName)
+                    if srv == region_servers[-1]:
+                        logging.error('\tNone of the servers responded')
+                        logging.error('\t\tAttempted: %s' % region_servers)
+                        if retry_cnt < max_attempts:
+                            log_msg = 'Waiting %d seconds before next attempt'
+                            logging.info(log_msg % retry_timeout)
+                            time.sleep(retry_timeout)
+                            attempt += 1
+                            continue
```
smarlowucf, line 310: This could be removed as there's no other logic before the loop restarts.

lib/cloudregister/registerutils.py
```diff
@@ -240,35 +240,77 @@ def fetch_smt_data(cfg, proxies):
...
+                    )
+                    if response.status_code == 200:
+                        attempt = max_attempts
+                        break
+                    else:
+                        logging.error('=' * 20)
+                        logging.error(
+                            'Server returned: %d' % response.status_code
+                        )
+                        logging.error('Server error: "%s"' % response.reason)
```
ikapelyukhin, line 298: `attempts` counter isn't increased in this `else` branch.

If the server returns a non-200 status, but no exception is raised -- wouldn't it be caught in an infinite loop?

There's handling for this down below (`if (not response) or (not response.status_code == 200):`) -- but it's outside the `while attempt` loop.

rjschwei:
@ikapelyukhin @smarlowucf pushed all the changes should be in good shape now. Thanks

smarlowucf:
You could use a finally clause to handle the `if srv == region_servers[-1]` check in a single place.

lib/cloudregister/registerutils.py
```diff
@@ -240,37 +240,86 @@ def fetch_smt_data(cfg, proxies):
...
+                        logging.error('=' * 20)
+                        if srv == region_servers[-1]:
+                            logging.error('\tAll servers reported an error')
+                            if retry_cnt < max_attempts:
+                                log_msg = 'Waiting %d ' % retry_timeout
+                                log_msg += 'seconds before next attempt'
+                                logging.info(log_msg )
+                                time.sleep(retry_timeout)
+                                attempt += 1
```
ikapelyukhin, line 307: Hmm, in both places `attempt` increment is inside `if retry_cnt < max_attemps` branch.

So if `attempt = 2`, then `retry_cnt = attempt + 1 = 3`, then the condition `if retry_cnt < max_attemps` turns into `if 3 < 3`, which is false and that's an infinite loop again.

I would suggest `for attempt in range(max_attempts):` instead of a `while` loop.

PR #34, opened 2020-06-09
Introduce containerbuild-regionsrv service

rhafer:
This provides a small service can be used by container building tools like "docker" or "podman build" to obtain the required info for being able to install packages inside the container image using the Public Cloud RMTs. It is susposed to be used in conjunction with "container-suseconnect" when building images inside on-demand public cloud instances.

This goes together with a couple of changes in container-suseconnect which we're currently finalizing in https://github.com/SUSE/container-suseconnect/tree/new-authentication-scheme.

### DIFF
--- a/usr/lib/systemd/system/containerbuild-regionsrv.service
+++ b/usr/lib/systemd/system/containerbuild-regionsrv.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Provides Cloud SMT info and credentials to container build tools
+After=network-online.target guestregister.service
+Wants=network-online.target
+
+[Service]
+ExecStart=/usr/sbin/containerbuild-regionsrv
+Type=simple
+
+[Install]
+WantedBy=multi-user.target
--- a/usr/lib/systemd/system/containerbuild-regionsrv.service
+++ b/usr/lib/systemd/system/containerbuild-regionsrv.service
@@ -4,6 +4,7 @@ After=network-online.target guestregister.service
 Wants=network-online.target
 
 [Service]
+EnvironmentFile=-/etc/sysconfig/containerbuild-regionsrv
 ExecStart=/usr/sbin/containerbuild-regionsrv
 Type=simple
 
--- a/usr/sbin/containerbuild-regionsrv
+++ b/usr/sbin/containerbuild-regionsrv
@@ -19,7 +19,22 @@ import base64
 import os
 import socketserver
 import json
+import logging
+import urllib3
 
+# Disable the urllib warnings
+# We have server certs that have no subject alt names
+# cloudregister.registerutils() checks the server state API without
+# certificate validation
+urllib3.disable_warnings()
+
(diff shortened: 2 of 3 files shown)


usr/sbin/containerbuild-regionsrv
```diff
@@ -0,0 +1,109 @@
...
+
+logging.basicConfig(level=loglevel,
+                    format='%(levelname)s: %(message)s'
+                    )
+
+LOG = logging.getLogger('containerbuild-regionsrv')
+
+class ContainerBuildTCPServer(socketserver.BaseRequestHandler):
+    """
+    A TCP server that submits configuration details that are relevant to
```
rjschwei, line 41: s/submits/emits/

usr/sbin/containerbuild-regionsrv
```diff
@@ -0,0 +1,109 @@
...
+# License as published by the Free Software Foundation; either
+# version 3.0 of the License, or (at your option) any later version.
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.
+
+from cloudregister.registerutils import get_instance_data, get_smt, get_config, get_credentials
```
rjschwei, line 17: Use
`import cloudregister.registerutils as utils`
style and then prefix calls with `utils.` to match the style in other parts of the code.

usr/lib/systemd/system/containerbuild-regionsrv.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Provides Cloud SMT info and credentials to container build tools
+After=network-online.target guestregister.service
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/sysconfig/containerbuild-regionsrv
```
smarlowucf, line 7: Typo? s/=-/=

rhafer: No. The `-` means that the file is optional and systemd won't complain when it is not present.

smarlowucf: Ah, ack.

usr/sbin/containerbuild-regionsrv
rhafer: :heavy_check_mark:

usr/sbin/containerbuild-regionsrv
rhafer: :heavy_check_mark:

rhafer:
@rjschwei would you mind taking another look at this?

rjschwei:
Thanks

PR #39, opened 2020-07-28
- Consider the proxy environment settings even if the configuration file

### LINKED ISSUE #38: Proxy environment variables not considered if file /etc/sysconfig/proxy does not exist
Due to how cloudregister.registerutils.set_proxy() function is written, if the file /etc/sysconfig/proxy does not exist, the function will return before even looking at environment variables "http_proxy" and "https_proxy".

See https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L845 for the relevant code.

I believe this is an unintended behaviour, and certainly is un-intuitive for user.

rjschwei:
does not exists
  + It is possible the system is set up to source proxy environment variable
    settings from a non standard location and we should honor the environment
    settings no matter how they got there.
- Closes #38

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -841,9 +841,6 @@ def set_as_current_smt(smt):
 # ----------------------------------------------------------------------------
 def set_proxy():
     """Set up proxy environment if applicable"""
-    proxy_config_file = '/etc/sysconfig/proxy'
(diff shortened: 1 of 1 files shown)


smarlowucf:
Does this change anything functionally? Since both routes return the same result does order matter?

PR #40, opened 2020-07-31
- Introduce fail-over delay

rjschwei:
+ Initial network connection during the early boot phase of a cloud VM
    my be unstable. This may trigger a fail-over during the registration
    process. This in turn leads to inconsistent entries in the DB of the
    update servers and my lead to update problems. Before failing over we
    give the network some extra time to settle, up to 9 s. Before switching
    to the fail-over server we verify that the system credentials are known.


lib/cloudregister/registerutils.py
```diff
@@ -608,8 +608,31 @@ def get_smt(cache_refreshed=None):
...
+                # iteration as the network has had more time to become
+                # operational.
+                for delay in [5, 3, 1]:
+                    time.sleep(delay)
+                    msg = 'Waiting for current server to show up for %d s'
+                    logging.info(msg % delay)
+                    if current_smt.is_responsive():
+                        logging.info(
+                            'No failover needed, system access recoverd'
+                        )
```
jesusbv, line 625: s/recoverd/recovered

lib/cloudregister/registerutils.py
```diff
@@ -608,8 +608,31 @@ def get_smt(cache_refreshed=None):
...
+                # itself out. We make the wait time shorter for each
+                # iteration as the network has had more time to become
+                # operational.
+                for delay in [5, 3, 1]:
+                    time.sleep(delay)
+                    msg = 'Waiting for current server to show up for %d s'
+                    logging.info(msg % delay)
+                    if current_smt.is_responsive():
+                        logging.info(
+                            'No failover needed, system access recoverd'
```
jesusbv, line 624: s/recoverd/recovered

rjschwei:
Waiting with merge for outcome of result of discussion w.r.t. killing traffic on port 80, i.e. switching everything to https, which would potentially break existing SMT setups.

rjschwei:
Based on feedback for the request w.r.t. killing traffic for http it is going to be a more involved change. Merging this and starting a new branch dedicated to supporting https only.

PR #42, opened 2020-09-23
- Use the return code of SUSEConnect to detect errors

rjschwei:
+ As a hold over from SMT and SMT registration script we were parsing
    the return string to detect an error condition. SUSEConnect sets an error
    code thus we should use it to be a more reliable detector for registration
    errors.


ikapelyukhin (approved):
LGTM. Looks like the old logic was there since `suse_register` times.

smarlowucf (approved):

PR #43, opened 2021-01-27
Add support for no_proxy env variable

krrenjith:
Currently registercloudguest command doesn't read NO_PROXY from /etc/sysconfig/proxy file.  This will cause EC2 metadata requests being sent to proxy despite NO_PROXY configuration exist in the proxy config file to bypass proxy for EC2 metadata requests. Reading NO_PROXY from the proxy config file and set it to the environment should fix this issue.


rjschwei (approved):

rjschwei:
@krrenjith Thanks for the contribution.

PR #45, opened 2021-04-23
- Check if zypper is running and wait for up to 30 seconds if zypper is

rjschwei:
running before abandoning product list acquisition and registration.
- Fix return type on error conditions in product list creation to avoid
  traceback
- Check zypper exit code to avoid race with other processes starting
  zypper.


jesusbv (approved):
LGTM

PR #46, opened 2021-04-29
IPv6 handling

bear454:
Make a best-effort connection - trying IPv6 addressed region servers
before falling back to IPv4 addresses; properly format the URL depending
on the addressing scheme in use.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -17,6 +17,7 @@ import base64
 import configparser
 import filecmp
 import glob
+import ipaddress
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -274,17 +287,20 @@ def fetch_smt_data(cfg, proxies):
                 'Getting update server information, attempt %d' % retry_cnt
             )
             for srv in region_servers:
-                srvName = srv.strip()
-                logging.info('\tUsing region server: %s' % srvName)
+                logging.info('\tUsing region server: %s' % srv)
```
rjschwei, line 290: There needs to be a test of the network configuration, if the network is not configured for IPv6 then the IPv6 addresses in the list should be skipped.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
bear454: Added in c9754a9 using `socket.has_ipv6` [1]

[1] https://docs.python.org/3/library/socket.html#socket.has_ipv6

rjschwei (approved):

PR #48, opened 2021-06-14
New package to enable/disable access due to AHB

jesusbv:
A package running with a timer to enable/disable
the access to the infrastructure from Azure instances.

If running instance is MSFT and the license type
is BYOS, guestregister service is not enable and
the infra repositories are not available.

This references bsc#1182026

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -97,6 +97,17 @@ Requires:       python3-dnspython
 %description plugin-azure
 Guest registration plugin for images intended for Microsoft Azure
 
+%package addon-azure
+Version:	0.0.1
+Release:	0
+Summary:	Enable/Disable Guest Registration for Microsoft Azure
+Group:		Productivity/Networking/Web/Servers
+Requires:	cloud-regionsrv-client >= 9.0.0
+Requires:	cloud-regionsrv-client-plugin-azure
+
+%description addon-azure
+Enable/Disable Guest Registration for Microsoft Azure
+
 %prep
 %setup -q
 
@@ -111,21 +122,35 @@ mkdir -p %{buildroot}/var/lib/regionService/certs
 mkdir -p %{buildroot}/var/lib/cloudregister
 install -d -m 755 %{buildroot}/%{_mandir}/man1
 install -m 644 man/man1/* %{buildroot}/%{_mandir}/man1
+install -m 644 usr/lib/systemd/system/regionsrv-enabler.service %{buildroot}%{_unitdir}
+install -m 644 usr/lib/systemd/system/regionsrv-enabler.timer %{buildroot}%{_unitdir}
 gzip %{buildroot}/%{_mandir}/man1/*
 
 %pre
 %service_add_pre guestregister.service containerbuild-regionsrv.service
+
+%pre addon-azure
+%service_add_pre regionsrv-enabler.timer
 
 %post
 %{_sbindir}/switchcloudguestservices
 %{_sbindir}/updatesmtcache
 %service_add_post guestregister.service containerbuild-regionsrv.service
 
+%post addon-azure
+%service_add_post regionsrv-enabler.timer
+
 %preun
 %service_del_preun guestregister.service containerbuild-regionsrv.service
 
+%preun addon-azure
+%service_del_preun regionsrv-enabler.timer
+
 %postun
 %service_del_postun guestregister.service containerbuild-regionsrv.service
+
+%postun addon-azure
+%service_del_postun regionsrv-enabler.timer
 
 %files
 %defattr(-,root,root,-)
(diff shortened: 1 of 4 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -1257,3 +1267,67 @@ def __replace_url_target(config_files, new_smt):
...
+    out, err = exec_subprocess(cmd, True)
+    if err:
+        logging.info('Can not get the license type: %s', err.decode('utf-8'))
+        return None
+
+    return 'microsoft' in out.decode('utf-8').lower()
+
+
+# ----------------------------------------------------------------------------
+def _is_cloud_guest_registered():
```
rjschwei, line 1314: What's wring with the `is_registered()` from utils?

lib/cloudregister/registerutils.py
```diff
@@ -1257,3 +1267,67 @@ def __replace_url_target(config_files, new_smt):
                 current_service_server,
                 new_smt.get_FQDN()))
             new_config.close()
+
+
+# ----------------------------------------------------------------------------
+def _get_license_type():
```
rjschwei, line 1273: But the server makes this decision, why do we have to care about he license type on the client side as well?

lib/cloudregister/registerutils.py
```diff
@@ -1100,6 +1100,16 @@ def update_ca_chain(cmd_w_args_lst):
     return 1
 
 
+# ----------------------------------------------------------------------------
+def is_plugin_allowed():
```
rjschwei, line 1104: The plugin is definitely allowed, we have to decide what actions to take.

lib/cloudregister/registerutils.py
```diff
@@ -1257,3 +1267,67 @@ def __replace_url_target(config_files, new_smt):
...
+            sys.exit(1)
+
+            imds_content = res.json()
+            return imds_content.get('compute').get('licenseType')
+
+    return ''
+
+
+# ----------------------------------------------------------------------------
+def _is_msft():
```
rjschwei, line 1302: Do we really care about framework specific setting? I think we shouldn't.

usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -31,6 +31,14 @@ utils.set_proxy()
 class SUSECloudPlugin(Plugin):
     def RESOLVEURL(self, headers, body):
         """Convert the URL for the repo"""
+        if not utils.is_plugin_allowed():
+            logging.warning('URL resolver plugin not available.')
+            self.error(
+                {'CREDENTIAL_ERROR': 'NOTAVAILABLE'},
+                'URL resolver plugin not available'
+            )
+            return
```
rjschwei, line 40: We want to get out of the way, not sabotage the update process. zypper expects an answer this just breaks zypper up

rjschwei (changes requested):

rjschwei:
Looks to me there is an insufficient understand of the task at hand.

1.) When an instance gets switched from PAYG to BYOS it will be registered
2.) When the user runs a `zypper` command such as `zypper up` the server will reject the request.

So far so good and everything works as expected.

3.) The user is responsible for setting up a new registration either with their own SUMa, RMT, or against SCC
4.) The code needs to detect that we are no longer pointing to a server in the update infrastructure and get out of the way, **NOT** abort any update process that may be in progress and is accessing repos from elsewhere.

All of this is framework independent.

lib/cloudregister/registerutils.py
```diff
@@ -1257,3 +1262,26 @@ def __replace_url_target(config_files, new_smt):
                 current_service_server,
                 new_smt.get_FQDN()))
             new_config.close()
+
+
+# ----------------------------------------------------------------------------
+def __is_cloud_guest_registered(server_name):
```
rjschwei, line 1268: Again whether or not the system is already registered is already implemented in utils. We are not going to have 2 different implementations for the same thing.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
rjschwei: This really has to be renamed. Or we have to come up with a new definition for what registration means. Today registration means:
- There is at least 1 service that has a url that matches the server name or our plugin name
- The system has credentials (we assume the credentials are accepted by the server, this is not verified)

Given that utils already has a `is_registered` function that implements those conditions having a second, private function that is called `__is_cloud_guest_registered` is simply confusing and no one in 1/2 year will be able to figure out why one is called in some cases and the other is called in other cases when they appear, by name to do the same thing.

rjschwei: And I don't see a call to this code anymore, maybe I missed it.

rjschwei (changes requested):

rjschwei:
@jesusbv I think we need to take this out of GH for now and first write up the various flows so we can develop a better understanding of what we actually need to implement. I'll start a confluence page.

rjschwei (approved):

PR #50, opened 2021-09-30
Added registration availability check

schaefi:
Added util method is_registration_supported() which checks
if a registration process is available based on the supported
package manager backend. This change helps to configure the
behavior of registercloudguest on non SUSE based products.
Product families for which there is no subscription/registration process available will
exit after the repository hosts setup from registercloudguest

### DIFF
--- a/etc/regionserverclnt.cfg
+++ b/etc/regionserverclnt.cfg
@@ -10,3 +10,4 @@ instanceArgs = none
 
 [service]
 verifyAccess = none
+supportsPackageBackend = zypper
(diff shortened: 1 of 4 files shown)


etc/regionserverclnt.cfg
```diff
@@ -10,3 +10,4 @@ instanceArgs = none
 
 [service]
 verifyAccess = none
+supportsPackageBackend = zypper
```
rjschwei, line 13: I think we should just call this `packageBackend`

rjschwei (approved):

PR #51, opened 2021-10-06
Add email and reg code (BYOS instances)

jesusbv:
- "--regcode" option is not compatible with "--instance-data" option
- Email and reg code options are used together

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -16,7 +16,7 @@
 #
 
 
-%define base_version 9.2.1
+%define base_version 9.3.1
 Name:           cloud-regionsrv-client
 Version:        %{base_version}
 Release:        0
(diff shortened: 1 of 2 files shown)


bear454:
Why make the arguments conflict?

cloud-regionsrv-client.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 
-%define base_version 9.2.1
+%define base_version 9.3.1
```
rjschwei, line 19: There are going to be a number of changes to support AHBv3 and RMT as SCC relay, version bump is pre-mature.

usr/sbin/registercloudguest
```diff
@@ -123,10 +123,16 @@ def register_modules(extensions, products, registered=[]):
                 '--product',
                 triplet
             ]
-            if os.path.exists(instance_data_filepath):
+            if os.path.exists(instance_data_filepath) and not args.reg_code:
```
rjschwei, line 126: We are going to need a change to SUSEConnect. We still want the instance data on the RMT server for which we are the proxy.

rjschwei (changes requested):

rjschwei:
FYI I started the conversation with the SCC team: https://github.com/SUSE/connect/pull/462

PR #52, opened 2021-10-07
- BYOS to RMT and AHBv3 support

rjschwei:
+ When a BYOS system pointing to SCC gets redirected to the update
    infrastructure we need to remove the system from SCC so it no longer
    gets tracked against he user subscription and we need to remove the
    repository artifacts.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1120,7 +1120,11 @@ def remove_registration_data():
         if response.status_code == 204:
             logging.info('System sucessfully removed from SCC')
         else:
-            logging.info('System is not known to SCC')
+            scc_check_msg = 'System not found in SCC. The system may still '
+            scc_check_msg += 'be tracked against your subscription. It is '
+            scc_check_msg += 'recommended to investigate the issue. '
+            scc_check_msg += 'Local changes applied.'
+            logging.info(scc_check_msg)
         __remove_repo_artifacts('suse.com')
     else:
         logging.info('No current registration server set.')


rjschwei:
@jesusbv @bear454 ping ping

lib/cloudregister/registerutils.py
```diff
@@ -904,6 +961,22 @@ def is_registered(smt):
     return None
 
 
+# ----------------------------------------------------------------------------
+def is_scc_connected():
+    """If any repo url points to suse.com then at least some of the
+       repositories come from SCC and the SCCcredentials file is
+       expected to have the system credentials for SCC access."""
```
bear454, line 968: If the SCC crendentials file is what we're after, shouldn't the check verify it is present as well? I can easily imagine a misconfigured system where there is a repo with baseurl of suse.com, but /etc/zypp/credentials.d/SCCcredentials has been removed.

rjschwei: Well that is handled here 

https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/52/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R1113

I don't think an explicit check is necessary as the result would be the same. If the file does not exist we get an auth object that is basically `NULL` and then SCC when we call `DELETE` will return something other than `204` and as such we will continue to proceed to remove any references to `suse.com` and the system if tracked by some means in SCC would remain tracked and would require manual clean up. We probably need a better error message.

lib/cloudregister/registerutils.py
```diff
@@ -1033,12 +1106,22 @@ def remove_registration_data():
...
-        __remove_credentials(server_name)
-        __remove_repos(server_name)
-        __remove_service(server_name)
+        __remove_repo_artifacts(server_name)
         os.unlink(smt_data_file)
-        if os.path.exists('/etc/SUSEConnect'):
-            os.unlink('/etc/SUSEConnect')
+    elif is_scc_connected():
+        logging.info('Removing system from SCC')
+        user, password = get_credentials(
```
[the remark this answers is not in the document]
rjschwei: Error message was improved in flight while this conversation was happening.

jesusbv (approved):

PR #58, opened 2021-11-05
- AWS handle local zones

rjschwei:
+ Properly return the region id for local zones and "regular" availability
    zones to extract the region information as the region hint.

### DIFF
--- a/lib/cloudregister/amazonec2.py
+++ b/lib/cloudregister/amazonec2.py
@@ -45,9 +45,18 @@ def generateRegionSrvArgs():
         return
 
     if zoneResp.status_code == 200:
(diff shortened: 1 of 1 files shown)


jesusbv (approved):
Good comment

bear454 (approved):
It works (I do find it a bit unintuitive). 

One nitpick: `zoneResp` is the only camel cased variable in the function; the rest are snake cased.

rjschwei:
> It works (I do find it a bit unintuitive).
> 
> One nitpick: `zoneResp` is the only camel cased variable in the function; the rest are snake cased.

Well there are also `metaDataUrl`, `zoneInfo` but good call I'll fix them all.

PR #60, opened 2021-11-10
- Region servers can have a DNS resolvable name

rjschwei:
+ Region servers can be configured with a DNS resolvable name. The current
    code assumed address based access only. This change handles region servers
    with DNS names. In a situation where region servers are specified as a
    mix of IP addresses and DNS names the servers with DNS names are at
    the end of the lookup list.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -255,6 +255,7 @@ def fetch_smt_data(cfg, proxies):
         region_servers_ipv4 = []
         region_servers_ipv6 = []
         region_servers_dns = []
+        ip_addr = None
         for srv in region_servers:
             srv_id = srv.strip()
             try:
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -255,13 +255,13 @@ def fetch_smt_data(cfg, proxies):
         region_servers_ipv4 = []
         region_servers_ipv6 = []
         region_servers_dns = []
-        ip_addr = None
         for srv in region_servers:
             srv_id = srv.strip()
             try:
                 ip_addr = ipaddress.ip_address(srv_id)
             except ValueError:
                 region_servers_dns.append(srv_id)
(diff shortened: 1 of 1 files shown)


bear454 (approved):
Thanks for catching that, Robert.

lib/cloudregister/registerutils.py
```diff
@@ -254,8 +254,12 @@ def fetch_smt_data(cfg, proxies):
         # trying IPv6 first, if available.
         region_servers_ipv4 = []
         region_servers_ipv6 = []
+        region_servers_dns = []
         for srv in region_servers:
-            ip_addr = ipaddress.ip_address(srv.strip())
+            try:
+                ip_addr = ipaddress.ip_address(srv.strip())
+            except ValueError:
+                region_servers_dns.append(srv)
```
jesusbv, line 262: shouldn't we keep it also with `strip()` (s/srv/srv.strip()) ?

rjschwei: Good catch, thanks. Fixed.

lib/cloudregister/registerutils.py
```diff
@@ -254,8 +254,13 @@ def fetch_smt_data(cfg, proxies):
...
         region_servers_ipv6 = []
+        region_servers_dns = []
         for srv in region_servers:
-            ip_addr = ipaddress.ip_address(srv.strip())
+            srv_id = srv.strip()
+            try:
+                ip_addr = ipaddress.ip_address(srv_id)
+            except ValueError:
+                region_servers_dns.append(srv_id)
             if isinstance(ip_addr, ipaddress.IPv6Address):
```
mrosile, line 264: If there are no region servers listed by IP address (only by hostname), then by the time execution reaches this line, an exception is raised because `ip_addr` is not defined.

UnboundLocalError: local variable 'ip_addr' referenced before assignment

jesusbv (approved):
LGTM

PR #63, opened 2022-01-03
- Update man page to include information about registration for BYOS …

rjschwei:
…instances

### DIFF
--- a/man/man1/registercloudguest.1
+++ b/man/man1/registercloudguest.1
@@ -3,18 +3,18 @@
 .\"
 .TH registercloudguest.1
 .SH NAME
-registercloudguest \- Register a SUSE Linux Enterprise on-demand instance
+registercloudguest \- Register a SUSE Linux Enterprise instance
 .SH SYNOPSIS
 .B registercloudguest [option]
 .SH DESCRIPTION
 .B registercloudguest
-Register a SUSE Linux Enterprise on-demand instance with the closest, in most
-cases region local SUSE operated update infrastructure. The client will reach
-out to a so called region server to obtain information about the region local
+Register a SUSE Linux Enterprise instance with the closest, in most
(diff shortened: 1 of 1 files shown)


man/man1/registercloudguest.1
```diff
@@ -3,18 +3,18 @@
...
 .SH SYNOPSIS
 .B registercloudguest [option]
 .SH DESCRIPTION
 .B registercloudguest
-Register a SUSE Linux Enterprise on-demand instance with the closest, in most
-cases region local SUSE operated update infrastructure. The client will reach
-out to a so called region server to obtain information about the region local
+Register a SUSE Linux Enterprise instance with the closest, in most
+cases region local, SUSE operated update infrastructure. The client will reach
+out to a so called region server to obtain information about the closest
```
jesusbv, line 13: 'closest (by default)' ?

Same for line 11
One can run `registercloudguest --smt-ip <IP_OF_NOT_THE_CLOSEST_UPDATE_SERVER>...`

Or maybe we don't care/mind too much that use case

man/man1/registercloudguest.1
```diff
@@ -45,6 +60,14 @@ Force a new registration. This causes the previous registration data to be
 discarded and the system to get re-registered as if it had never been
 registered previously. This is an "emergency" switch and should be used only
 in exceptional circumstances. This option and "--clean" are mutually exclusive.
+This option does not work for BYOS instances that use the update
```
asmorodskyi, line 63: Since it is one sentence I think one mention of "BYOS instances" can be omitted : 

` This option does not work for BYOS instances that use the update infrastructure and will result in a registration failure.`

or 

`This option does not work and will result in a registration failure if used with BYOS instances.`

jesusbv (approved):
LGTM

asmorodskyi:
LGTM

PR #65, opened 2022-02-02
Do not run registercloudguest if no AHB

jesusbv:
This Fixes bsc#1195156

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -99,7 +99,7 @@ Requires:       python3-dnspython
 Guest registration plugin for images intended for Microsoft Azure
 
 %package addon-azure
-Version:	1.0.0
+Version:	1.0.1
 Release:	0
 Summary:	Enable/Disable Guest Registration for Microsoft Azure
 Group:		Productivity/Networking/Web/Servers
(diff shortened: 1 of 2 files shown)


usr/sbin/regionsrv-enabler-azure
```diff
@@ -81,7 +81,9 @@ license_type = get_license_type()
 if 'BYOS' in license_type and not utils.uses_rmt_as_scc_proxy():
     run_command(['registercloudguest', '--clean'])
     run_command(['systemctl', 'disable', service_name])
-else:
+elif license_type:
```
rjschwei, line 84: `and 'BYOS' not in license_type:`

rjschwei (changes requested):

rjschwei (approved):

cloud-regionsrv-client.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 
-%define base_version 9.3.1
+%define base_version 9.3.2
```
rjschwei, line 19: No need for this bump, the main code has not changed and the sub-package has it's own version

rjschwei (approved):

PR #66, opened 2022-02-18
Lowercase MSFT region hints

jesusbv:
- Fix bsc#1196146

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -88,7 +88,7 @@ Requires:       cloud-regionsrv-client >= 6.0.0
 Guest registration plugin for images intended for Amazon EC2
 
 %package plugin-azure
-Version:        1.0.1
+Version:        2.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Microsoft Azure
 Group:          Productivity/Networking/Web/Servers
--- a/tests/test_azureplugin.py
+++ b/tests/test_azureplugin.py
@@ -85,6 +85,12 @@ def test_metadata_request_success(mock_request):
     result = azure.generateRegionSrvArgs()
     assert result == 'regionHint=useast'
 
+    mock_request.reset()
+    mock_request.return_value = _get_expected_response_metadata(True)
+    result = azure.generateRegionSrvArgs()
+    assert result == 'regionHint=useast'
+
+
 
 # ----------------------------------------------------------------------------
 @patch('msftazure.dns.resolver.get_default_resolver')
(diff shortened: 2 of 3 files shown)


cloud-regionsrv-client.spec
```diff
@@ -16,7 +16,7 @@
 #
 
 
-%define base_version 10.0.0
+%define base_version 10.1.0
```
rjschwei, line 19: There's no change to the base code. Please revert to 10.0.0

lib/cloudregister/VERSION
```diff
@@ -1 +1 @@
-10.0.0
+10.1.0
```
rjschwei, line 1: ditto

rjschwei (changes requested):

cloud-regionsrv-client.spec
jesusbv: Updated the wrong version, fixed it now.
Version updated for the plugin-azure

cloud-regionsrv-client.spec
```diff
@@ -88,7 +88,7 @@ Requires:       cloud-regionsrv-client >= 6.0.0
 Guest registration plugin for images intended for Amazon EC2
 
 %package plugin-azure
-Version:        1.0.1
+Version:        1.1.0
```
rjschwei, line 91: Should be 2.0.0 we are changing the output in an incompatible way.

rjschwei (changes requested):

cloud-regionsrv-client.spec
jesusbv: Done

joelgordon:
Since this is on the client, we will simply start seeing the request as the lowercase version, correct? I was worried about duplicate region names, but I assume this lower-casing will happen on the client side. If this is the case, this will only solve the issue for instances that have the latest cloud-regionsrv-client.

rjschwei (approved):

tests/test_azureplugin.py
```diff
@@ -277,12 +298,16 @@ def _get_no_nameservers_resolver():
...
     """Return a response that matches extensions config"""
     response = Response()
+    region = 'useast'
+    if upper:
+        # test region hint from MSFT are treated in lowercase
+        region = region.upper()
     response.status_code = 200
     data = 'the_doc '
-    data += '<Location>useast</Location>'
+    data += f"<Location>{region}</Location>"
```
rjschwei, line 310: This will break the tests on SLE 12 which has Python 3.4 and does not support format strings. I'll fix it in a follow on PR. "Cosmetic" changes like this are really unnecssary.

rjschwei:
Merging to make sure we get this in and then out in the next release of the package due to bsc#[1196305](https://bugzilla.suse.com/show_bug.cgi?id=1196305)

PR #67, opened 2022-02-23
Fix bsc#1196305

jesusbv:


lib/cloudregister/registerutils.py
```diff
@@ -967,6 +967,8 @@ def is_new_registration():
 def is_registered(smt_server_name):
     """Check if the instance is already registerd"""
     # For a "valid" registration we need to have credentials and a service
+    if isinstance(smt_server_name, smt.SMT):
+        smt_server_name = smt_server_name.get_FQDN()
```
rjschwei, line 971: :( The argument that is expected for the method is fairly clear with `smt_server_name`, i.e. we expect a name and not an object. As such the fix should be where the call is made and the object is passed. It should not be on this method to do any testing of the argument type.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: > :( The argument that is expected for the method is fairly clear with `smt_server_name`, i.e. we expect a name and not an object. As such the fix should be where the call is made and the object is passed. It should not be on this method to do any testing of the argument type.

I agree, but the error came from the test `registerutils.is_registered(registerutils.get_current_smt())`
(`get_current_smt` returning a SMT object) so, then the calls to `is_registered` should be updated (including that test) otherwise, the error will show up so, if the caller needs to update the way they call...test (and probably others) need to be updated and there is no regression ?

rjschwei: In the refactor introduced with commit 4b478bbb3025dd the interface was changed, which is why there was a major version bump to 10.0.0. 

the `is_registered()` function used to take an object and now takes name as indicated by `smt_server_name`. The test will have to be updated. The same problem that we see in the test also exists in https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/regionsrv-enabler-azure#L89 which also needs to be fixed.

This slipped through during review of 4b478bbb3025dd, and of course in general that commit was not really a necessary change.

We can revert 4b478bbb3025dd but not th eversion bump that came along with it. So that would put us into weird territory. @smarlowucf can you please update the test to work with both versions?

if `from cloudregister import check_registration` succeeds pass the SMT object, to `is_registered` if it fails pass the name. This makes the function in img-prrof more complicated, but we can switch on the code implementation rather than looking up a specific version.

I will fix regionsrv-enabler-azure. And this PR gets closed.

jesusbv: > @smarlowucf can you please update the test to work with both versions? 
> if `from cloudregister import check_registration` succeeds pass the SMT object, to `is_registered` if it fails pass the name. This makes the function in img-prrof more complicated, but we can switch on the code implementation rather than looking up a specific version.
> 
> I will fix regionsrv-enabler-azure. And this PR gets closed.

Another (easier) solution to the fix is to keep accepting a SMT object for `is_registered`, 
call it with the SMT object in [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L557) instead of the string, and update the [check](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L967) as follows
```python
has_services(smt.get_FQDN()) and __has_credentials(smt.get_FQDN())
```

This would keep the test, img_proof and regionsrv-enabler-azure without change plus keeping the refactor as intended

Thoughts ?

jesusbv: I have committed the changes, so it is easier to see

rjschwei:
Superseded by #68 and https://github.com/SUSE-Enceladus/img-proof/pull/329

PR #68, opened 2022-02-23
Adapt to interface changes for is_registered()

rjschwei:
The interface for is_registered expects a string of the FQDN for the update server, it no longer takes and object

### DIFF
--- a/tests/test_azureplugin.py
+++ b/tests/test_azureplugin.py
@@ -307,9 +307,7 @@ def _get_proper_extensions_response(upper = False):
         region = region.upper()
     response.status_code = 200
     data = 'the_doc '
-    data += '<Location>useast</Location>'
-    if upper:
-        data += '<Location>USEAST</Location>'
+    data += '<Location>{0}</Location>'.format(region)
     data += 'last'
     response.text = data
     return response
--- a/tests/test_azureplugin.py
+++ b/tests/test_azureplugin.py
@@ -307,7 +307,9 @@ def _get_proper_extensions_response(upper = False):
         region = region.upper()
     response.status_code = 200
     data = 'the_doc '
-    data += f"<Location>{region}</Location>"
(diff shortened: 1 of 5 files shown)


tests/test_azureplugin.py
```diff
@@ -307,7 +307,9 @@ def _get_proper_extensions_response(upper = False):
         region = region.upper()
```
jesusbv, line 307: if we are not going to use f-string and the upper check, (as lines 310-312) in favour of more explicit info, shouldn't we remove this ? or keep this and remove lines 310-312 ?

rjschwei: Sorry, I do not understand. You added the part that triggers on `upper`, I just changed it such that we do not use format string which is not available on Python 3.4 (SLE 12). If the `upper` part id not used why did you add it in the first place?

jesusbv: so before, I added f-string as:
```python
region = 'foo'
if upper:
    region = region.upper()
bar = f"{foo}"
```
and AFAICT the intention is to use upper but not f-string
```python
if upper:
    data += 'USEAST'
else:
    data += 'useast'
```
instead of having two locations in data (for the purpose of the test) and removing the now unnecesary
```
if upper:
    region = region.upper()
```
so, I meant to remove the previous `if-upper+region.upper` and adapt the `data +=useast` to be lower or upper case accordingly

tests/test_azureplugin.py
```diff
@@ -307,7 +307,9 @@ def _get_proper_extensions_response(upper = False):
...
     response.status_code = 200
     data = 'the_doc '
-    data += f"<Location>{region}</Location>"
+    data += '<Location>useast</Location>'
+    if upper:
```
jesusbv, line 311: this should be `if-else` IMHO with `data += ...` being set inside each block

jesusbv (approved):

PR #69, opened 2022-03-17
Add status registration message

jesusbv:
- Show registration status in console

This Fixes bsc#1197113

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -49,6 +49,7 @@ from requests.auth import HTTPBasicAuth
 # We have server certs that have no subject alt names
 # We have to check the server state API without certificate validation
 urllib3.disable_warnings()
+registration_returncode = 0
 
 # ----------------------------------------------------------------------------
 def register_modules(extensions, products, registered=[]):
@@ -79,6 +80,7 @@ def register_modules(extensions, products, registered=[]):
             )
             res = p.communicate()
             if p.returncode != 0:
+                registration_returncode = p.returncode
                 error_message = res[0].decode()
                 logging.error('\tRegistration failed: %s' % error_message)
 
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -417,6 +419,7 @@ while not base_registered:
         stderr=subprocess.PIPE)
     res = p.communicate()
     if p.returncode != 0:
+        print('Registration failed, see /var/log/guestregister for details')
```
rjschwei, line 422: We do not want to print the failure message for every update server we are trying, one summary message, you'll have to set a flag.

usr/sbin/registercloudguest
```diff
@@ -79,9 +79,11 @@ def register_modules(extensions, products, registered=[]):
             )
             res = p.communicate()
             if p.returncode != 0:
+                print('Registration failed, see /var/log/guestregister for details')
```
rjschwei, line 82: Ditto, one summary message if registration succeeds or fails.

rjschwei (changes requested):

usr/sbin/registercloudguest
jesusbv: Ack, done.

usr/sbin/registercloudguest
```diff
@@ -484,6 +488,11 @@ prod_data = json.loads(res.text)
 extensions = prod_data.get('extensions')
 register_modules(extensions, products)
 
+if registration_returncode != 0:
+    print('Registration failed, see /var/log/guestregister for details')
```
rjschwei, line 492: This is an error and needs to go to stderr, i.e. `file=sys.stderr'

exit code is set later in the code.

rjschwei (changes requested):

rjschwei (approved):

PR #71, opened 2022-03-17
Remvoe enabler dependency on the registration service

rjschwei:
+ The enabler for the repos is an independent service and must remain
  independent such that the registration can be enabled in code under the
  pre-determined conditions.


jgleissner (approved):

jesusbv (approved):
Good catch. LGTM

PR #73, opened 2022-04-07
Add license cache

jesusbv:
- Cache the license type to track PAYG license switch to PAYG license
- Bump version

### DIFF
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -106,7 +106,8 @@ elif license_type and 'BYOS' not in license_type:
     # or PAYG without AHB
     update_server = utils.get_smt()
     if update_server:
-        if not has_license_changed(license_type) and utils.is_registered(update_server.get_FQDN()):
+        if (not has_license_changed(license_type) and
+            utils.is_registered(update_server.get_FQDN())):
             sys.exit(0)
         update_license_cache(license_type)
         update_server = None
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -20,7 +20,7 @@ import sys
 
 import cloudregister.registerutils as utils
 
-CACHE_LICENSE_PATH = utils.REGISTRATION_DATA_DIR + 'cached_license'
+CACHE_LICENSE_PATH = utils.get_state_dir() + 'cached_license'
 
 
 def update_license_cache(license_type):
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -20,19 +20,19 @@ import sys
 
 import cloudregister.registerutils as utils
 
+CACHE_LICENSE_PATH = '/tmp/cached_license'
+
 
 def update_license_cache(license_type):
     """Update the cache to track the license type"""
-    cached_license_path = '/tmp/cached_license'
-    with open(cached_license_path, 'w+') as cached_license:
+    with open(CACHE_LICENSE_PATH, 'w+') as cached_license:
         cached_license.write(license_type)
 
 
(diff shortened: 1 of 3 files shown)


usr/sbin/regionsrv-enabler-azure
```diff
@@ -20,6 +20,23 @@ import sys
...
+    with open(cached_license_path, 'w+') as cached_license:
+        cached_license.write(license_type)
+
+
+def has_license_changed(license_type):
+    cached_license_path = '/tmp/cached_license'
+    try:
+        with open(cached_license_path, 'r') as cached_license:
+            old_license = cached_license.read()
+            return license_type != old_license
```
jesusbv, line 35: we could do 
```python
license_type != cached_license.read()
```
maybe `old_license` is better for redability

usr/sbin/regionsrv-enabler-azure
```diff
@@ -20,6 +20,25 @@ import sys
 
 import cloudregister.registerutils as utils
 
+CACHE_LICENSE_PATH = '/tmp/cached_license'
```
smarlowucf, line 23: Does cache need to persist long term?

rjschwei, line 23: The client already has a directory where it stores state, `REGISTRATION_DATA_DIR = '/var/lib/cloudregister/'` in utils. This would warrant a new function in the utils `get_state_dir()`.

Further if/when `/tmp` moves to tmpfs the information will be lost on reboot.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -86,8 +105,9 @@ elif license_type and 'BYOS' not in license_type:
     # or PAYG without AHB
     update_server = utils.get_smt()
     if update_server:
-        if utils.is_registered(update_server.get_FQDN()):
```
rjschwei, line 89: Not the only change needed, we also have

`if 'BYOS' in license_type and not utils.uses_rmt_as_scc_proxy():`

(line 100)

currently this means that the commands in the if condition get executed every time the timer runs and the condition is met. However now that we have a cache we can optimize this behavior and stop sending so many messages to the systemlog.

That means the new condition is:

`if 'BYOS' in license_type and has_license_changed(license_type) and not utils.uses_rmt_as_scc_proxy():`

rjschwei (changes requested):

usr/sbin/regionsrv-enabler-azure
jesusbv: yes

usr/sbin/regionsrv-enabler-azure
jesusbv: True, I though it would not be a big issue to clean every BYOS attempt, updating it now

usr/sbin/regionsrv-enabler-azure
```diff
@@ -78,16 +97,18 @@ utils.start_logging()
...
+    not utils.uses_rmt_as_scc_proxy()):
     run_command(['registercloudguest', '--clean'])
     run_command(['systemctl', 'disable', service_name])
 elif license_type and 'BYOS' not in license_type:
     # if no license type, system could be a BYOS without AHB
     # or PAYG without AHB
     update_server = utils.get_smt()
     if update_server:
-        if utils.is_registered(update_server.get_FQDN()):
+        if not has_license_changed(license_type) and utils.is_registered(update_server.get_FQDN()):
```
smarlowucf, line 109: This line pretty long compared to the rest of the code.

smarlowucf (approved):

usr/sbin/regionsrv-enabler-azure
jesusbv: Fixed

rjschwei (approved):

PR #74, opened 2022-04-08
Register dependent modules of skipped modules

rjschwei:
+ If a specific module or product is recommended it gets installed with the
  product or module that has the recommendation. However the extensions of
  the recommended module do not get installed automatically. In cases where
  those extensions are requested we still need to install them even if the
  top level module was installed as recommended.


jesusbv (approved):
LGTM

PR #75, opened 2022-04-11
When license is BYOS and has changed, update license

jesusbv:


smarlowucf (approved):
LGTM.

As a note/question not related to this PR. If license_type is coming from metadata it might not be in our control. It may be good to treat this as case insensitive when doing comparisons in case it changes. `license_type.upper() == 'BYOS'`

rjschwei (approved):

PR #76, opened 2022-04-12
Use new method in executable as well

rjschwei:
- With e7cda81 the get_state_dir() function was introduced, use it when
  access to the state directory is needed rather than the utils static
  variable.


jesusbv (approved):
Good catch

PR #80, opened 2022-05-02
Disable checking of service starts

jesusbv:


pdostal:
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval

usr/lib/systemd/system/regionsrv-enabler-azure.service
```diff
@@ -2,8 +2,7 @@
 Description=Enable/Disable Guest Registration for Microsoft Azure
 After=network-online.target
 Wants=network-online.target
-StartLimitIntervalSec=500
-StartLimitBurst=5
+StartLimitBurst=0
```
rjschwei, line 5: Per the man page this is not correct, at least not the way I read it.

"Use StartLimitIntervalSec= to configure the checking interval and StartLimitBurst= to configure how many starts per interval are allowed."

Meaning `StartLimitBurst=0` allows 0 starts of the unit per interval and the checking interval is a default.

"interval is a time span with the default unit of seconds, but other units may be specified, see systemd.time(5). Defaults to DefaultStartLimitIntervalSec= in manager configuration file, and may be set to 0 to disable any kind of rate limiting"

Meaning we should be setting `StartLimitIntervalSec` to disable the checking for rate limits.

"burst is a number and defaults to DefaultStartLimitBurst= in manager configuration file."

does not define 0 to have a special meaning. If we want the timer to run 1 time every minute, as indicated in the .timer file

`OnCalendar=*:0/1`

and we do not want to trigger any rate limiting, even if the timer is started manually, then we need to have

`StartLimitIntervalSec=0`

and drop the 

`StartLimitBurst`

setting.

jesusbv: Done.

(guess [this](https://www.suse.com/support/kb/doc/?id=000019750) is out of date)

jesusbv: Updated the commit message, should be good now.

rjschwei: Well the knowledge base article states

"To disable checking of service starts, please set StartLimitBurst to a value of 0 (read : 'zero')"

This is not mentioned in the man page for systemd timers, as such one would have to look at the code to confirm the information. We should stick to the documented behavior. The article should be reviewed, I'll see if I can poke someone to to that.

rjschwei (approved):

PR #81, opened 2022-05-20
Changes to support SLE-Micro

rjschwei:
+ SLE-Micro has a read-only root file system and handles files in /var
  in a specific manner. These changes modify locations of specific files
  to accomodate the SLE-Micro setup.


jesusbv (approved):
LGTM

PR #83, opened 2022-07-20
Handle deregistration exception

rjschwei:
+ The request to remove the client system from the update server as registered
  client may throw an exception, for example when the certs have been tampered
  with. Handle the exception and clean up the local artifacts.


jesusbv (approved):
LGTM

PR #85, opened 2022-08-24
Fixes for ro root

rjschwei:
- While the code was updated for the read-only root setup of SLE-Micro the
  spec file was not updated to accomodate the new locations for the cache
  that is needed by cloud-regionsrv-client and the generic config was not
  updated for the cert location of region server certs. This update addresses
  the problem introduced with 531d162a2cfb


KeithMnemonic (approved):
LGTM

PR #91, opened 2022-11-12
Handle update server cert change

rjschwei:
- At present when an update server has a cert change the client will fail and the user has to force a re-registration in order to get the new cert onto the client system. This change makes the import of the new cert automatic. A cert check is run every time zypper is executed.
- Improve test coverage of the utils code

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, SUSE LLC, All rights reserved.
+# Copyright (c) 2022, SUSE LLC, All rights reserved.
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -1274,6 +1274,50 @@ def update_ca_chain(cmd_w_args_lst):
...
+    for region_rmt_server in region_rmt_servers:
+        region_ipv4 = region_rmt_server.get_ipv4()
+        region_ipv6 = region_rmt_server.get_ipv6()
+        for cached_rmt_server in cached_rmt_servers:
+            cached_ipv4 = cached_rmt_server.get_ipv4()
+            cached_ipv6 = cached_rmt_server.get_ipv6()
+            if (region_ipv4 == cached_ipv4) and (region_ipv6 == cached_ipv6):
+                if region_rmt_server == cached_rmt_server:
+                    break
+        else:
```
apozsuse, line 1311: This else block in under indented (x2), right?

apozsuse:
In this function to save the update servers cache:

```
def __populate_srv_cache():
    """Populate the registration server cache"""
    proxies = None
    if set_proxy():
        proxies = {
            'http_proxy': os.environ.get('http_proxy'),
            'https_proxy': os.environ.get('https_proxy')
        }
    cfg = get_config()
    region_smt_data = fetch_smt_data(cfg, proxies)
    cnt = 1
    for child in region_smt_data:
        update_server = smt.SMT(child)
        server_cache_file_name = AVAILABLE_SMT_SERVER_DATA_FILE_NAME % cnt
        store_smt_data(
            get_state_dir() + server_cache_file_name, update_server
        )
 ```
 
 The cnt counter should be increased in the for loop, shouldn't it?

File: lib/cloudregister/registerutils.py [hunk 2]
```diff
@@ -1274,6 +1274,50 @@ def update_ca_chain(cmd_w_args_lst):
...
+    if is_new_registration():
+        return
+    cached_rmt_servers = get_available_smt_servers()
+    proxy = set_proxy()
+    proxies = None
+    if proxy:
+        http_proxy = os.environ.get('http_proxy')
+        https_proxy = os.environ.get('https_proxy')
+        no_proxy = os.environ.get('no_proxy')
+        proxies = {'http_proxy': http_proxy,
```
smarlowucf, line 1291: Nitpick: Seem like this could all be in dict constructor and drop the temporary variables.

tests/test_registerutils.py
```diff
@@ -12,29 +12,180 @@
...
 import inspect
+import glob
 import os
 import sys
 
 test_path = os.path.abspath(
     os.path.dirname(inspect.getfile(inspect.currentframe())))
 code_path = os.path.abspath('%s/../lib' % test_path)
 config_path = os.path.abspath('%s/../etc' % test_path)
+data_path = test_path + os.sep + 'data/'
```
smarlowucf, line 23: Could use os.path.join with a trailing empty string to get the correct sep based on os.

lib/cloudregister/registerutils.py
```diff
@@ -1274,6 +1274,50 @@ def update_ca_chain(cmd_w_args_lst):
...
+def update_rmt_certs():
+    """Check if the update server certs have changed and if yes update the
+       system accordingly"""
+    # We are currently getting the latest cert information, nothing to do
+    if is_new_registration():
+        return
+    cached_rmt_servers = get_available_smt_servers()
+    proxy = set_proxy()
+    proxies = None
+    if proxy:
```
jesusbv, line 1287: variable `proxy` can be removed and use
```python
if set_proxy():
```
then the name of the method could be discussed as well, but that's outside the review

usr/lib/zypp/plugins/urlresolver/susecloud
```diff
@@ -39,6 +39,7 @@ class SUSECloudPlugin(Plugin):
         repo_credentials = headers.get('credentials')
         if zypper_pid != prev_zypper_pid:
             verify_credentials = True
+            utils.update_rmt_certs()
```
jesusbv, line 42: removing the condition inside the method, mentioned above
could be called here
```python
if not is_new_registration():
    utils.update_rmt_certs()
```

rjschwei:
> 

Long existing bug, good catch, thanks. Fixed.

rjschwei:
Sorry force pushed, lack of attention this A.M.

lib/cloudregister/registerutils.py
```diff
@@ -1362,6 +1403,7 @@ def __populate_srv_cache():
         store_smt_data(
             get_state_dir() + server_cache_file_name, update_server
         )
+        cnt += 1
```
jesusbv, line 1406: IMHO this should be a different PR ?

apozsuse:
LGTM :+1:

apozsuse (approved):

smarlowucf (approved):

PR #95, opened 2022-11-16
Plugin load refactor

rjschwei:
This is only the refactoring. The additional use case will be part of a follow up PR. The idea is to write the fqdn of the target update server and the region to a file in cache. WE can then use that information to check if we are still in the same region or not.


lib/cloudregister/registerutils.py
```diff
@@ -1355,6 +1359,29 @@ def __get_referenced_credentials(smt_server_name):
...
+            msg = 'Configured and loaded module "%s" does not provide the '
+            msg += 'required generateRegionSrvArgs function.'
+            logging.error(msg % plugin.__file__)
+            return region_srv_args
+        if not region_srv_args:
+            retry_cnt += 1
+            time.sleep(1)
+            continue
+        logging.info('Region server arguments: ?%s' % region_srv_args)
+        break
```
jesusbv, line 1380: Consider return here instead of break and return

apozsuse (approved):
LGTM :+1:

lib/cloudregister/registerutils.py
```diff
@@ -1333,6 +1318,25 @@ def get_state_dir():
...
 # Private
+# ----------------------------------------------------------------------------
+def __get_framework_plugin(cfg):
+    """Return the configured framework specific plugin module"""
+    mod = None
+    if cfg.has_section('instance'):
+        module = None
+        if cfg.has_option('instance', 'instanceArgs'):
+            module = cfg.get('instance', 'instanceArgs')
+        if module and module != 'none':
```
jesusbv, line 1329: should `'none'` be `None` ?

PR #97, opened 2022-11-16
Removes unused imports and other flake8 errors

apozsuse:
Pending PR to remove some unused imports and remove some pending flake8 issues.

After this PR is merged, the only errors for flake8 are:

```
flake8 tests/ lib/
tests/test_azureplugin.py:27:1: E402 module level import not at top of file
tests/test_ec2plugin.py:27:1: E402 module level import not at top of file
tests/test_gceplugin.py:27:1: E402 module level import not at top of file
tests/test_registerutils.py:28:1: E402 module level import not at top of file
tests/test_smt.py:28:1: E402 module level import not at top of file
lib/cloudregister/registerutils.py:544:22: W605 invalid escape sequence '\s'
lib/cloudregister/registerutils.py:545:22: W605 invalid escape sequence '\s'
lib/cloudregister/registerutils.py:546:20: W605 invalid escape sequence '\s'
lib/cloudregister/registerutils.py:546:24: W605 invalid escape sequence '\s'
```

Not sure if changing (in the test files with E402 error) the import for this own module before the path modification could have some impact...
```python
code_path = os.path.abspath('%s/../lib/cloudregister' % test_path)

sys.path.insert(0, code_path)

import msftazure as azure
```

I guess if we install the module with `python setup.py install -e` to run the tests we could use a regular import for the module and remove that hack? :thinking:

### DIFF
--- a/lib/cloudregister/msftazure.py
+++ b/lib/cloudregister/msftazure.py
@@ -16,12 +16,10 @@ import html
 import logging
 import requests
 import re
-import time
 import urllib.request
 import urllib.parse
 import urllib.error
 
-from html.parser import HTMLParser
 
 extensionConfigRx = re.compile(
     r'.*<ExtensionsConfig>(.*?)</ExtensionsConfig>.*', re.S | re.M
--- a/tests/test_azureplugin.py
+++ b/tests/test_azureplugin.py
@@ -12,9 +12,7 @@
 # License along with this library.
 
 import inspect
-import logging
 import os
-import pytest
 import requests
 import sys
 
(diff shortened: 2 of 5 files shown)


rjschwei:
> I guess if we install the module with `python setup.py install -e` to run the tests we could use a regular import for the module and remove that hack? thinking

If you remove the hack then we have to switch to tox for testing as tox will setup a venv and thus will test the code in the project. While all the scaffolding, such a tox, has advantages it also brings with it a heavy burden of dependencies on pypi and a bunch of other stuff. Modifying the path to find he module in the project instead of he one installed on the system is simple and works reliably without any additional dependencies, which is why I opted for the latter approach. I like simple and lightweight stuff ;)

jesusbv (approved):

PR #98, opened 2022-11-16
Write a framework identifier file to cache

rjschwei:
We write the file at boot time to identify the framework based on dmi information and region. We record the plugin so we can use the same module to regenerate the region information later. This protects us from errant configuration changes.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -37,6 +37,7 @@ from requests.auth import HTTPBasicAuth
 from cloudregister import smt
 
 AVAILABLE_SMT_SERVER_DATA_FILE_NAME = 'availableSMTInfo_%s.obj'
+FRAMEWORK_IDENTIFIER = 'framework_info'
 HOSTSFILE_PATH = '/etc/hosts'
 NEW_REGISTRATION_MARKER = 'newregistration'
 OLD_REGISTRATION_DATA_DIR = '/var/lib/cloudregister/'
@@ -107,6 +108,14 @@ def clean_hosts_file(domain_name):
     with open(HOSTSFILE_PATH, 'wb') as hosts_file:
         for entry in new_hosts_content:
             hosts_file.write(entry)
+
+
+# ----------------------------------------------------------------------------
+def clean_framework_idetntifier():
+    """Remove the framework identification data"""
+    framework_file_path = os.path.join(get_state_dir(), FRAMEWORK_IDENTIFIER)
+    if os.path.exists(framework_file_path):
+        os.unlink(framework_file_path)
 
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -109,6 +110,14 @@ def clean_hosts_file(domain_name):
             hosts_file.write(entry)
 
 
+# ----------------------------------------------------------------------------
+def clean_framework_idetntifier():
```
jesusbv, line 114: s/idetntifier/identifier

usr/sbin/registercloudguest
```diff
@@ -101,6 +101,7 @@ def cleanup():
     utils.remove_registration_data()
     utils.clean_smt_cache()
     utils.clear_new_registration_flag()
+    utils.clean_framework_idetntifier()
```
jesusbv, line 104: ditto

usr/sbin/registercloudguest
```diff
@@ -217,6 +218,8 @@ if cached_smt_servers:
     # If we have an update server cache the system is registered in
     # some way shape or form
     utils.clear_new_registration_flag()
+else:
+    utils.write_framework_identifier(cfg)
```
jesusbv, line 222: ditto

lib/cloudregister/registerutils.py
```diff
@@ -1313,8 +1327,23 @@ def uses_rmt_as_scc_proxy():
 
 
 # ----------------------------------------------------------------------------
-def get_state_dir():
-    return REGISTRATION_DATA_DIR
+def write_framework_identifier(cfg):
+    """Write a file we use as identifier to detect movement of a golden
+       image created from a registered instance"""
+    identifier = {}
+    vendor, error = exec_subprocess(
```
jesusbv, line 1334: do we not check error value ?

jesusbv: or that `vendor` is not -1  ?
which , in case `exec_subprocess` failed with `OSError`, that line would generate a `TypeError` 
because it can't unpack two values

apozsuse (commented):
Nothing additional to Jesus' comments.

The only additional thing that came my head was the change we had to do in img-proof recently to include the whole dmidecode path in the call we were doing with sudo (as there was some bug in sudo cfg file that was removing the default path), but it makes no difference here.

rjschwei:
> Nothing additional to Jesus' comments.
> 
> The only additional thing that came my head was the change we had to do in img-proof recently to include the whole dmidecode path in the call we were doing with sudo (as there was some bug in sudo cfg file that was removing the default path), but it makes no difference here.

sudo has a new environment that gets constructed as part of the sudo executable, thus the PATH settings may be different. everything that is part of cloud-regionsrv-client runs as `root` and has roots environment.

lib/cloudregister/registerutils.py
rjschwei: It is expected that the return code is **not** `-1`, i.e. everything works. The `TypeError` is triggered if `-1` is returned because `-1` is not iterable. That said, yes a total failure of the `dmidecode` command should be handled gracefully.

jesusbv (approved):
LGTM

apozsuse (approved):
LGTM :+1:

PR #99, opened 2022-11-18
Create region info on package update

rjschwei:
We are going to depend on the region info cache file for vwraious conditions. Therefore we have to make sure it exists. Create the file when the package gets updated.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -140,6 +140,7 @@ gzip %{buildroot}/%{_mandir}/man1/*
 %post
 %{_sbindir}/switchcloudguestservices
 %{_sbindir}/updatesmtcache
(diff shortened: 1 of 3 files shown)


jesusbv (approved):

rjschwei:
> LGTM +1
> 
> For the sake of my understanding, is there any reason why we do not include the scripts in `setup.py` (in scripts?), but only in OBS spec file?
> 
> Is it because we only plan to use this client with the OBS generated rpm (and NOT with as module with pip)?

The scripts are not in setup.py because we do not use the pythonic (sdist) way to create the tarball. That is encoded in the Makefile. The Makefile is also used to check for version consistency between the spec file and the version file of the code. And yes there is no intention to distribute the code in any other way than with SLE. Everything is very SLE specific.

PR #100, opened 2022-11-18
Refactor the way we detect a region change

rjschwei:
Switch the region change detaction over from using cached update server data to using the new region information file. This allows us to handle cert changes of the update servers in a cleaner way. Previously a cert change on an update server would have triggered the new region detection path unnecessarily.


apozsuse (approved):
LGTM, that part is WAY clearer now!

jesusbv (approved):

PR #103, opened 2022-11-21
Handle cert changes

rjschwei:
Detect if the cert of the current target update server has changed. If it has update the client system by importing the new cert. Also check the hosts entry and update if necessary.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -917,7 +917,7 @@ def has_region_changed(cfg):
         region = region_hint.split('=')[-1]
 
     if framework == 'unknown' or region == 'unknown':
-        # We cannot determine with certainy if anything has changed
+        # We cannot determine with certainty if anything has changed
         # Assume everything is as it was
         return False
 
@@ -1363,6 +1363,9 @@ def update_rmt_cert(server):
             if region_rmt_server != server:
                 import_smt_cert(region_rmt_server)
                 return True
+
+    return False
+
 
 # ----------------------------------------------------------------------------
 def uses_rmt_as_scc_proxy():
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -350,6 +350,14 @@ def fetch_smt_data(cfg, proxies):
 
 
 # ----------------------------------------------------------------------------
+def finalize_registration():
+     """During registration we only write information about the update server
+        that is used for the registration to the cache. But we want to cache
+        information about all update servers."""
+     __populate_srv_cache()
+
+
+# ----------------------------------------------------------------------------
 def find_equivalent_smt_server(configured_smt, known_smt_servers):
     """Find an SMT server that is equivalent to the currently configured
        SMT server, only consider responsive servers"""
@@ -908,6 +916,11 @@ def has_region_changed(cfg):
         region_hint = __get_region_server_args(plugin)
         region = region_hint.split('=')[-1]
 
+    if framework == 'unknown' or region == 'unknown':
+        # We cannot determine with certainy if anything has changed
+        # Assume everything is as it was
+        return False
+
     try:
         registered_region = json.loads(
             open(get_framework_identifier_path()).read()
@@ -922,6 +935,19 @@ def has_region_changed(cfg):
         return False
 
     return True
+
+
+# ----------------------------------------------------------------------------
+def has_rmt_in_hosts(server):
+    """Check if an entry for the given update server is in the hosts file"""
+    hosts_content = open('/etc/hosts').read()
+    srv_ipv4 = server.get_ipv4()
+    srv_ipv6 = server.get_ipv6()
+
+    if srv_ipv4 in hosts_content or srv_ipv6 in hosts_content:
+        return True
+
+    return False
 
 
 # ----------------------------------------------------------------------------
@@ -1313,13 +1339,13 @@ def update_ca_chain(cmd_w_args_lst):
 
 
 # ----------------------------------------------------------------------------
-def update_rmt_certs():
-    """Check if the update server certs have changed and if yes update the
-       system accordingly"""
+def update_rmt_cert(server):
+    """Import the cert for the given server if it has changed"""
     # We are currently getting the latest cert information, nothing to do
     if is_new_registration():
         return
-    cached_rmt_servers = get_available_smt_servers()
+    target_ipv4 = server.get_ipv4()
+    target_ipv6 = server.get_ipv6()
     proxies = None
     if set_proxy():
         proxies = {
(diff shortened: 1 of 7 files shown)


rjschwei:
@apozsuse @jesusbv sorry this got a little bigger and more complex than I had hoped especially since I ended up ripping out code that was just introduced to handle the cert update in a different way. Anyway, this is a better approach than what I was going for with the other implementation.

lib/cloudregister/registerutils.py
```diff
@@ -908,6 +916,11 @@ def has_region_changed(cfg):
         region_hint = __get_region_server_args(plugin)
         region = region_hint.split('=')[-1]
 
+    if framework == 'unknown' or region == 'unknown':
+        # We cannot determine with certainy if anything has changed
```
apozsuse, line 920: s/certainy/certainty/

lib/cloudregister/registerutils.py
```diff
@@ -1330,28 +1356,13 @@ def update_rmt_certs():
...
-            )
-            import_smt_cert(region_rmt_server)
-            clean_smt_cache()
-            __populate_srv_cache()
-            return
-
+        if (region_ipv4 == target_ipv4) and (region_ipv6 == target_ipv6):
+            if region_rmt_server != server:
+                import_smt_cert(region_rmt_server)
+                return True
```
apozsuse, line 1366: Including the 'return False' explicitly might be clearer?

lib/cloudregister/smt.py
```diff
@@ -33,6 +33,10 @@ def __init__(self, smtXMLNode, https_only=False):
             self._ipv6 = smtXMLNode.attrib['SMTserverIPv6']
         except KeyError:
             pass
+        try:
+            self._region = smtXMLNode.attrib['region']
+        except KeyError:
```
apozsuse, line 38: there's a trailing space after the colon, not sure why it's not highlighted by github web

lib/cloudregister/smt.py
```diff
@@ -108,13 +117,20 @@ def get_ipv6(self):
...
+    def get_region(self):
+        """Return the region name this server is associated with"""
+        return self._region
+
     # --------------------------------------------------------------------
     def is_equivalent(self, smt_server):
-        """When 2 SMT servers have the same cert fingerprint and their
-           FQDN is the same they are equivalent."""
+        """Have both an ipv4 address and/or both an ipv6 address and they
+           are in the same region they are interchangable and considered
```
apozsuse, line 128: s/interchangable/interchangeable/

tests/test_registerutils.py
```diff
@@ -152,52 +157,86 @@ def test_is_registration_supported_RHEL_Family():
...
-    assert res is None
+def test_has_rmt_in_hosts_has_ipv4():
+    hosts_content = """
+    # simulates hosts file containing the ipv4 we are looking for in the test
 
+    1.1.1.1   smt-foo.susecloud.net  smt-foo
+    """
+    server = MockServer()
+    with mock.patch('builtins.open', mock.mock_open(read_data=hosts_content)):
+        has_entry = utils.has_rmt_in_hosts(server)
```
apozsuse, line 168: Trailing whitespace and the following empty line has 4 spaces also.

tests/test_registerutils.py
```diff
@@ -152,52 +157,86 @@ def test_is_registration_supported_RHEL_Family():
...
+    hosts_content = """
+    # simulates hosts file containing the ipv4 and iv6 we are looking for
+    # in the test
+
+    1.1.1.1   smt-foo.susecloud.net  smt-foo
+    11:22:33:44::::00   smt-foo.susecloud.net  smt-foo
+    """
+    server = MockServer()
+    with mock.patch('builtins.open', mock.mock_open(read_data=hosts_content)):
+        has_entry = utils.has_rmt_in_hosts(server)
```
apozsuse, line 183: trailing whitespace + empty line with 4 spaces

lib/cloudregister/registerutils.py
```diff
@@ -924,6 +937,19 @@ def has_region_changed(cfg):
...
+    hosts_content = open('/etc/hosts').read()
+    srv_ipv4 = server.get_ipv4()
+    srv_ipv6 = server.get_ipv6()
+
+    if srv_ipv4 in hosts_content or srv_ipv6 in hosts_content:
+        return True
+
+    return False
+
+
```
apozsuse, line 952: Unlike as ipv4, ipv6 comparisons can be a bit tricky because of the different ways ipv6 can be written.

For example all these IPs are the same but written differently:
```
11:22:33:44:0:0:0:55
0011:0022:0033:0044:0000:0000:0000:0000:0055
11:22:33:44::55
```
If we want to match IPv6, parsing the string into a ipaddress struct is better approach.

tests/test_registerutils.py
```diff
@@ -152,52 +157,86 @@ def test_is_registration_supported_RHEL_Family():
...
+    assert True == has_entry
+
+
+def test_has_rmt_in_hosts_has_ipv4_6():
+    hosts_content = """
+    # simulates hosts file containing the ipv4 and iv6 we are looking for
+    # in the test
+
+    1.1.1.1   smt-foo.susecloud.net  smt-foo
+    11:22:33:44::::00   smt-foo.susecloud.net  smt-foo
```
apozsuse, line 179: That ipv6 format is invalid, should be `11:22:33:44:0:0:0:00` or `11:22:33:44::00`

tests/test_registerutils.py
```diff
@@ -152,52 +157,86 @@ def test_is_registration_supported_RHEL_Family():
...
-    utils.update_rmt_certs()
-    assert import_cert.called
-    assert cache_clean.called
-    assert pop_cache.called
+
+def test_has_rmt_in_hosts_has_ipv6():
+    hosts_content = """
+    # simulates hosts file containing the ipv6 we are looking for in the test
+
+    11:22:33:44::::00   smt-foo.susecloud.net  smt-foo
```
apozsuse, line 205: ditto

apozsuse (changes requested):
ipv6 matching comparing strings is bad approach, parsing the ipv6 string into a struct is a better way to handle the different ways of writing ipv6 addresses.

rjschwei:
> ipv6 matching comparing strings is bad approach, parsing the ipv6 string into a struct is a better way to handle the different ways of writing ipv6 addresses.

Well we are not doing a general ipv6 comparison. We already know that the source, i.e. the information we get from the region server is valid. As such the simple string comparison is sufficient.

apozsuse (approved):
LGTM :+1:

jesusbv (approved):
Not sure `__ne__(self, other_smt)` method is being used

Other than that, LGTM

PR #104, opened 2022-11-23
Update to version 1.0.7

rjschwei:
- Fix annoying new line addition in /etc/hosts every time the target server gets updated. This used to leave many empty lines between the server entry created by the client and other entries
- Fixes needed discovered by testing for commits e56088e8 - de01b311b
- Update the test instructions
- Integrate patch file we have in the build service in the spec file to avoid clobbering by copying the spec file from git to the build service package
- test updates


lib/cloudregister/registerutils.py
```diff
@@ -327,6 +344,8 @@ def fetch_smt_data(cfg, proxies):
                         if srv == region_servers[-1]:
                             logging.error('\tAll servers reported an error')
                 except requests.exceptions.RequestException:
+                    if quiet:
+                        continue
```
jesusbv, line 348: with this we are not showing the logging errors, I understand that is intentional
shouldn't we show the logging errors instead of no info ?

IMHO the logging of the error should show up unconditionally

apozsuse (approved):
LGTM :+1:

lib/cloudregister/registerutils.py
rjschwei: We do not want to pollute the log file every time zypper runs. I was trying to find a better way to skip trying to access the region servers via IPv6 but have not yet has a good idea. Ultimately if there is a total failure, i.e. we cannot fins any regionsrv that responds then we get the message from line 345 and during debugging we known what happened. The detailed messages about the region servers are in the log at the beginning when we first register the system. I see no point in polluting the log file with these messages over and over again.

jesusbv: >  I was trying to find a better way to skip trying to access the region servers via IPv6 

Maybe check if the srv is IPv6 and skip it ?
```python
if isinstance(ipaddress.ip_address(srv), ipaddress.IPv6Address):
    continue
```
could have an extra param too, `skip_ipv6` (similar to `quiet`)

PR #105, opened 2022-11-24
Disable IPv6 on SLES12 builds (bsc#1203382)

bear454:
Add a patch to disable IPv6, only apply it when building SLE12 packages.


bear454:
Sorry for not submitting this before the SR.

rjschwei:
We do not need the patch in this project as one would typically copy the spec file and the tarball into the build service project. The spec file changes are already part of #104

PR #107, opened 2022-12-19
Add system UUID info to instance data

jesusbv:
Send system UUID in the metadata
to use as the system identifier
instead of system_token

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -16,7 +16,7 @@
 #
 
 
-%define base_version 10.0.8
+%define base_version 10.1.0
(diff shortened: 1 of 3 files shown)


jesusbv:
> Why do we need it? 

Currently, `uuid` info is sent in `SUSEConnect`, if that changed or we did not use `SUSEConnect`, 
`cloud-regionsrv-client` would not have a way to send the `uuid`
I'm OK closing the PR but `uuid` not working would be an issue (as it was `system_token`)  

> And the new version would be 10.0.9

It is not a bug fix, so it would not be a patch update right ?

jesusbv:
we will use instance id, not the system UUID

PR #108, opened 2023-01-18
Bug 1207133 - Adds rotating logfile feature

apozsuse:
This PR implements the functionality to rotate the log file in order not to allow the log file to grow indefinitely.

The log rotation is ruled by 2 constants:
- DEFAULT_LOG_ROTATING_SIZE = 2 * 1024 * 1024  #2MB
- DEFAULT_LOG_BACKUP_COUNT = 2

With this default values, once the log file goes over 2MB the files is rotated and 2 backup files are kept ( so we will end up with `/var/log/cloudregister`, `cloudregister.1` and `cloudregister.2`. 

The PR includes a test to verify the log rotation works as expected.

There are some caveats about the tests:
- Pytest has to be run with the `-p no:logging` to disable pytest capturing log. This option has been included in a pytest configuration file `pytest.ini`.
- The unit test for the log rotation has to be the first run in the file. If some log is written before calling `start_logging` calls in utils, the configuration is not applied.


rjschwei:
We do not want to implement our own log rotation. This is handled by system services. Please consult the rsyslog documentation https://www.rsyslog.com/doc/master/tutorials/log_rotation_fix_size.html  or we can modify the unit file https://bbs.archlinux.org/viewtopic.php?id=200009 to let systemd handle the rotation.

rjschwei (changes requested):
See general comment about log rotation

apozsuse:
> See general comment about log rotation

OK!
As far as I checked (and if I checked it right), it seems not possible yet to set up a custom configuration in journald for a single systemd service, so I'll explore the rsyslog option....
I'll close this PR and address the issue in a new one.
Thanks!

PR #109, opened 2023-01-20
Includes context managers for file management in utils

apozsuse:
I spotted some parts in the utils file that could be improved with the use context managers.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -52,16 +52,16 @@ def add_hosts_entry(smt_server):
 
     smt_hosts_entry_comment = '\n# Added by SMT registration do not remove, '
     smt_hosts_entry_comment += 'retain comment as well\n'
+    smt_ip = smt_server.get_ipv4()
+    if has_ipv6_access(smt_server):
+        smt_ip = smt_server.get_ipv6()
+    entry = '%s\t%s\t%s\n' % (
+        smt_ip,
+        smt_server.get_FQDN(),
+        smt_server.get_name()
+    )
     with open('/etc/hosts', 'a') as hosts_file:
         hosts_file.write(smt_hosts_entry_comment)
-        smt_ip = smt_server.get_ipv4()
-        if has_ipv6_access(smt_server):
-            smt_ip = smt_server.get_ipv6()
-        entry = '%s\t%s\t%s\n' % (
-            smt_ip,
-            smt_server.get_FQDN(),
-            smt_server.get_name()
-        )
         hosts_file.write(entry)
     logging.info('Modified /etc/hosts, added: %s' % entry)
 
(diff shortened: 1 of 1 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -482,15 +482,16 @@ def get_credentials(credentials_file):
...
-        if entry.startswith('username'):
-            username = entry.split('=')[-1].strip()
-        elif entry.startswith('password'):
-            password = entry.split('=')[-1].strip()
-        else:
-            logging.warning('Found unknown entry in '
-                            'credentials file "%s"' % entry)
+    with open(credentials_file) as cred_file:
+        credentials = cred_file.readlines()
+        for entry in credentials:
```
smarlowucf, line 487: All the logic doing work on the file data can be outside of the context manager. This allows the file to close right after readlines.

lib/cloudregister/registerutils.py
```diff
@@ -551,18 +552,19 @@ def get_current_smt():
     ipv4_search = '%s\s' % smt_ipv4
     ipv6_search = '%s\s' % smt_ipv6
     fqdn_search = '\s%s\s' % smt_fqdn
-    hosts = open(HOSTSFILE_PATH, 'rb').read()
-    if (
+    with open(HOSTSFILE_PATH, 'rb') as hosts_file:
+        hosts = hosts_file.read()
+        if (
```
smarlowucf, line 557: Ditto

smarlowucf (commented):
I think general python convention is to exit context manager as soon as possible. I.e. once the content is read from file it can be closed. Any processing of that content can happen outside the context manager.

lib/cloudregister/registerutils.py
```diff
@@ -52,18 +52,17 @@ def add_hosts_entry(smt_server):
...
+        hosts_file.write(smt_hosts_entry_comment)
+        smt_ip = smt_server.get_ipv4()
+        if has_ipv6_access(smt_server):
+            smt_ip = smt_server.get_ipv6()
+        entry = '%s\t%s\t%s\n' % (
+            smt_ip,
+            smt_server.get_FQDN(),
+            smt_server.get_name()
+        )
+        hosts_file.write(entry)
```
rjschwei, line 65: If you are using context manager, I know calling close() is an extra line of code ;), then the creation of the string we write should be outside, i.e.

```
smt_ip = smt_server.get_ipv4()
if has_ipv6_access(smt_server):
    smt_ip = smt_server.get_ipv6()
entry = '%s\t%s\t%s\n' % (
    smt_ip,
    smt_server.get_FQDN(),
    smt_server.get_name()
)
with open('/etc/hosts', 'a') as hosts_file:
    hosts_file.write(smt_hosts_entry_comment)
    hosts_file.write(entry)
```

rjschwei (changes requested):
See inline comments about scope

jesusbv (approved):
LGTM

rjschwei (approved):

smarlowucf (approved):

PR #110, opened 2023-01-23
Add instance data cache

### LINKED ISSUE #11: consider instance data caching
Instance data is sent to the repository server with every zypper access via the URL resolver. For everyone of these requests we access the metadata server. We should support a caching mechanism that caches the metadata for a period of time and re-uses it for the zypper requests.

jesusbv:
- Add cache handling for instance data
      in order to do fewer requests to the
      metadata server
    
    This Fixes #11


lib/cloudregister/registerutils.py
```diff
@@ -44,6 +44,7 @@
 REGISTRATION_DATA_DIR = '/var/cache/cloudregister/'
 REGISTERED_SMT_SERVER_DATA_FILE_NAME = 'currentSMTInfo.obj'
 RMT_AS_SCC_PROXY_MARKER = 'rmt_is_scc_proxy'
+TIME_CACHED = 60 * 5  # 5 min in seconds
```
apozsuse, line 47: Nitpic: May 'CACHE_EXPIRATION_TIME' be more descriptive name? :thinking:

apozsuse:
Minor comment. LGTM :+1:

apozsuse (approved):

lib/cloudregister/registerutils.py
rjschwei: What is the GCE JWT token expiration time? And yes a more descriptive name is in order.

lib/cloudregister/registerutils.py
```diff
@@ -1398,6 +1405,44 @@ def write_framework_identifier(cfg):
...
+    try:
+        if instance_data_cache_expired():
+            return None
+        else:
+            cache_instance_data_path = os.path.join(
+                get_state_dir(), 'cache_instance_data'
+            )
+            with open(cache_instance_data_path) as f:
+                cached_instance_data = f.read()
+            return cached_instance_data
```
rjschwei, line 1420: This is not good enough, it opens the system to attack by replacing the `cache_instance_data` file. We'd have to prevent this by storing a checksum somewhere or doing some other tricks. Storing such state across invocations is tricky. It may be better to store the data in a tmpfile such that the cache is available within one invocation, i.e. the same process rather than trying to be available across invocations.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: Ack, I will change it to be available per invocation instead of several invocations

I understand the concern and I agree that is serious/worth changing the approach
That said, if someone wanted to fiddle with the metadata, they can do that now, right ?
I'm not saying we should not worry about this case, I'm saying that concern is valid for
other cases

rjschwei: If someone wanted to inject suitable metadata into the process now they would have to intercept the traffic to the metadata server, i.e. they need something that intercepts `http://169.254.169/254` and then they'd need to implement the suitable API. That's a lot of work. Compared to

- Run 1000 instances A (BYOS)
- Run 1 instance B (PAYG)
- Every 4 minutes rsync the metadata from instance B to 1000 instances of A into the cache file location

The later approach I can implement in about 20 minutes.

lib/cloudregister/registerutils.py
jesusbv: GCE JWT token for our rmt-verification is [300](https://gitlab.suse.de/public-cloud-team/rmt-server-verify-instance-gce/-/blob/master/gce.rb#L143)

jesusbv:
@rjschwei @apozsuse let's get this merged so it goes into the new release 
plus handy that it has a version bump

lib/cloudregister/registerutils.py
```diff
@@ -1398,6 +1404,41 @@ def write_framework_identifier(cfg):
...
+        os.unlink(get_instance_data_filepath())
+    except FileNotFoundError:
+        return
+
+def get_instance_data_filepath():
+    instance_data_cache_filepath = glob.glob(
+        os.path.join(REGISTRATION_DATA_DIR, '5U53-*')
+    )
+    try:
+        return instance_data_cache_filepath[0]
```
apozsuse, line 1435: Thinking if it might be good to sort the files by date to assure we pick the most recent one...

lib/cloudregister/registerutils.py
```diff
@@ -1398,6 +1404,41 @@ def write_framework_identifier(cfg):
...
+def set_instance_data_cache(instance_data):
+    """Update the instance data cache file with new instance data."""
+    with open(get_instance_data_filepath(), 'w') as f:
+        f.write(instance_data)
+
+# ----------------------------------------------------------------------------
+def clear_instance_data_cache():
+    """Remove instance data cache file."""
+    try:
+        os.unlink(get_instance_data_filepath())
```
apozsuse, line 1426: Should we clear all the files with the cache filename format?

usr/sbin/registercloudguest
```diff
@@ -377,32 +376,30 @@ if not os.path.exists(utils.get_state_dir()):
...
-    inst_data_out.write(instance_data)
-    inst_data_out.close()
-
 # Check if registration is supported
 if not utils.is_registration_supported(cfg):
     sys.exit(0)
 
+# Check if we need to send along any instance data
+utils.get_instance_data(cfg)
+
```
apozsuse, line 385: For a future improvement, it might be good to clear the cache with a context manager, to reduce the risk of forgetting to include the `clear_instance_data_cache` before each `exit`

lib/cloudregister/registerutils.py
jesusbv: There is only one instance data cache file per invocation

lib/cloudregister/registerutils.py
jesusbv: There is only one instance data cache file per invocation

rjschwei (changes requested):
I have already explained the problem with writing the data to disk. I will not repeat myself. Obfuscating the filename is not really a solution.

jesusbv:
> I have already explained the problem with writing the data to disk. I will not repeat myself. Obfuscating the filename is not really a solution.

before this PR, we wrote the instance data to disk anyway
that data from disk was used once and the follow up/subsequent operations that needed
the same instance data did queries to IMDS, this change write the same data to disk (as before)
with a uuid name (as before) but adds the benefits that uses a cache which is what the issue #11 suggested

rjschwei:
> > I have already explained the problem with writing the data to disk. I will not repeat myself. Obfuscating the filename is not really a solution.
> 
> before this PR, we wrote the instance data to disk anyway in [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L382)
> 
> ```python
> instance_data_filepath = utils.get_state_dir() + str(uuid.uuid4())
> instance_data = utils.get_instance_data(cfg)
> if instance_data:
>     inst_data_out = open(instance_data_filepath, 'w')
>     inst_data_out.write(instance_data)
>     inst_data_out.close()
> ```
> 
> that data from disk was used once and the follow up/subsequent operations that needed the same instance data did queries to IMDS, this change writes the same data to disk (as before) in the same directory (as before) with a uuid name (as before) but adds the benefit that uses a cache which is what the issue #11 suggested I understand your feedback/concern on having the instace data for a long period of time (in the previous changes, 5min) we agreed, based on your feedback. to use the data cached per invocation

There appears to be a lack of understanding of the issue files #11 this implementation, and the current code.

1.) The issue #11 clearly states caching for the urlresolver, this PR does not touch the urlresolver
2.) The current code already uses a cache during the `registercloudguest` process. https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L380 writes the instance data to disk. https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L79 and https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L419 use that cahce. Finally https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L512 unlinks the file such that the cache only exists during this process.

Now there may be a couple of places in the code where the cache is not properly cleaned if there is an early exit. But other than this issue, which can be addressed in other ways, this PR is a subjective cosmetic change that adds functions for the sake of having functions and ultimately is unrelated to #11, which is claimed to be addressed.

PR #111, opened 2023-01-25
Use os.path.join for paths

jesusbv:
The reasons to use `os.path.join` are that is less error prone (no need to worry if the path had a trailing slash or not) and is clearer to others that the code is dealing with filepaths

No much on portability as not sure if `cloud-regionsrv-client` was going to be installed in other platforms


rjschwei (approved):

apozsuse (approved):
LGTM :+1:

rjschwei:
@jesusbv please rebase

jesusbv:
Done, will merge it if no one objects otherwise

PR #114, opened 2023-02-08
Do not warn about system_token entry

jesusbv:
-- SCC uses system_token for duplicate systems 
  and as such, SCC writes the system_token value 
  in the credentials file, versions newer than 0.3.36 
  will not write the system_token 
  Regardless, cloud-regionsrv-client only checks 
  username and password in the credentials file so, 
  code ignores (does not show a warning) if system_token 
  entry is present


apozsuse (approved):
LGTM :+1:

rjschwei (approved):

PR #118, opened 2023-03-31
Clean up and logging

### LINKED ISSUE #117: When using --clean update the log to indicate the use of clean

rjschwei:
- Log a message when the user initiates a registration clean up with --clean. Fixes #117
- If baseproduct registration fails call cleanup, otherwise the SCCCredentials file is left behind an dthe system is not left in a prestine state.


jesusbv (approved):
LGTM

PR #120, opened 2023-05-04
Edit behaviour for IPv6

jesusbv:
- Try IPv6, if not present or fails, try IPv4 
  Methods affected: __request_cert and is_responsive

- Method __request_cert returns the cert (request text) and not the request

### DIFF
--- a/lib/cloudregister/smt.py
+++ b/lib/cloudregister/smt.py
@@ -138,14 +138,14 @@ class SMT:
     def is_responsive(self):
         """Check if the SMT server is responsive"""
         # Per rfc3986 IPv6 addresses in a URI are enclosed in []
-        ips = [self.get_ipv6(), self.get_ipv4]
+        ips = [self.get_ipv6(), self.get_ipv4()]
         for ip in filter(None, ips):
             if ip == self.get_ipv6():
-                health_url = 'https://[%s]/api/health/status' % self.get_ipv6()
-                cert_url = '%s://[%s]/smt.crt' % (self._protocol, self.get_ipv6())
+                health_url = 'https://[%s]/api/health/status' % ip
+                cert_url = '%s://[%s]/smt.crt' % (self._protocol, ip)
             else:
-                health_url = 'https://%s/api/health/status' % self.get_ipv4()
-                cert_url = '%s://%s/smt.crt' % (self._protocol, self.get_ipv4())
+                health_url = 'https://%s/api/health/status' % ip
+                cert_url = '%s://%s/smt.crt' % (self._protocol, ip)
 
             # We cannot know if the server cert has been imported into the
             # system cert hierarchy, nor do we know if the hostname is resolvable
@@ -247,7 +247,7 @@ class SMT:
                                 # enclosed in []
                                 cert_res = requests.get(
                                     '%s://[%s]/%s' % (
-                                        self._protocol, self.get_ipv6(), cert_name
+                                        self._protocol, ip, cert_name
                                     ),
                                     verify=False
                                 )
--- a/lib/cloudregister/smt.py
+++ b/lib/cloudregister/smt.py
@@ -67,11 +67,9 @@ class SMT:
     def get_cert(self):
         """Return the CA certificate for the SMT server"""
         if not self._cert:
-            cert_rq = self.__request_cert()
-            if cert_rq:
-                cert = cert_rq.text
-                if self.__is_cert_valid(cert):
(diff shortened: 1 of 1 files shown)


lib/cloudregister/smt.py
```diff
@@ -140,33 +138,35 @@ def is_equivalent(self, smt_server):
...
-                # We are pointing to an SMT server, the health status API
-                # is not available. Download the cert to at least make sure
-                # Apache is responsive
-                cert_response = requests.get(cert_url, verify=False)
-                if cert_response and cert_response.status_code == 200:
-                    return True
-        except Exception:
-            # Something is wrong with the server
-            pass
+        ips = [self.get_ipv6(), self.get_ipv4]
```
bear454, line 141: You're passing in the method `self.get_ipv4` instead of the result.

lib/cloudregister/smt.py
```diff
@@ -140,33 +138,35 @@ def is_equivalent(self, smt_server):
...
+        for ip in filter(None, ips):
```
bear454, line 142: What's the point of fetching to the list if you have to run `self.get_ipv6()` again in the same if condition you removed?

bear454 (changes requested):

lib/cloudregister/smt.py
bear454: The code only uses `ip` to again load the IP using `self.get_ipv[4,6]()` ... so I don't the value.

jesusbv: The idea is to have a loop that tries IPv6 first

lib/cloudregister/smt.py
jesusbv: Yes, I noticed it and corrected it already

lib/cloudregister/smt.py
rjschwei: The code already does IPv6 first per standard, which states that IPv6 should be preferred.

lib/cloudregister/smt.py
```diff
@@ -140,33 +138,35 @@ def is_equivalent(self, smt_server):
...
-                # Apache is responsive
-                cert_response = requests.get(cert_url, verify=False)
-                if cert_response and cert_response.status_code == 200:
-                    return True
-        except Exception:
-            # Something is wrong with the server
-            pass
+        ips = [self.get_ipv6(), self.get_ipv4()]
+        for ip in filter(None, ips):
+            if ip == self.get_ipv6():
```
rjschwei, line 143: This should then be a check if the address is IPv6, don't call `get_ipv6()` again

rjschwei (changes requested):
What is the problem that is being addressed? This looks to me as a "style preference" thta moves a bunch of code around without changing functionality or fixing an existing problem.

lib/cloudregister/smt.py
jesusbv: yes but if IPv6 failed then code fails without trying IPv4

lib/cloudregister/smt.py
jesusbv: > What is the problem that is being addressed? This looks to me as a "style preference" thta moves a bunch of code around without changing functionality or fixing an existing problem.

The issue is with the code as it is released, if IPv6 failed, no IPv4 is tried, making the registration fail

rjschwei: True, good catch. Why does this fix require the change in behavior of `__request_cert()` ?

jesusbv: It does not, it could be and should be a different PR
I will ammend this one and creates the subsequent one once merged (after version bumped)

rjschwei (changes requested):

rjschwei:
I think this should be done differently seeing the repetition of code and checks and using filter with None. We should probably form the urls once and cache them and the loop over the cache. Let me give that a try to see how it works out.

jesusbv:
> I think this should be done differently seeing the repetition of code and checks and using filter with None. We should probably form the urls once and cache them and the loop over the cache. Let me give that a try to see how it works out.

Yes, there are different refactors/changes once this is merged like, e.g., [this loop](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/8cb382dbf06d149c456cb1fc7a24fa553e48a736/lib/cloudregister/smt.py#L251-L286) can be refactored

jesusbv:
superseded by #121

PR #121, opened 2023-05-09
Server access testing expansion

rjschwei:
At present when an IPv6 address is present in the configuration received from the region server but the client network has no ipv6 capabilities the detection if he server is responsive fails as the ipv4 access is not checked. This update expands the code to check access via ipv4 shold the ipv6 access fail.

### DIFF
--- a/lib/cloudregister/smt.py
+++ b/lib/cloudregister/smt.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, SUSE LLC, All rights reserved.
+# Copyright (c) 2023, SUSE LLC, All rights reserved.
 #
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
@@ -14,6 +14,7 @@
 """Class to hold the information we need to connect and identify an SMT
    server."""
 
+import ipaddress
 import logging
 import requests
 
@@ -40,9 +41,11 @@ class SMT:
         self._fqdn = smtXMLNode.attrib['SMTserverName']
         self._fingerprint = smtXMLNode.attrib['fingerprint']
         self._cert = None
+        self._cert_names = ('smt.crt', 'rmt.crt')
         self._protocol = 'http'
         if https_only:
             self._protocol = 'https'
+        self._check_urls = self._form_srv_check_urls()
 
     # --------------------------------------------------------------------
     def __eq__(self, other_smt):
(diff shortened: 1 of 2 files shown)


rjschwei:
@jesusbv this would supersede #120 thoughts?

lib/cloudregister/smt.py
```diff
@@ -40,9 +41,11 @@ def __init__(self, smtXMLNode, https_only=False):
         self._fqdn = smtXMLNode.attrib['SMTserverName']
         self._fingerprint = smtXMLNode.attrib['fingerprint']
         self._cert = None
+        self._cert_names = ('smt.crt', 'rmt.crt')
         self._protocol = 'http'
         if https_only:
             self._protocol = 'https'
+        self._form_srv_check_urls()
```
bear454, line 48: Consider doing the assignment to `self._check_urls` here, consistent with the other assignments, instead of inside the `self._form_srv_check_urls()` method.

bear454: e.g. ending up with

```
self._check_urls = self._form_srv_check_urls()
```

jesusbv: +1 or consider renaming the method to `self._set_check_urls()`

lib/cloudregister/smt.py
```diff
@@ -139,34 +142,31 @@ def is_equivalent(self, smt_server):
...
-                # We are pointing to an SMT server, the health status API
-                # is not available. Download the cert to at least make sure
-                # Apache is responsive
-                cert_response = requests.get(cert_url, verify=False)
-                if cert_response and cert_response.status_code == 200:
-                    return True
-        except Exception:
-            # Something is wrong with the server
-            pass
+        for health_url, cert_url in self._check_urls.items():
```
jesusbv, line 151: A bit nitpicking, maybe consider using the key only 
as the chances of pointing to a server with health status 
are quite high, then getting the `cert_url` with the key, looking like

```python
for health_url in self._check_urls.keys():
    ...
    elif response.status_code == 404:
        cert_url = self._check_urls.get(health_url)
        # or 
        cert_url = self._check_urls[health_url]
        # or no variable at all and
        cert_response = requests.get(
            self._check_urls.get(health_url) + cert_name, verify=False
        )
```
so the variable would get used near(er) the place it is declared

jesusbv (approved):

PR #122, opened 2023-05-10
Prep for container use

rjschwei:
This is the start of being able to use cloud-regionsrv-client from a container. Inside a container we do not want systemd and it's dependencies. Change the macro used such that systemd is not automatically set up as package dependency on SLE 15 and later distributions.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -56,6 +56,7 @@ BuildRequires:  python3-lxml
 BuildRequires:  python3-requests
 BuildRequires:  python3-setuptools
 BuildRequires:  python3-zypp-plugin
+BuildRequires:  systemd-rpm-macros
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 
 BuildArch:      noarch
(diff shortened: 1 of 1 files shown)


cloud-regionsrv-client.spec
```diff
@@ -45,7 +45,11 @@ Requires:       regionsrv-certs
 Requires:       zypper
 BuildRequires:  systemd
 Conflicts:      container-suseconnect
+%if 0%{?suse_version} == 1315
 %{?systemd_requires}
+%else
+%{?systemd_oredering}
```
jgleissner, line 51: Shouldn't that be `systemd_ordering`?

jgleissner (changes requested):
The spec should probably contain `BuildRequires: systemd-rpm-macros`.

jgleissner:
> The spec should probably contain `BuildRequires: systemd-rpm-macros`.

It appears rpm already requires this so I guess adding this wasn't really necessary (sorry), but obviously won't harm.

rjschwei:
It's always there as was at one point the python-rpm-macros until it was not and then created a bunch of failures. Good to be explicit.

PR #124, opened 2023-07-21
Add version option

jesusbv:
Add -v and --version CLI option for registercloudguest


lib/cloudregister/version.py
```diff
@@ -0,0 +1 @@
+__VERSION__ = '10.1.2'
```
rjschwei, line 1: Let's keep the current implementation with `VERSION` and `read`

rjschwei (changes requested):

lib/cloudregister/version.py
jesusbv: OK, done
out of curiosity, is there a reason to keep it like that ?
other than that it is the current implementation

jgleissner (approved):
Seems a little more complicated than the initial implementation, but looks fine.

PR #126, opened 2023-08-28
Validate IP address

jesusbv:
With IPv6 being enabled, better check the input is OK

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -197,7 +197,7 @@ if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
             ip_addr=args.user_smt_ip,
             err=err
         )
-        print(msg)
+        print(msg, file=sys.stderr)
         sys.exit(1)
 
 
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -28,6 +28,7 @@
    3.) Register"""
 
 import argparse
+import ipaddress
 import json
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -189,6 +190,17 @@ if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
         print(msg, file=sys.stderr)
         sys.exit(1)
 
+    try:
+        ipaddress.ip_address(args.user_smt_ip)
```
rjschwei, line 194: This test can only happen `if args.user_smt_ip`

rjschwei (changes requested):
See inline comment. GitHub is definitely broken with comment detection.

usr/sbin/registercloudguest
jesusbv: but if we reach that code, it means `args.user_smt_ip` is present with a value

before that line, there is 
```python
if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
     if not (args.user_smt_ip and args.user_smt_fqdn and args.user_smt_fp):
            ...
            sys.exit(1)
     # meaning if the code reaches this part, there is args.user_smt_ip
     # otherwise, it would have exited on line 191 of this PR
     # so at this point it is guaranteed args.user_smt_ip is present and it has a value
     # no need for "if args.user_smt_ip:"

rjschwei (approved):

PR #127, opened 2023-08-29
Refactor get metadata

jesusbv:
Before the fix, the code always returned
"[Errno 2] No such file or directory: 'which ec2metadata'"
The call to subprocess should be a list, comma separated

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -180,7 +180,7 @@ def enable_repository(repo_name):
 
 
 # ----------------------------------------------------------------------------
-def exec_subprocess(cmd, return_output=False):
+def exec_subprocess(cmd, return_output=False, close_fds=True):
     """Execute the given command as a subprocess (blocking)
        Returns one off:
            - exit code of the command
@@ -190,8 +190,9 @@ def exec_subprocess(cmd, return_output=False):
         proc = subprocess.Popen(
             cmd,
             stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE
-        )
+            stderr=subprocess.PIPE,
+            close_fds=close_fds
+            )
         out, err = proc.communicate()
         if return_output:
             return (out, err)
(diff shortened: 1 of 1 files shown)


jesusbv:
> Why do we need this? There is no description.

I added the message in the commit, usually GH uses that to populate the description,
the reason for the fix is that before the fix, the code always returned
"[Errno 2] No such file or directory: 'which ec2metadata'"
The call to subprocess should be a list, comma separated
I added it as the description, the refactor was because I noticed during the fix
we have `exec_subprocess` and that could be used instead of calling `subprocess.Popen`
and reduce code/errors

rjschwei (changes requested):
Based on the description this should probably a 1 line change not a complete re-arrangement of a routine based on personal preferences in style.

jesusbv:
> Based on the description this should probably a 1 line change not a complete re-arrangement of a routine based on personal preferences in style.

Fair enought, the description + PR should be one line but even on a different PR, this change is not based on personal preference style, we have a method to wrap subprocess calls `exec_subprocess` then we have subprocess calls across the codebase that do not use that method, shouldn't every call of `subprocess` use `exec_subprocess` method/wrapper ?

rjschwei:
> > Based on the description this should probably a 1 line change not a complete re-arrangement of a routine based on personal preferences in style.
> 
> Fair enought, the description + PR should be one line but even on a different PR, this change is not based on personal preference style, we have a method to wrap subprocess calls `exec_subprocess` then we have subprocess calls across the codebase that do not use that method, shouldn't every call of `subprocess` use `exec_subprocess` method/wrapper ?

Does it really matter to have the subprocess calls? Even if we go as far and say it is nice to use the exec_subprocess implementation everywhere a switch to that does not require re-arrangement of the if conditions. What is wrong with
```
if (
            config.has_section('instance') and
            config.has_option('instance', 'dataProvider')
    ):
```
?

Well apparently you think a try-except block is a nicer style.

jesusbv:
> > > Based on the description this should probably a 1 line change not a complete re-arrangement of a routine based on personal preferences in style.
> > 
> > 
> > Fair enought, the description + PR should be one line but even on a different PR, this change is not based on personal preference style, we have a method to wrap subprocess calls `exec_subprocess` then we have subprocess calls across the codebase that do not use that method, shouldn't every call of `subprocess` use `exec_subprocess` method/wrapper ?
> 
> Does it really matter to have the subprocess calls? Even if we go as far and say it is nice to use the exec_subprocess implementation everywhere a switch to that does not require re-arrangement of the if conditions. What is wrong with
> 
> ```
> if (
>             config.has_section('instance') and
>             config.has_option('instance', 'dataProvider')
>     ):
> ```
> 
> ?
> 
> Well apparently you think a try-except block is a nicer style.

I do not, I changed that not based on nicer style but following our coding practices doc,
before that doc, I adhered to the project's coding style
I will adhere to this project's coding style instead of following/trading off our doc, no problem  :+1:

rjschwei:
Necessary code changes included in #136

PR #128, opened 2023-08-29
Check if instance has Ipv4 or IPv6 enabled

jesusbv:
When using '--force-new --smt-ip ...' options,
validate that the instance can use the --smt-ip value in order to save debugging or unnecessary errors


lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1407,40 @@ def write_framework_identifier(cfg):
...
+
+    if 'Error' in errors.decode():
+        # get 'ipv6' or 'ipv4' string from the command
+        ip_format = cmd_list[-1][-4:]
+        msg = 'This instance does not have {} enabled.'.format(ip_format)
+        return False, msg
+
+    metadata_ip_addr = output.decode().strip()
+    try:
+        return ipaddress.ip_address(metadata_ip_addr), None
```
jesusbv, line 1437: this IP address comes from the CSP, so this check could be skipped ?

usr/sbin/registercloudguest
```diff
@@ -200,6 +201,17 @@ if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
         print(msg, file=sys.stderr)
         sys.exit(1)
 
+    # check the instance has IPv6 or Ipv4 enabled
+    # so smt-ip can be either
+    ip_format = ['ipv4']
+    if isinstance(ipaddress.ip_address(arg_ip_addr), ipaddress.IPv6Address):
```
ricardobranco777, line 207:
```suggestion
    if isinstance(ipaddress.ip_address(arg_ip_addr), IPv6Address):
```

lib/cloudregister/registerutils.py
```diff
@@ -1635,3 +1644,25 @@ def __replace_url_target(config_files, new_smt):
                     current_service_server,
                     new_smt.get_FQDN())
                 )
+
+
+# -----------------------------------------------------------------------------
+def __connection_check(smt_ip_addr, ip_format):
```
rjschwei, line 1650: `ip_format` is not required as an argument. This is a simple check that can be performed in the function.

lib/cloudregister/registerutils.py
```diff
@@ -1635,3 +1644,25 @@ def __replace_url_target(config_files, new_smt):
...
+    if not smt_ip_addr:
+        # in case RMT ever dropped IPv4 support
```
rjschwei, line 1652: What does not providing the `smt_ip_addr` argument have to do with whether or not the update servers have IPv4 addresses or not?

lib/cloudregister/registerutils.py
```diff
@@ -1635,3 +1644,25 @@ def __replace_url_target(config_files, new_smt):
...
+        return False
+    logging.info('Attempt to access update server over {}'.format(ip_format))
+    protocol = 'http'  # Default for backward compatibility
+    if https_only(get_config()):
+        protocol = 'https'
+    try:
+        cert_res = requests.get(
+            '%s://%s/smt.crt' % (protocol, smt_ip_addr),
```
rjschwei, line 1660: This breaks if the IP address has IPv6 format in that case `[]` are required.

rjschwei (changes requested):
This litters various pieces of knowledge in many places making it difficult to follow the code and the intention. Combined with generic names such as `__connection_check` that appears to be implemented for IPv4 use only.

lib/cloudregister/registerutils.py
jesusbv: Well, in this PR, the call to this method adds the [] but I agree the method should add it (its responsibility) to remove the possibility of the caller forgetting to do it

rjschwei (changes requested):
The state purpose, which I agree with does not match the implementation. If we want an early exit when the user gives us an IP address, then construct an smt object in the cli code and call `is_responsive` then get out, it will amount to about a 3 line change.

lib/cloudregister/registerutils.py
```diff
@@ -1635,3 +1649,29 @@ def __replace_url_target(config_files, new_smt):
...
+            verify=False
+        )
+    except requests.exceptions.ConnectionError as err:
+        if propagate:
+            raise
+        logging.error(err)
+    except Exception:
+        logging.error('Update server not reachable for %s' % smt_ip_addr)
+        return False
+
```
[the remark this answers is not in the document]
jesusbv: The intend is to determine if the client/instance has IPv4|IPv6 enabled, `is_responsive` could be extended for that purpose if we added the exception handling (with the proper messaging) of this PR
I did not do that so `is_responsive` remains as is and no other code updates/calls to `is_responsive` needed

jesusbv:
> The state purpose, which I agree with does not match the implementation. If we want an early exit when the user gives us an IP address, then construct an smt object in the cli code and call `is_responsive` then get out, it will amount to about a 3 line change.

We want to check if the IP provided (IPv4 or IPv6) is enabled on the client, so they can actually use that IP

lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1392,35 @@ def write_framework_identifier(cfg):
         framework_file.write(json.dumps(identifier))
 
 
+# ----------------------------------------------------------------------------
+def can_reach_infra(rmt_ip):
```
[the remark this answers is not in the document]
jesusbv: I have renamed the method and update the code in that method to reflect and handle better the intend

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
rjschwei: Then check if the client has IPv6 or IPv4 enabled don't implement some detour way of making that check. There are system tools to do such checks, it is not python-requests. Also the intend stated here "The intend is to determine if the client/instance has IPv4|IPv6 enabled" does not match with the descriptive comment "When using '--force-new --smt-ip ...' options, validate that the instance can use the --smt-ip value in order to save debugging or unnecessary errors". Well if one squints really hard then maybe.

The code tries to get the cert from the update server while `is_responsive` uses the designated API of the update server. So ultimately you are achieving the same effect as `is_responsive`

lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1392,26 @@ def write_framework_identifier(cfg):
         framework_file.write(json.dumps(identifier))
 
 
+# ----------------------------------------------------------------------------
+def can_make_ip_request(rmt_ip):
+    """Check if the instance can make IPv4 or IPv6 requests."""
+    try:
+        __connection_check(rmt_ip, True)
```
rjschwei, line 1399: We do not need to make a connection check to determine if the client has IPv6 or IPv4 internet capability. Also as stated the connection check is basically the same as `is_responsive`

jesusbv: the method `is_responsive` checks that any RMT update server (either IPv6 or IPv4) are reachable,
we can not know if the instance has IPv6 enabled only with `is_responsive`  as the list of IPs checked on
`is_responsive` could be _only_ IPv4, it could return `True` but that would not confirm IPv6 enabled on
the client, so that method does not guarantee enablement of IPv4/IPv6 on the client 
Another example, if RMT provided IPv6 addresses and the client did not have IPv6 enabled, it would generate
an exception, pass because that's how the method is implemented and would try IPv4 after the IPv6 attempts,
let's say IPv4 succeeded, `is_responsive` would return `True` and we would still not know if the client has
IPv6 enabled
I understand the intend of `is_responsive` what I mean is it is benefitial to tell the user that they are trying 
something that they can't/do not have enabled, `is_responsive` can't tell that
I understand this PR/check is for `smt-ip` in the `force-new` path, but the [IPv4|IPv6] check should be in place
for other parts of this codebase, like [this check](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L268) that checks is always `True` even when the instance 
does not have IPv6 enabled 

IMO, we should have a way to check the instance/client has IPv4 or IPv6 enabled, whether using `is_responsive` 
or not I dont really mind, currently `is_responsive` does not guarantee whether the client has IPv4/IPv6 enabled or not

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
jesusbv: I get that there are system tools to do such checks, [this way](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L268) does not work, I had dismissed the checking of the metadata
for that info, as that would mean a different way per CSP

I followed the pattern of `is_responsive` or [has_ipv6_access](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L877) which is making a query and check/handle the response

I can explore other system options instead of querying the RMT update server, no problem

lib/cloudregister/registerutils.py
rjschwei: > 
> IMO, we should have a way to check the instance/client has IPv4 or IPv6 enabled, 

I am agree wit this idea. But implementing the concepts of `is_responsive` is not the way to achieve that goal.

That said when the user specifies update server information then we end up here https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L298 where I can insert:

```
smt = smt.SMT(region_smt_data)
if not smt.is_responsive():
    print('Provided update server IP address %s with "--smt-ip" is not reachable' % args.user_smt_ip')
    sys.exit(1)
```

So what you state in the description as a goal to reduce debugging time and provide an early exit condition can be achieved in 4 lines of code with any refactoring needed and we tell the user that the information provided cannot be used, whether it is IPv4 or IPv6 is immaterial.

So it appears that the mixing of "it would be nice to know if the instance has IPv4, IPV6 or both" and "let's have an early exit condition when the user gives us a bad ip-address" is creating problems with the clarity of implementation.

lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1407,19 @@ def write_framework_identifier(cfg):
         framework_file.write(json.dumps(identifier))
 
 
+# ----------------------------------------------------------------------------
+def instance_has_ip_enabled(rmt_ip):
```
rjschwei, line 1411: Lets call this `has_network_access_by_ipversion`

lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1407,19 @@ def write_framework_identifier(cfg):
...
+    """Check if the instance can make IPv4 or IPv6 requests."""
+    try:
+        socket.create_connection((rmt_ip, 443), timeout=2)
```
rjschwei, line 1414: Needs to be assigned to a variable and on success we need to call `close()`. While the temporary socket will be reaped on the Python side the Python garbage collector will not close the socket meaning it will be open on the server side because no request was received by the server. Meaning we will keep connections open unnecessarily and consuming resources on the server for no reason.

lib/cloudregister/registerutils.py
```diff
@@ -1407,6 +1407,19 @@ def write_framework_identifier(cfg):
...
+    except OSError as e:
+        # check socket connection produced
+        # Network is unreachable error
+        return e.errno != errno.ENETUNREACH
```
rjschwei, line 1418: So if we get an access error for whatever reason we return `True`, i.e. indicating that we can reach the given IP address. I think any error on the socket creation should return `False` there is no reason to believe that if the error is access or resource limitation, i.e. no more sockets available the problem would go away by the time we need the network path.

usr/sbin/registercloudguest
```diff
@@ -200,6 +200,15 @@ if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
         print(msg, file=sys.stderr)
         sys.exit(1)
 
+    if not utils.instance_has_ip_enabled(args.user_smt_ip):
+        error_message = (
+            'Connection error: Could not establish a connection to {ip}.'
+            'Please, make sure the instance has that IP format enabled '
```
rjschwei, line 206: s/the instance has that IP format enabled and working properly./the network configuration supports the provided IP address version./

usr/sbin/registercloudguest
```diff
@@ -214,6 +223,7 @@ if (args.email and not args.reg_code):
 
 time.sleep(int(args.delay_time))
 
+
```
rjschwei, line 226: ? why the additional whitespace?

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: The goal is to check if the instance is able to make IPv4 or IPv6 requests, not if we can reach the RMT server (which coud be down or not available for different reasons)
Agree that no more sockets available should not be treated as `OK you can make $IP_format requests`

usr/sbin/registercloudguest
jesusbv: fat fingers, overlooked that change, fixed

lib/cloudregister/registerutils.py
jesusbv: Updated to return `False` to any socket error, I still think we could fine grain it more
as the purpose is only if the instance can make $IP_format requests

jesusbv: Having the check for only that purpose would save us debugging and conversations and bugzilla entries when users start using IP4 and IPv6 combined or just IPv4

rjschwei: At present the implementation is used to specifically check for being able to access the update server IP provided by the user on the command line. This will catch for example if the user provides an IPv6 address but the network is not configured for IPv6. It will also fail if the network is configured for IPv6 but the target server is down. Which is fine in this specific use case but would not be OK in the general use case where we want to know if we can reach addresses over IPv4 and/or IPv6 to potentially reduce the current probing that we do when processing the information we get from the region server. Both can be accomplished with the following implementation:


```
def has_ipv4_access():
   """Check if we have IPv4 network configuration"""
  return has_network_access_by_ipversion('8.8.8.8')

def has_ipv6_access():
   """Check if we have IPv6 network configuration"""
   return has_network_access_by_ipversion('2001:4860:4860::8888')

def has_network_access_by_ipversion(server_ip):
    """Check if we can connect to the given server"""
    try:
    except OSError as e:
        logging.info('Network access error: "%s"', e)
        return False

    connection.close()
    return True
```

This makes the implementation of `has_network_access_by_ipversion` generic and can be used for the purposes of the command line argument verification where we want to know if a specific update server can be reached with the given IP address. And for the use case where we want to find out if IPv6 and/or IPv4 work we have the functions that use the Google DNS server to check if things work. The logging of a potential error message takes out the guess work why the connection may have failed.

jesusbv: Updated the method for the logging message, the rest of the changes should be a new/different PR after merging this one

lib/cloudregister/registerutils.py
```diff
@@ -1442,6 +1442,31 @@ def write_framework_identifier(cfg):
...
+def has_ipv6_access():
+    """Check if we have IPv6 network configuration"""
+    return has_network_access_by_ipversion('2001:4860:4860::8888')
+
+
+# ----------------------------------------------------------------------------
+def has_network_access_by_ipversion(server_ip):
+    """Check if we can connect to the given server"""
+    try:
+        rmt_connection = socket.create_connection((server_ip, 443), timeout=2)
```
rjschwei, line 1461: s/rmt_connection/connection/ we are getting any IP address as such the variable name should be generic and not imply a connection to a specific system.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: Updated

lib/cloudregister/registerutils.py
```diff
@@ -1442,6 +1442,31 @@ def write_framework_identifier(cfg):
...
+
+
+# ----------------------------------------------------------------------------
+def has_ipv6_access():
+    """Check if we have IPv6 network configuration"""
+    return has_network_access_by_ipversion('2001:4860:4860::8888')
+
+
+# ----------------------------------------------------------------------------
+def has_network_access_by_ipversion(server_ip):
```
rjschwei, line 1458: Last nit pick, s/ipversion/ipaddress/

rjschwei (approved):

PR #129, opened 2023-09-14
Py pckg cert chek

rjschwei:
Check if a Python package cert exists and warn of potential registration failure.


lib/cloudregister/registerutils.py
```diff
@@ -1034,6 +1035,14 @@ def import_smt_cert(smt):
...
         logging.error('SMT certificate import failed')
         return None
+    # Check if the underlying Python packages use certs that are built in
+    # bsc#1214801
+    for site_path in site.getsitepackages():
+        py_pack_certs = glob.glob(os.path.join(site_path, 'certifi/*.pem'))
+        if py_pack_certs:
+            logging.warning(
+                'SMT certificate imported, but "%s" exist. '
+                'This may lead to registration failure' % py_pack_certs)
```
jesusbv, line 1045: should we log a `str` better than `[]`, 
```python
% (', ').join(py_pack_certs)
```
?

jesusbv (approved):

PR #131, opened 2023-09-19
Rename registration_returncode to registration_failed

jesusbv:
More meaningful name

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -50,12 +50,12 @@ from requests.auth import HTTPBasicAuth
 # We have server certs that have no subject alt names
 # We have to check the server state API without certificate validation
 urllib3.disable_warnings()
-registration_failed = False
+registration_returncode = 0
 
 # ----------------------------------------------------------------------------
 def register_modules(extensions, products, registered=[]):
     """Register modules obeying dependencies"""
-    global registration_failed
+    global registration_returncode
     for extension in extensions:
         # If the extension is recommended it gets installed with the
         # baseproduct registration. No need to run another registration
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -50,12 +50,12 @@ from requests.auth import HTTPBasicAuth
 # We have server certs that have no subject alt names
 # We have to check the server state API without certificate validation
 urllib3.disable_warnings()
-registration_returncode = 0
+registration_failed = False
 
 # ----------------------------------------------------------------------------
 def register_modules(extensions, products, registered=[]):
     """Register modules obeying dependencies"""
-    global registration_returncode
+    global registration_failed
     for extension in extensions:
         # If the extension is recommended it gets installed with the
         # baseproduct registration. No need to run another registration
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -88,8 +88,8 @@ def register_modules(extensions, products, registered=[]):
                 stderr=subprocess.PIPE
             )
             res = p.communicate()
-            if p.returncode != 0:
-                registration_returncode = p.returncode
+            registration_failed = p.returncode != 0
```
smarlowucf, line 91: Nitpick: this is a bit difficult to read. Could make it clearer with a "ternary conditional".

```
registration_failed = False if p.returncode == 0 else True
```

Or maybe it could be simplified to:

```
registration_failed = True if p.returncode else False
```

If we know return code is always an integer and non zero values are all failures.

smarlowucf (approved):

usr/sbin/registercloudguest
jesusbv: While I addressed the suggestion, I wonder if it would be simpler 
```python
registration_failed = p.returncode
```
and have 0 and > 0 instead of False and True

usr/sbin/registercloudguest
```diff
@@ -50,12 +50,12 @@ from requests.auth import HTTPBasicAuth
 # We have server certs that have no subject alt names
 # We have to check the server state API without certificate validation
 urllib3.disable_warnings()
-registration_returncode = 0
+registration_failed = False
```
rjschwei, line 53: Is effectively the same.

usr/sbin/registercloudguest
```diff
@@ -50,12 +50,12 @@ from requests.auth import HTTPBasicAuth
...
 
 # ----------------------------------------------------------------------------
 def register_modules(extensions, products, registered=[]):
     """Register modules obeying dependencies"""
-    global registration_returncode
+    global registration_failed
```
rjschwei, line 58: Effectively the same, superfluous code change.

rjschwei (changes requested):
Not going to go along with this anymore. 90% of this is cosmetic and personal preference as such obfuscating the actual fix for the issue. Fix the issue, be done

jesusbv:
> Not going to go along with this anymore. 90% of this is cosmetic and personal preference as such obfuscating the actual fix for the issue. Fix the issue, be done

The issue is fixed in a different PR

rjschwei:
Ultimately this breaks the original intend. The return code of SUSEConnect is meaningful which is why this was not treated as True/False setup to begin with. Granted there may be a bug in that the exit code is not propagated properly. However, removing the intend is not something we should do, it makes debugging more difficult.

jesusbv:
> Ultimately this breaks the original intend. The return code of SUSEConnect is meaningful which is why this was not treated as True/False setup to begin with. 

Correct, however before #130 (which will use the `SUSEConnect` returncode as exit code of the call of `registercloudguest`) 
`returncode` was not used at all

Now we are going to use that returncode, 
we do not need to rename the variable, or do any other change
so [this if](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/cbe076cfaad62cc49d7019a8318b68986e52fba2/usr/sbin/registercloudguest#L534) **could/should** be
`if registration_returncode:`

rjschwei (approved):

PR #132, opened 2023-10-09
Handle extra extensions or modules when registering

jesusbv:
If the instance has extra modules activated before running registercloudguest command, we should register the instance and inform the user about those extra modules not installed and how to install them

Fixes bsc#1215198

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -103,7 +103,7 @@ def register_modules(extensions, products, registered=[], failed=[]):
                     # 67: Server responded with error: see log output
                     failed.append(triplet)
                     # the registration should not be considered failed
-                    # because of extra modules installed
+                    # because of extra modules registered
                     registration_returncode = 0
                 else:
                     logging.error('\tRegistration failed: %s' % error_message)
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -540,6 +556,15 @@ if registration_returncode != 0:
 else:
     print('Registration succeeded')
 
+if failed_extensions:
+    print(
+        'There are products that were not installed because they need '
+        'an additional registration code, to install them please call'
```
rjschwei, line 562: s/installed/registered/g/

rjschwei (changes requested):

usr/sbin/registercloudguest
smarlowucf, line 562: Not sure if "call" is a universal term here. Suggest something like "run the following command:".

jesusbv: Done

jesusbv: Done

smarlowucf (approved):

rjschwei (approved):

PR #134, opened 2023-11-28
Missing ip exception

jesusbv:
Log IP from cert_url

This Fixes bsc#1217583


lib/cloudregister/smt.py
```diff
@@ -250,10 +254,8 @@ def __is_cert_valid(self, cert):
     def __request_cert(self):
         """Request the cert from the SMT server and return the request"""
         cert_res = None
-        attempts = 0
         retries = 3
-        while attempts < retries:
-            attempts += 1
+        for attempts in range(retries):
```
rjschwei, line 258: What does changing the loop construct have to do with the fix?

rjschwei (changes requested):

lib/cloudregister/smt.py
jesusbv: I will revert that change

jesusbv: Removed that commit with that change, tests added

lib/cloudregister/smt.py
```diff
@@ -17,7 +17,9 @@
 import ipaddress
 import logging
 import requests
+import os
 
+from urllib.parse import urlparse
```
rjschwei, line 22: Not really a need for another dependency even if the dependency is part of CPyton standard library. We know all the information this gives us already.

lib/cloudregister/smt.py
```diff
@@ -195,9 +197,9 @@ def write_cert(self, target_dir):
             ipv6 = self.get_ipv6()
             if ipv6:
                 cert_id = ipv6.replace(':', '_')
-        ca_file_path = (
-            target_dir +
-            '/registration_server_%s.pem' % cert_id
+        ca_file_path = os.path.join(
+            target_dir,
+            'registration_server_%s.pem' % cert_id
```
rjschwei, line 202: Style and personal preference, but fine albeit an unnecessary change.

lib/cloudregister/smt.py
```diff
@@ -254,6 +259,7 @@ def __request_cert(self):
         retries = 3
         while attempts < retries:
             attempts += 1
+            print('eeee')
```
rjschwei, line 262: Debug print in PR.....

lib/cloudregister/smt.py
```diff
@@ -267,6 +273,9 @@ def __request_cert(self):
                             'Attempt %s with %s of %s' % (
                                 attempts, cert_name, retries)
                         )
-                        logging.error('Server %s is unreachable' % ip)
+                        parsed_cert_url = urlparse(cert_url)
+                        logging.error(
+                            'Server %s is unreachable' % parsed_cert_url.netloc
+                        )
```
rjschwei, line 279: We already know the IP, so no real need to have it parsed via yet another function call.

tests/test_smt.py
```diff
@@ -132,6 +132,52 @@ def test_get_cert_invalid_cert(mock_cert_pull, mock_logging):
...
+    smt = SMT(etree.fromstring(smt_data_ipv46))
+    mock_os_path_join.return_value = '/tmp/registration_server_fc00__1.pem'
+    assert not smt.get_cert()
+    assert mock_logging.error.called
+    mock_logging.error.assert_called_with('Server 192.168.1.1 is unreachable')
+    assert not smt.write_cert('tmp')
+    os.remove('/tmp/registration_server_fc00__1.pem')
+    msg = 'write() argument must be str, not None'
+    mock_logging.error.assert_called_with(msg)
+
```
rjschwei, line 155: This does not sufficiently test the `write_cert()` method. Otherwise a follow on problem from the fix for the original problem would have been noticed. The cert on a client that has IPv4 only but receives a config for IPv4 and Ipv6 update server gets witten with the IPv6 address. While ultimately immaterial for cert validation as the name doesn't matter it would break SUMa. When adding tests the effort to test comprehensively is relatively small. This fans out into 4 tests:

1. Update server with IPv4 only
2. Update server with IPv6 only
3. Update server with dual stack config
4. The exception path in `write_cert()`

tests/test_smt.py
```diff
@@ -132,6 +132,52 @@ def test_get_cert_invalid_cert(mock_cert_pull, mock_logging):
...
+    response.status_code = 200
+    response.text = 'Not a cert'
+    mock_request_get.return_value = response
+    smt = SMT(etree.fromstring(smt_data_ipv46))
+    mock_os_path_join.return_value = '/tmp/registration_server_fc00__1.pem'
+    mock_get_cert.return_value = 'what a cert'
+    assert smt.write_cert('tmp') == '/tmp/registration_server_fc00__1.pem'
+    os.remove('/tmp/registration_server_fc00__1.pem')
+    assert not mock_logging.called
+
```
rjschwei, line 179: This test is dealing with unnecessary indirection. Ultimately the goal is to catch the exception path in `__request_cert()` the most direct path to test this is by using the `get_cert()` public method rather than a method `write_cert()` that in turn calls `get_cert()`.  Also this would need to fan out into 2 tests as there are two paths, one where IPv4 is not reachable and 1 where IPv6 is not reachable.

rjschwei (changes requested):

lib/cloudregister/smt.py
jesusbv: Good catch

lib/cloudregister/smt.py
jesusbv: True

rjschwei:
Superseded by #137 
 
Unfortunately I screwed something up when rebasing  back to master after #136 was merged. AS such it was best to start over.

PR #135, opened 2023-12-11
Unhandeled Exception with IPv4 only client with IPv6 update server config

rjschwei:
When the region server returns an update server configuration with IPv4 and IPv6 but the client only has IPv4 capability an unhandeld exception is triggered due to an uninitialized variable.

Once the unitialized variable issue is fixed, the code will write the server cert to the trust anchors only with the IPv6 name as reference. This is confusing. The updated code writes the cert with botht the IPv4 and IPv6 names.

Add tests for the updates smt.py implementation.

### DIFF
--- a/lib/cloudregister/smt.py
+++ b/lib/cloudregister/smt.py
@@ -19,7 +19,6 @@ import logging
 import requests
 import os
 
-from urllib.parse import urlparse
 from M2Crypto import X509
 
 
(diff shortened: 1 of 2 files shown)


rjschwei:
This addresses the issues raised in #134

jesusbv (approved):
LGTM

PR #137, opened 2023-12-20
Update to 10.1.5 follow up fixes to IPv6

rjschwei:
When the region server returns an update server configuration with IPv4 and IPv6 but the client only has IPv4 capability an unhandeld exception is triggered due to an uninitialized variable.

Once the unitialized variable issue is fixed, the code will write the server cert to the trust anchors only with the IPv6 name as reference. This is confusing. The updated code writes the cert with both the IPv4 and IPv6 names.

In an IPv6 only setup no metadata could be retieved via the EC2 plugin. Try to access the IPv6 endpoint for IMDS when IPv4 access fails.

Add tests


rjschwei:
Merging the request as the bits have been reviewed already as part of #134. Since something went wrong during rebase of #134 after #136 was merged this new branch was created.

Code has been tested on 3 15 SP5 instances in a dual stack, IPv4 only and IPv6 only setup.

PR #138, opened 2023-12-20
Fix case when switching license from any not empty non SLES_BYOS to SLES_BYOS without cache file

jesusbv:
When switching from SLES to BYOS and no cache file exists, the license changes to BYOS and no other action is done, instance/system is still registered

This change fixes that scenario

### DIFF
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -98,8 +98,10 @@ service_name = 'guestregister'
 license_type = get_license_type()
 
 if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    # Registered to SCC
     license_has_changed = has_license_changed(license_type)
     if license_has_changed in [False]:
+        # System is still BYOS, nothing to do
         sys.exit(0)
     if license_has_changed:
         update_license_cache(license_type)
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -37,6 +37,7 @@ def has_license_changed(license_type):
             return license_type != old_license
     except FileNotFoundError:
         update_license_cache(license_type)
+        raise
 
 
 def get_license_type():
@@ -97,20 +98,33 @@ utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
 
-if ('BYOS' in license_type and has_license_changed(license_type) and
-    not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    try:
+        if not has_license_changed(license_type):
+            sys.exit(0)
+        update_license_cache(license_type)
+    except FileNotFoundError:
+        # if cache file did not exist, it got created
+        # only thing left to do is to de-register
+        pass
     run_command(['registercloudguest', '--clean'])
     run_command(['systemctl', 'disable', service_name])
-    update_license_cache(license_type)
 elif license_type and 'BYOS' not in license_type:
     # if no license type, system could be a BYOS without AHB
     # or PAYG without AHB
     update_server = utils.get_smt()
     if update_server:
-        if (not has_license_changed(license_type) and
-            utils.is_registered(update_server.get_FQDN())):
-            sys.exit(0)
-        update_license_cache(license_type)
+        try:
+            license_has_changed = has_license_changed(license_type)
+            if (not license_has_changed and
+                utils.is_registered(update_server.get_FQDN())):
+                sys.exit(0)
+            if license_has_changed:
+                update_license_cache(license_type)
+        except FileNotFoundError:
+            # if cache file did not exit, it got created
+            # only thing left to do is to register
+            pass
         update_server = None
     if not update_server:
         run_command(['registercloudguest', '--force-new'])
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -154,15 +154,16 @@ elif license_type and 'BYOS' not in license_type:
         if (not license_has_changed and
             utils.is_registered(update_server.get_FQDN())):
             # covers cases 7, 15, 16, 17, 18
+            if utils.is_scc_connected():
+                # cover cases 11
+                run_command(['registercloudguest', '--force-new'])
+                run_command(['systemctl', 'enable', service_name])
             sys.exit(0)
-        if license_has_changed:
-            # cases 12
-            update_license_cache(license_type)
         update_server = None
     if not update_server:
-        # cover cases 1, 2, 8
+        # cover cases 1, 2, 8, 12
         if has_license_changed(license_type):
-            # cover cases 2, 8
+            # cover cases 2, 8, 12
             update_license_cache(license_type)
         run_command(['registercloudguest', '--force-new'])
         run_command(['systemctl', 'enable', service_name])
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -12,6 +12,21 @@
 # Lesser General Public License for more details.
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
+
+"""
+Check the different states of the license type:
+    START   LICENSE CHANGED TO     CACHE   ACTION
+    BYOS    SLES                   N/A     Register
+    BYOS    SLES                   BYOS    Register
+    BYOS    BYOS                   N/A     Do nothing
+    BYOS    BYOS                   BYOS    Do nothing
+
+    SLES    BYOS                   N/A     De-register
+    SLES    BYOS                   SLES    De-register
+    SLES    SLES                   N/A     Do nothing
+    SLES    SLES                   SLES    Do nothing
+"""
+
 import logging
 import os
 import requests
@@ -19,6 +34,7 @@ import subprocess
 import sys
 
 import cloudregister.registerutils as utils
+
 
 CACHE_LICENSE_PATH = os.path.join(utils.get_state_dir(), 'cached_license')
 
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -127,18 +127,20 @@ service_name = 'guestregister'
 license_type = get_license_type()
 update_server = utils.get_smt()
 
-if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy() and
+    not utils.is_scc_connected()
+):
     license_has_changed = has_license_changed(license_type)
-    if (
-        license_has_changed or
+    if (license_has_changed or
         (update_server and utils.is_registered(update_server.get_FQDN()))
     ):
         # either license has changed or
-        # system was previously registered
+        # system was previously registered to update infrastructure
+        # and neither SCC nor RMT as an SCC proxy
         if license_has_changed:
             update_license_cache(license_type)
-    run_command(['registercloudguest', '--clean'])
-    run_command(['systemctl', 'disable', service_name])
+        run_command(['registercloudguest', '--clean'])
+        run_command(['systemctl', 'disable', service_name])
 elif license_type and 'BYOS' not in license_type:
     # if no license type, system could be a BYOS without AHB
     # or PAYG without AHB
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -100,7 +100,7 @@ license_type = get_license_type()
 if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
     # Registered to SCC
     license_has_changed = has_license_changed(license_type)
-    if license_has_changed in [False]:
+    if license_has_changed == False:
         # System is still BYOS, nothing to do
         sys.exit(0)
     if license_has_changed:
@@ -113,7 +113,7 @@ elif license_type and 'BYOS' not in license_type:
     update_server = utils.get_smt()
     if update_server:
         license_has_changed = has_license_changed(license_type)
-        if (license_has_changed in [False] and
+        if (license_has_changed == False and
             utils.is_registered(update_server.get_FQDN())):
             sys.exit(0)
         if license_has_changed:
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -16,28 +16,29 @@
 """
 Check the different states of the license type:
     START   LICENSE CHANGED TO     CACHE   ACTION
-    BYOS-1  SLES                   N/A     Register
-    BYOS-1  SLES                   BYOS    Register
-    BYOS-1  BYOS-1                 N/A     Do nothing
+    BYOS-1  SLES                   No      Register - Update Infra
+    BYOS-1  SLES                   BYOS    Register - Update Infra
+    BYOS-1  BYOS-1                 No      Do nothing
     BYOS-1  BYOS-1                 BYOS    Do nothing
-    BYOS-1  BYOS-2                 N/A     Do nothing
+    BYOS-1  BYOS-2                 No      Do nothing
     BYOS-1  BYOS-2                 BYOS    Do nothing
 
-    BYOS-2  SLES                   N/A     Do nothing
-    BYOS-2  SLES                   BYOS    Re-register
-    BYOS-2  BYOS-1                 N/A     De-register
-    BYOS-2  BYOS-1                 BYOS    De-register
+    BYOS-2  SLES                   No      Do nothing
+    BYOS-2  SLES                   BYOS    Re-register - Update Infra
+    BYOS-2  BYOS-1                 No      De-register - Update Infra
+    BYOS-2  BYOS-1                 BYOS    De-register - Update Infra
 
-    SLES    BYOS-1                 N/A     De-register
-    SLES    BYOS-1                 N/A     De-register
-    SLES    BYOS-2                 N/A     Do nothing
+    SLES    BYOS-1                 No      De-register - Update Infra
+    SLES    BYOS-1                 No      De-register - Update Infra
+    SLES    BYOS-2                 No      Do nothing
     SLES    BYOS-2                 SLES    Do nothing
-    SLES    SLES                   N/A     Do nothing
+    SLES    SLES                   No      Do nothing
     SLES    SLES                   SLES    Do nothing
 
 BYOS-1 -> BYOS unregistered
 BYOS-2 -> BYOS registered
 NOTE: When an instance starts the license_type value from IMDS is none (empty)
+START represents the state of the license type before running AHB
 """
 
 import logging
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -37,7 +37,7 @@ def has_license_changed(license_type):
             return license_type != old_license
     except FileNotFoundError:
         update_license_cache(license_type)
-
+        return False
 
 def get_license_type():
     proxies = {
@@ -96,21 +96,23 @@ def run_command(command):
 utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
+update_server = utils.get_smt()
 
 if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
-    # Registered to SCC
     license_has_changed = has_license_changed(license_type)
-    if license_has_changed == False:
-        # System is still BYOS, nothing to do
-        sys.exit(0)
-    if license_has_changed:
-        update_license_cache(license_type)
+    if (
+        license_has_changed or
+        (update_server and utils.is_registered(update_server.get_FQDN()))
+    ):
+        # either license has changed or
+        # system was previously registered
+        if license_has_changed:
+            update_license_cache(license_type)
     run_command(['registercloudguest', '--clean'])
     run_command(['systemctl', 'disable', service_name])
 elif license_type and 'BYOS' not in license_type:
     # if no license type, system could be a BYOS without AHB
     # or PAYG without AHB
-    update_server = utils.get_smt()
     if update_server:
         license_has_changed = has_license_changed(license_type)
         if (license_has_changed == False and
--- a/usr/sbin/regionsrv-enabler-azure
+++ b/usr/sbin/regionsrv-enabler-azure
@@ -15,30 +15,35 @@
 
 """
 Check the different states of the license type:
+START -> The state of the instance when this code is executed
+         This code is executed by a timer
+
     START   LICENSE CHANGED TO     CACHE   ACTION
-    BYOS-1  PAYG                   No      Register - Update Infra
-    BYOS-1  PAYG                   BYOS    Register - Update Infra
-    BYOS-1  BYOS-1                 No      Do nothing
-    BYOS-1  BYOS-1                 BYOS    Do nothing
-    BYOS-1  BYOS-2                 No      Do nothing
-    BYOS-1  BYOS-2                 BYOS    Do nothing
+1   BYOS-1  PAYG                   No      Register - Update Infra
+2   BYOS-1  PAYG                   BYOS    Register - Update Infra
+3   BYOS-1  BYOS-1                 No      Do nothing
+4   BYOS-1  BYOS-1                 BYOS    Do nothing
+5   BYOS-1  BYOS-2                 No      Do nothing
+6   BYOS-1  BYOS-2                 BYOS    Do nothing
 
-    BYOS-2  PAYG                   No      Do nothing
-    BYOS-2  PAYG                   BYOS    Re-register - Update Infra
-    BYOS-2  BYOS-1                 No      De-register - Update Infra
-    BYOS-2  BYOS-1                 BYOS    De-register - Update Infra
+7   BYOS-2  PAYG                   No      Do nothing
+8   BYOS-2  PAYG                   BYOS    Re-register - Update Infra
+9   BYOS-2  BYOS-1                 No      De-register - Update Infra
+10  BYOS-2  BYOS-1                 BYOS    De-register - Update Infra
+11  BYOS-3  PAYG                   No      No de-register from SCC, register with update infrastructure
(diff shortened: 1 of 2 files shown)


smarlowucf:
I don't know the code enough to give a good review here. Also, is it intended to be s/SLES to BYOS/PAYG to BYOS?

jesusbv:
> Also, is it intended to be s/SLES to BYOS/PAYG to BYOS?

Well, it is a license change from `SLES` to `SLES_BYOS`
I will edit the title of the PR to clarify it

jgleissner:
I'm trying to understand what the license cache is. So the pre-PR version simply updates the cache when the file is missing, which is kinda what you expect when it's trying to access a non-existent cache. But it didn't return anything in that case, from a function that is clearly supposed to, so that seems off. Plus it sets the license type to what it's supposed to compare it with, which also seems off.

It seems to me that the license cache isn't actually a cache (== stored data that is time-consuming to compute or retrieve) but actually the registration information (I vaguely recall we've had this conversation before). So that would mean, missing cache means it's assumed the system is not registered?

jesusbv:
> I'm trying to understand what the license cache is. So the pre-PR version simply updates the cache when the file is missing, which is kinda what you expect when it's trying to access a non-existent cache. But it didn't return anything in that case, from a function that is clearly supposed to, so that seems off. Plus it sets the license type to what it's supposed to compare it with, which also seems off.
> 
So, pre-PR, the cache gets checked and if the file didn't exist, it gets created and the check method returned `None`
meaning **license has not changed**, whenever the check returned `True` cache file gets updated outside the check

This leads to an scenario where switching from `SLES` (or any `PAYG` license) to `SLES_BYOS` and no cache file, the instance is registered against the infrastructure and benefiting from that while being `BYOS` 

> It seems to me that the license cache isn't actually a cache (== stored data that is time-consuming to compute or retrieve) but actually the registration information (I vaguely recall we've had this conversation before). 

I'm not sure I understand, the cache file stores the license value only
> So that would mean, missing cache means it's assumed the system is not registered?

Missing the cache or not missing the cache does not affect registration per se, it affects registration when switching licenses. For example, if the instance is registered and license switched to `SLES_BYOS` (and cache had `SLES` for the sake of the example) that check and code would take care of de-registering the instance

Another approach to fix this issue is to check for `BYOS` license in the license check and de-register + disable the service
```python
except FileNotFoundError:
   if 'BYOS' in license_type:
       run_command(['systemctl', 'disable', 'guestregister'])
   update_license_cache(license_type)
   # Notice we would keep returning None
```
This would put some responsibility into `has_license_changed` which IMHO does not belong
but we may be open to that trade-off to fix the scenario of "`BYOS` benefiting from infra", again
this would add a side-effect into the `check license` code where one would expect the method to
check the license or do things license related, not to do things like de-registering the instance for example

jgleissner:
A function called `has_license_changed` should not return `None`, but only `True`, `False`, or throw an exception if it can't tell.

Generally how I would expect it to work is to populate the cache when trying to access data in the cache that is not there. So the question is, can it figure out the current license type from the active registration (not by re-registering)? If not, it's not actually a cache.

jgleissner (approved):
LGTM. I'd still suggest moving the stored license type into `/var/lib` and not calling it cache, as it really persists the last known license type and rather than caching metadata, which got me confused in the first place.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -97,20 +98,33 @@ utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
 
-if ('BYOS' in license_type and has_license_changed(license_type) and
-    not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    try:
+        if not has_license_changed(license_type):
```
rjschwei, line 103: The `has_license_changed()` function,  https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/regionsrv-enabler-azure#L32, does not throw, there is no reason to have the call wrapped in a `try - except` block.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -97,20 +98,33 @@ utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
 
-if ('BYOS' in license_type and has_license_changed(license_type) and
-    not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
```
rjschwei, line 101: I think it would be great to have some contextual comments in the code. Looking at it all a while after the initial implementation I think those comments would greatly help. For this condition the comment could be `# Registered to SCC`

usr/sbin/regionsrv-enabler-azure
rjschwei: We could add a `#Still BYOS nothing to do` comment here

usr/sbin/regionsrv-enabler-azure
```diff
@@ -97,20 +98,33 @@ utils.start_logging()
...
 elif license_type and 'BYOS' not in license_type:
     # if no license type, system could be a BYOS without AHB
     # or PAYG without AHB
     update_server = utils.get_smt()
     if update_server:
-        if (not has_license_changed(license_type) and
-            utils.is_registered(update_server.get_FQDN())):
-            sys.exit(0)
-        update_license_cache(license_type)
+        try:
```
rjschwei, line 117: ditto,`has_license_changed()` does not throw, no `try-except` block needed.

rjschwei (changes requested):

usr/sbin/regionsrv-enabler-azure
jesusbv: > The `has_license_changed()` function, https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/regionsrv-enabler-azure#L32, does not throw, there is no reason to have the call wrapped in a `try - except` block.

The re raise of the exception is added in this [PR](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/a8d053d0ba259b57473922d1339f231845049153/usr/sbin/regionsrv-enabler-azure#L40) ,
as the method should return either `True` or `False` but on that case was returning `None` when it actually 
can not know if it did change

 We could leave the method returning `None`, the approach would be very similar

usr/sbin/regionsrv-enabler-azure
jesusbv: I'll add them

jesusbv: Done

usr/sbin/regionsrv-enabler-azure
jesusbv: No exception re raised, using `None` value as different to `False`

usr/sbin/regionsrv-enabler-azure
```diff
@@ -97,20 +97,27 @@ utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
 
-if ('BYOS' in license_type and has_license_changed(license_type) and
-    not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    # Registered to SCC
+    license_has_changed = has_license_changed(license_type)
+    if license_has_changed == False:
```
rjschwei, line 103: If we are using `False` here then `has_license_changed()`  needs get a `return False` at the end of the function. 
```
>>> None == False
False
```
This code will fail when there is no state file because `has_license_changed()` will return `None`. Alternatively stick with `if not has_license_changed`

If there is no state file, i.e. `CACHE_LICENSE_PATH` does not exist then the system has not ever had a license setting in the metadata. If there is no license in the metadata we should not ever take any action because we cannot determine whether or not the instance is PAYG or BYOS only from looking at the metadata delivered by IMDS.

Ultimately this code is a state machine. It is probably good to put the states we need to handle into a doc string at the top of the file. That way it is easier in the future to correlate the implemented logic with the documented states we need to handle.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -97,20 +97,27 @@ utils.start_logging()
 service_name = 'guestregister'
 license_type = get_license_type()
 
-if ('BYOS' in license_type and has_license_changed(license_type) and
-    not utils.uses_rmt_as_scc_proxy()):
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    # Registered to SCC
+    license_has_changed = has_license_changed(license_type)
```
rjschwei, line 102: The assignment can be moved outside the if condition then we do not have to make a second function call in the `else:` clause. Meaning in the case where there is no state file we also save ourselves one stack unwind do to the exception.

rjschwei (changes requested):

usr/sbin/regionsrv-enabler-azure
jesusbv: > If we are using `False` here then `has_license_changed()` needs get a `return False` at the end of the function.
> 
> ```
> >>> None == False
> False
> ```
> 
> This code will fail when there is no state file because `has_license_changed()` will return `None`. Alternatively stick with `if not has_license_changed`
> 
> If there is no state file, i.e. `CACHE_LICENSE_PATH` does not exist then the system has not ever had a license setting in the metadata. If there is no license in the metadata we should not ever take any action because we cannot determine whether or not the instance is PAYG or BYOS only from looking at the metadata delivered by IMDS.

Fair point, the use of the re raise of the exception or `None` to determine if any de-registering action should be taken
when `BYOS` could have a better approach

> Ultimately this code is a state machine. It is probably good to put the states we need to handle into a doc string at the top of the file. That way it is easier in the future to correlate the implemented logic with the documented states we need to handle.

Will do

usr/sbin/regionsrv-enabler-azure
jesusbv: I do not agree, 
- if license type includes BYOS and not RMT as SCC proxy -> we check the license
- else if it does not include BYOS and there is an update server -> we do not check the license

Unless I'm missing a path or workflow of the code (in the current PR), we do not always check the license ?
If we do not mind doing it outside even in the case where license type does not include BYOS
and system has an SMT server, fine by me

There is another option, moving the check from
```python
if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
    license_has_changed = has_license_changed(license_type)
    if (
        license_has_changed or
        (update_server and utils.is_registered(update_server.get_FQDN()))
    ):
        ...
```
to
```python
license_has_changed = has_license_changed(license_type)
if ('BYOS' in license_type and
    not utils.uses_rmt_as_scc_proxy() and
    (license_has_changed or 
    (update_server and utils.is_registered(update_server.get_FQDN()))
):
    ...
```

usr/sbin/regionsrv-enabler-azure
jesusbv: Done

usr/sbin/regionsrv-enabler-azure
rjschwei: Correct I missed the nested ` if update_server:` in the `else:` condition.

rjschwei (changes requested):

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,34 @@
 # Lesser General Public License for more details.
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
+
+"""
+Check the different states of the license type:
+    START   LICENSE CHANGED TO     CACHE   ACTION
```
rjschwei, line 18: We'll need to clarify what `START` means. `BYOS-2` at instance start is only possible if a customer  passes user data and makes cloud-init run the registration process via the passed data. This process is probably rare but worth to be considered. In any event if the "register via extra setup" is considered as `BYOS-2` it should be in the comment.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,34 @@
...
+    BYOS-1  SLES                   N/A     Register
+    BYOS-1  SLES                   BYOS    Register
+    BYOS-1  BYOS-1                 N/A     Do nothing
+    BYOS-1  BYOS-1                 BYOS    Do nothing
+    BYOS-1  BYOS-2                 N/A     Do nothing
+    BYOS-1  BYOS-2                 BYOS    Do nothing
+
+    BYOS-2  SLES                   N/A     Do nothing
+    BYOS-2  SLES                   BYOS    Re-register
+    BYOS-2  BYOS-1                 N/A     De-register
```
rjschwei, line 28: What are the actions taken by the user here? I do not understand this state transition in the context of this code. When a BYOS instance is registered, either to SCC or the update infrastructure then in order to transition from the registered state to the unregistered state `BYOS-2` to `BYOS-1` the user has to run `registercloudguest --clean`  meaning this code when it runs again will find a cache and no registration.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,34 @@
...
+Check the different states of the license type:
+    START   LICENSE CHANGED TO     CACHE   ACTION
+    BYOS-1  SLES                   N/A     Register
+    BYOS-1  SLES                   BYOS    Register
+    BYOS-1  BYOS-1                 N/A     Do nothing
+    BYOS-1  BYOS-1                 BYOS    Do nothing
+    BYOS-1  BYOS-2                 N/A     Do nothing
+    BYOS-1  BYOS-2                 BYOS    Do nothing
+
+    BYOS-2  SLES                   N/A     Do nothing
```
rjschwei, line 26: Does `SLES` mean `PAYG`? These state transitions are not product specific but billing model specific, i.e. they apply equally to SLES For SAP and SLE HPC

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,34 @@
 # Lesser General Public License for more details.
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
+
+"""
+Check the different states of the license type:
+    START   LICENSE CHANGED TO     CACHE   ACTION
+    BYOS-1  SLES                   N/A     Register
```
rjschwei, line 19: They way I understand this this would read as:

"""
An unregistered BYOS instance that gets moved to PAYG via AHB and has no cahce file needs to get registered to the update infrastructure
"""

If this is correct I suggest we express this as

`
BYOS-1   PAYG      No  Register - Update Infra
`

usr/sbin/regionsrv-enabler-azure
rjschwei: The diff is not showing the contextual comment added per the previous comment. For this condition I suggest something along the lines of:
`# Instance has been set to BYOS via AHB but has not been registered to the update infrastructure with a registration code`

usr/sbin/regionsrv-enabler-azure
```diff
@@ -96,21 +125,30 @@ def run_command(command):
...
+if ('BYOS' in license_type and not utils.uses_rmt_as_scc_proxy()):
+    license_has_changed = has_license_changed(license_type)
+    if (
+        license_has_changed or
+        (update_server and utils.is_registered(update_server.get_FQDN()))
+    ):
+        # either license has changed or
+        # system was previously registered
+        if license_has_changed:
+            update_license_cache(license_type)
```
rjschwei, line 139: Running `--clean` here indiscriminately will bring a bug we fixed previously. The scenario is a user starting a BYOS instance, then registering it to SCC then using AHB to set he instance as BYOS.

In this case we will detect `BYOS` in the metadata and since the system is registered to SCC it will not use the update infrastructure as a proxy. Meaning we enter the code branch at line 130 and therefore we will wipe out the users SCC registration, which we should not do.

rjschwei (changes requested):

usr/sbin/regionsrv-enabler-azure
jesusbv: True that the user has to run `registercloudguest --clean` command, but the code would check if license is BYOS,
which it would, it would check if it uses RMT as SCC proxy, which it wouldn't, then re-run the same command and
run the command to disable `guestregister`, no cache would be found on that scenario

If it did run again, after that scenario, yes there would be a cache saying `SLES_BYOS` on an unregistered system

usr/sbin/regionsrv-enabler-azure
jesusbv: Yes, `SLES` represents all possible PAYG possible states, meaning `SES_SAP` for example

usr/sbin/regionsrv-enabler-azure
jesusbv: Yes, that is the intend

I will update the wording

usr/sbin/regionsrv-enabler-azure
jesusbv: but this states can happen at any time, correct ? if customer removed the cache file, and run the AHB command then this `BYOS-2` would be possible

usr/sbin/regionsrv-enabler-azure
jesusbv: I think I fixed both scenarios (the one we want to fix on this PR and the wonrgly de-registration if system is registered to SCC) in this commit [2bec9b4](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/138/commits/2bec9b4871e0bf41449a9930212b43368bf33cd0)

usr/sbin/regionsrv-enabler-azure
rjschwei: Which is why `START` as a term needs to be defined. Since this is a cloud-instance it is arguably somewhat "natural" to associate the word "start" with "the start of an instance". As you point out if a user messes with the cache, which they shouldn't but can, and we need to handle that case, then a certain condition can be triggered. This is why I pointed out that we need to define what `START` means.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,35 @@
 # Lesser General Public License for more details.
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
+
+"""
+Check the different states of the license type:
+    START   LICENSE CHANGED TO     CACHE   ACTION
+    BYOS-1  SLES                   No      Register - Update Infra
+    BYOS-1  SLES                   BYOS    Register - Update Infra
```
rjschwei, line 20: I think `SLES` should be `PAYG` the state transitions are product independent and apply equally to SLES For SAP and SLE HPC. Mixing a license state with product nomenclature is confusing.

usr/sbin/regionsrv-enabler-azure
rjschwei: Agreed with the addition of the scc check in the if condition on line 131 the issue I pointed out is resolved.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -96,21 +126,34 @@ def run_command(command):
...
+    # system license type has been set to BYOS via AHB but
+    # neither has been registered to the update infrastructure with a reg code
+    # nor to SCC
+    license_has_changed = has_license_changed(license_type)
+    if (license_has_changed or
+        (update_server and utils.is_registered(update_server.get_FQDN()))
+    ):
+        # either license has changed or
+        # system was previously registered to update infrastructure
+        if license_has_changed:
```
rjschwei, line 143: This condition is tested twice in consecutive conditionals, on line 138 and then again here. Can be changed to (starting at line 137)

```
if has_license_changed(license_type):
    update_license_cache(license_type)
    if update_server and utils.is_registered(update_server.get_FQDN()):
```

to avoid checking the same condition in consecutive instructions and make the flow a little easier to follow.

jesusbv: We should not do that, the de-registration must happen even if license hasnt changed
I will think a better way to check the conditions

usr/sbin/regionsrv-enabler-azure
jesusbv: I defined `START` [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/432d94a48cad7aebd71382edf9db5b64cbfe8e2a/usr/sbin/regionsrv-enabler-azure#L41)
```
START represents the state of the license type before running AHB
```

usr/sbin/regionsrv-enabler-azure
jesusbv: I think you meant outside the if
```python
if has_license_changed(license_type):
    update_license_cache(license_type)
if update_server and utils.is_registered(update_server.get_FQDN()):
```

usr/sbin/regionsrv-enabler-azure
rjschwei: Missed that, so let's have the definitions above the table.

That said, AHB is the name for a feature in Azure, it cannot be run. Maybe we can define `START` as follows:

"The state of the instance when this code is executed. This code is executed by a timer"

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,35 @@
...
+
+    PAYG    BYOS-1                 No      De-register - Update Infra
+    PAYG    BYOS-1                 No      De-register - Update Infra
+    PAYG    BYOS-2                 No      Do nothing
+    PAYG    BYOS-2                 PAYG    Do nothing
+    PAYG    PAYG                   No      Do nothing
+    PAYG    PAYG                   PAYG    Do nothing
+
+BYOS-1 -> BYOS unregistered
+BYOS-2 -> BYOS registered
```
rjschwei, line 39: We need to consider the a `BYOS-3` case where `BYOS-3` is defined as registered to SCC and we need to handle that transition for example

`BYOS-3  PAYG  No  dergister from SCC register with update infrastructure.`

is one of the state transitions we need to handle.

It may also be a good idea to number the state transitions and then in the implementation reference the numbers the various if conditions handle.

Let's focus on getting the state transitions captured and then take another look at the implementation.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,40 @@
...
+START -> The state of the instance when this code is executed
+         This code is executed by a timer
+
+    START   LICENSE CHANGED TO     CACHE   ACTION
+1   BYOS-1  PAYG                   No      Register - Update Infra
+2   BYOS-1  PAYG                   BYOS    Register - Update Infra
+3   BYOS-1  BYOS-1                 No      Do nothing
+4   BYOS-1  BYOS-1                 BYOS    Do nothing
+5   BYOS-1  BYOS-2                 No      Do nothing
+6   BYOS-1  BYOS-2                 BYOS    Do nothing
```
rjschwei, line 27: Case 5 & 6 collapse to cases 3 and 4 because the customer cannot change the licence via AHB to a registered state. The action is always

- Change license to BYOS, this is always BYOS-1 to BYOS-1
- Register, this would then result in BYOS-2 or BYOS-3

Given the table we have this is hard to express. At the same time adding more columns such as "Additional user action" is not conducive to making it more readable. I suggest we  make entries 3 & 4 look like this:

`BYOS-1  BYOS-1  (BYOS_2, BYOS-3)* .....`

and the add a footnote like so

`* The user has manually registered he instance with their registration key`
-

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,40 @@
...
+
+7   BYOS-2  PAYG                   No      Do nothing
```
rjschwei, line 29: If the instance is switched to PAYG and the cache is not PAYG we always have to re-register.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,40 @@
...
+8   BYOS-2  PAYG                   BYOS    Re-register - Update Infra
+9   BYOS-2  BYOS-1                 No      De-register - Update Infra
+10  BYOS-2  BYOS-1                 BYOS    De-register - Update Infra
```
rjschwei, line 32: But the customer cannot switch to an unregistered state via AHB. So the condition expressed here really is

- Customer ran `registercloudguest --clean`
- Customer used AHB to explicitly set the BYOS license

As such there is nothing to do. If the starting state is BYOS, any form not registered or registered to the update infra or SCC and the cache says BYOS we have nothing to do. If there is no cache we also do nothing because we cannot do anyting.

usr/sbin/regionsrv-enabler-azure
```diff
@@ -12,6 +12,40 @@
...
+11  BYOS-3  PAYG                   No      No de-register from SCC, register with update infrastructure
+12  BYOS-3  PAYG                   BYOS    No de-register from SCC, register with update infrastructure
+
+13  PAYG    BYOS-1                 No      De-register - Update Infra
+14  PAYG    BYOS-1                 PAYG    De-register - Update Infra
+15  PAYG    BYOS-2,3               No      Do nothing
+16  PAYG    BYOS-2,3               PAYG    Do nothing
```
rjschwei, line 39: Here too, the customer cannot change the license state to `BYOS-2` or `BYOS-3`. Since we have trouble with this I think the column should be renamed from `LICENSE CHANGED TO` to `Reg State at runtime` this would indicate that we start with a given state and then detect the state during runtime of the code. And the state is set by the user as a combination of running AHB commands and registration commands. This should get us out of the quandary we are in. Let's try that and see what that looks like.

rjschwei (changes requested):
We need another iteration on the state transition table. Thanks for your patience.

schaefi:
@rjschwei @jesusbv this one here has its last comment on January, 9 month ago. That feels pretty orphan and after such a long time I usually tend to close PRs because it can't be that important, or the idea turned wrong, or people can't remember anymore :) In any case I'd like to get this resolved. Since I wasn't even part of the team by that time, can I get an estimate how to proceed here ? Thanks

jesusbv:
> @rjschwei @jesusbv this one here has its last comment on January, 9 month ago. That feels pretty orphan and after such a long time I usually tend to close PRs because it can't be that important, or the idea turned wrong, or people can't remember anymore :) In any case I'd like to get this resolved. Since I wasn't even part of the team by that time, can I get an estimate how to proceed here ? Thanks

Do not close the PR, this PR fixed a somewhat important case where users can benefit for free from something they should pay for

schaefi:
>Do not close the PR, this PR fixes a somewhat important case where users can benefit for free from something they should pay for

ok, I wonder how something important like that can stay open for >9months ? What holds us back ?

rjschwei:
The code was removed with #242 as such this code is no longer applicable. The issue that motivated this change is expressed in #243

PR #143, opened 2024-01-26
Add handling case when no credentials present

jesusbv:
In the case of no credentials present,
the check of credentials should not generate
an exception

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -160,6 +160,9 @@ def credentials_files_are_equal(repo_credentials):
     """Compare the base credentials files the the repo header and make
        sure they have the same values."""
     credentials_location = '/etc/zypp/credentials.d/'
+    if repo_credentials is None:
+        return False
+
     credentials_base = os.path.join(credentials_location, 'SCCcredentials')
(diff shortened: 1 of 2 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -160,8 +160,15 @@ def credentials_files_are_equal(repo_credentials):
...
        sure they have the same values."""
     credentials_location = '/etc/zypp/credentials.d/'
+    try:
+        credentials_header = os.path.join(
+            credentials_location,
+            repo_credentials
+        )
+    except TypeError:
+        return False
+
```
rjschwei, line 170: No `try-except` block this is a simple if condition
```
if not repo_credentials:
    return False
```
Receiving an empty velu for an argument is not an exceptional condition.

lib/cloudregister/VERSION
```diff
@@ -1 +1 @@
-10.1.6
+10.1.7
```
rjschwei, line 1: Don't need a version bump for this, we are not going to release just this change.

rjschwei (changes requested):

rjschwei (approved):

PR #144, opened 2024-01-29
Better no valid credentials check

jesusbv:
Little too strict check so a better guard clause


lib/cloudregister/registerutils.py
```diff
@@ -160,7 +160,7 @@ def credentials_files_are_equal(repo_credentials):
     """Compare the base credentials files the the repo header and make
        sure they have the same values."""
     credentials_location = '/etc/zypp/credentials.d/'
-    if repo_credentials is None:
+    if not isinstance(repo_credentials, str):
```
rjschwei, line 163: We also need to return `False` for en empty string. this needs to be `if not repo_credentials or not isinstance(repo_credentials, str)`

rjschwei (changes requested):

rjschwei (approved):

PR #146, opened 2024-02-16
Add registry sidecar support

jesusbv:


rjschwei:
Switching to draft, we are still in the scoping phase of the project.

lib/cloudregister/smt.py
```diff
@@ -40,6 +40,11 @@ def __init__(self, smtXMLNode, https_only=False):
         except KeyError:
             self._region = 'unknown'
         self._fqdn = smtXMLNode.attrib['SMTserverName']
+        try:
+            self._registry_fqdn = 'registry-' + self._fqdn.split('-')[1]
```
bear454, line 44: I think we need to talk about extending the region server to send this FQDN, instead of inferring it.

lib/cloudregister/registerutils.py
```diff
@@ -499,6 +508,44 @@ def get_credentials(credentials_file):
     return (username, password)
 
 
+# ----------------------------------------------------------------------------
+def set_registry_credentials(registry_fqdn, username, password):
+    """Set the auth token to pull images from SUSE registry."""
+    auth_token = base64.b64encode('{username}:{password}'.format(
+        username=username,
+        password=password
+    ).encode()).decode()
```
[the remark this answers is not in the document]
jesusbv: it is not because of security, it is because we _have to_, docker|podman dont like anything that is not base64

```bash
> cat ${XDG_RUNTIME_DIR}/containers/auth.json
{
	"auths": {
		"registry.suse-local.com:5000": {
			"auth": "SCC_foo:password_bar"
		}
	}
}
> podman login  --tls-verify=false
Error: get credentials: 1 error occurred:
	* illegal base64 data at input byte 72
> cat ~/.docker/config.json
{
        "auths": {
                "registry.suse-local.com:5000": {
                        "auth": "SCC_foo:password_bat"
                }
        }
}
> docker login registry.suse-local.com:5000
WARNING: Error loading config file: ~/.docker/config.json: illegal base64 data at input byte 3
Username:
```

from the [podman doc](https://docs.podman.io/en/stable/markdown/podman-login.1.html)
```
[...] Podman creates ${XDG_RUNTIME_DIR}/containers/auth.json (if the file does not exist) and then stores the username and password from STDIN as a base64 encoded string in it
```
from the [docker doc](https://docs.docker.com/reference/cli/docker/login/) 
```
 [...]
 If none of these binaries are present, it stores the credentials (i.e. password) in base64 encoding in the config files described above
 ```

File: lib/cloudregister/registerutils.py [hunk 3]
```diff
@@ -45,7 +45,7 @@
 REGISTRATION_DATA_DIR = '/var/cache/cloudregister/'
 REGISTERED_SMT_SERVER_DATA_FILE_NAME = 'currentSMTInfo.obj'
 RMT_AS_SCC_PROXY_MARKER = 'rmt_is_scc_proxy'
-
+DOCKER_REGISTRY_CREDENTIALS_PATH = '.docker/config.json'
```
schaefi, line 48: You can keep it that way but just to mention ```~/.docker/config.json``` is read if it exists by podman but is only searched for compat reasons for compatibility with docker. If not present the default location for the login credentials is

```
$XDG_RUNTIME_DIR/containers/auth.json
```

jesusbv: I'm not sure I understand, from the podman doc

```
Podman first searches for the username and password in the ${XDG_RUNTIME_DIR}/containers/auth.json, if they are not valid, Podman then uses any existing credentials found in $HOME/.docker/config.json
```
so as I understand it, it will try 
```
$XDG_RUNTIME_DIR/containers/auth.json
```
and it will fail then it will try docker (which we are setting) so it will work

jesusbv: Meaning setting only ~/.docker/config.json is enough for both podman and docker

schaefi: yes it will work as long as podman keeps compatible with docker in this way. As I said you can keep it the way it is. If you want to be super accurate you write both. By the way there are more container engines that supports OCI containers  e.g containerd (k3s/kubernetes) which does it different when using ```ctr image import ...```. Thus I don't think a generic handling is possible and that means explicit is better than implicit. But all that can be added later, just keep in mind this path is not common for everyone

jesusbv: OK, AFAIK we are supporting docker and podman
That said, I will update the PR to set it explicitly for podman, in case compatibility changes in the future

lib/cloudregister/registerutils.py
```diff
@@ -1651,3 +1715,81 @@ def __replace_url_target(config_files, new_smt):
...
+        # one or both keys do not exist
+        if registries_conf.get('unqualified-search-registries') is None:
+            registries_conf['unqualified-search-registries'] = \
+                ["{}".format(public_registry_fqdn)]
+        if registries_conf.get('registry') is None:
+            registries_conf['registry'] = \
+                [private_registry] + [public_registry]
+
+    with open(REGISTRIES_CONF_PATH, 'w') as registries_conf_file:
+        toml.dump(registries_conf, registries_conf_file)
```
schaefi, line 1757: I think this part could be written a bit shorter and more clear with less duplication e.g 

```python
if not os.path.isfile(REGISTRIES_CONF_PATH):
    os.makedirs(os.path.dirname(REGISTRIES_CONF_PATH), exist_ok=True)

registries_conf = _get_registry_conf_file(
    REGISTRIES_CONF_PATH, 'podman'
)

registries_search: List[str] = registries_conf.get('unqualified-search-registries') or []
registries: List[str] = registries_conf.get('registry') or []

if public_registry_fqdn not in registries_search:      
    registries_search.append(public_registry_fqdn)

if public_registry not in registries:
    registries.append(public_registry)

if private_registry not in registries:
    registries.append(private_registry)

registries_conf['unqualified-search-registries'] = registries_search
registries_conf['registry'] = registries

with open(REGISTRIES_CONF_PATH, 'w') as registries_conf_file:
    toml.dump(registries_conf, registries_conf_file)
```

lib/cloudregister/registerutils.py
```diff
@@ -1651,3 +1715,81 @@ def __replace_url_target(config_files, new_smt):
...
+
+
+# ----------------------------------------------------------------------------
+def _get_registry_conf_file(container_path, container):
+    if container in 'podman':
```
schaefi, line 1762: This should be a ```==``` comparision

```if container == 'podman':```

The same for the docker match

lib/cloudregister/registerutils.py
schaefi: and we are only adding the public registry to the search list, which I like because otherwise errors on authentication expiration will not be visible to the user, as we discussed it on the slack channel :+1:

schaefi (commented):
A few comments from my side

lib/cloudregister/registerutils.py
jesusbv: that would set the public registry fqdn at the list, I think the agreement was at the beginning ?

schaefi: ok so

```registries_search.insert(0, public_registry_fqdn)```

sorry I overlooked the order

jesusbv: No worries

I took the same approach (insert o)

jesusbv:
Closing in favor of #157

PR #150, opened 2024-03-05
Add tests for flake8 check

jesusbv:
- Add tests to flake8 check
- Update tests
- Update registerutils


schaefi (approved):
yep looks good :+1: 

The import trick is a little bit of a downside as we have to mark it as "noqa" but this is a good step into the right direction. Thanks much @jesusbv

PR #152, opened 2024-04-12
New command cloudguestregistryauth

jesusbv:
This command refresh the system credentials for the registry


jesusbv:
I chose `get_activations` as 

```python
    ...

    instance_data = bytes(get_instance_data(get_config()), 'utf-8')
    headers = {'X-Instance-Data': base64.b64encode(instance_data)}

    res = requests.get(
        'https://%s/connect/systems/activations' % update_server.get_FQDN(),
        auth=auth_creds,
        headers=headers
    )
```
it includes user + pass + instance metadata and the query is quite cheap

schaefi (commented):
The code proposed here looks good to me. However, all the magic is in ```get_activations``` and we just assume it does the job and has the desired effect, right ?

Other than that I wondered; There are tests but no action runs for the tests. I think the test runs should be active on each PR.

So the python code looks good, as per functionality I have no idea if this solves it. Thus I only looked at the python part of the code and don't feel like the right person to send an approval :)

jesusbv:
> The code proposed here looks good to me. However, all the magic is in `get_activations` and we just assume it does the job and has the desired effect, right ?
> 
> Other than that I wondered; There are tests but no action runs for the tests. I think the test runs should be active on each PR.

Yes, I agree, I have been making efforts to that effect with different PRs, and currently open branch (that depends on an open/unmerged PR) to have the tests running on each PR/commit

> 
> So the python code looks good, as per functionality I have no idea if this solves it. Thus I only looked at the python part of the code and don't feel like the right person to send an approval :)

lib/cloudregister/registerutils.py
```diff
@@ -503,6 +503,23 @@ def get_credentials(credentials_file):
     return (username, password)
 
 
+# ----------------------------------------------------------------------------
+def refresh_registry_credentials():
+    """Refresh registry credentials."""
+    requests.packages.urllib3.disable_warnings() # to silence InsecureRequestWarning
```
bear454, line 509: We shouldn't have an insecure request from a client.

lib/cloudregister/registerutils.py
```diff
@@ -503,6 +503,23 @@ def get_credentials(credentials_file):
...
+    current_server = get_current_smt()
```
bear454, line 510: `get_current_smt` calls `‎get_smt_from_store` which rewrites the local smt settings. IMO that's an unwanted side effect.

jesusbv: I'm not sure it rewrites the local SMT settings ?
`get_smt_from_store` does create an SMT instance from the stored data, 
and `get_current_smt` makes sure that it matches /etc/hosts and that it is registered

is there another step I skipped on those methods ?

lib/cloudregister/registerutils.py
jesusbv: Agree, but the fix for that should be in a different PR, IMO

lib/cloudregister/registerutils.py
jesusbv: do you mean that the stored data gets deleted if the stored data does not match `/etc/hosts` ?

rjschwei: `get_smt_from_store()` loads the current target update server, it will not modify any data. That said, he code path activated does not guarantee that we find a replacement update server if the current server should not be reachable. Meaning as currently implemented we try the current target update server and if for some reason it is not reachable then we give up an tell the user that we failed. That is terrible. more in the next comment......

lib/cloudregister/registerutils.py
```diff
@@ -502,6 +502,25 @@ def get_credentials(credentials_file):
...
+# ----------------------------------------------------------------------------
+def refresh_registry_credentials():
+    """Refresh registry credentials."""
+    # to silence InsecureRequestWarning
+    # should be fixed on a different PR
+    requests.packages.urllib3.disable_warnings()
+    current_server = get_current_smt()
+    if current_server:
+        # system is registered
+        if get_activations():
```
rjschwei, line 514: the `get_activations()` call already handles update server information processing with the `get_smt()` call as the first executed line in the code https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L409 and the `get_smt()` function makes sure we find an update server.

Meaning the proposed implementation in this PR is sloppy at best, at worst it shows a certain laziness about working to understand the implementation of the utils functions that exist.

rjschwei (changes requested):

rjschwei:
And also this only partially covers the requirements. The requirement is that systems accessing the registry end up in a cache server side that last longer than systems that access the repositories. This means somehow we have to tell the server that this is a registry path re-auth. The differentiation server side also needs implementation.

lib/cloudregister/registerutils.py
```diff
@@ -502,6 +502,25 @@ def get_credentials(credentials_file):
...
+            sys.exit(0)
+        message = 'Could not refresh credentials'
+    else:
+        message = 'No current update server configured or ' \
+            'system is not registered'
+    logging.info(message)
+    sys.exit(message)
```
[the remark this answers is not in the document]
jesusbv: from `sys.exit` [docs](https://docs.python.org/3/library/sys.html) on 
```
If another type of object (that are not integers) is passed, None is equivalent to passing zero, 
and any other object is printed to [stderr] (https://docs.python.org/3/library/sys.html#sys.stderr) 
and results in an exit code of 1. In particular,  
sys.exit("some error message") is a quick way to exit a program when an error occurs.
```
When they mentioned stderr, they mean `sys.stderr` 

I will change it, I consider this feedback more a preference than an actual error, I do not have strong opinion on that
I understood to that using one call that is available and does the both things we want it to do (sys.stderr + return 1 meaning error) was better than doing two explicit calls

lib/cloudregister/registerutils.py
```diff
@@ -502,6 +502,19 @@ def get_credentials(credentials_file):
...
 
 
+# ----------------------------------------------------------------------------
+def refresh_registry_credentials():
+    """Refresh registry credentials."""
+    # to silence InsecureRequestWarning
+    # should be fixed on a different PR
+    requests.packages.urllib3.disable_warnings()
+    if get_activations():
+        sys.exit(0)
```
rjschwei, line 512: exiting is the responsibility of the caller. The function shold report appropriate return codes.

lib/cloudregister/VERSION
```diff
@@ -1 +1 @@
-10.1.8
+10.2.0
```
rjschwei, line 1: DO not bump the version just yet, we know this feature is incomplete, we still need a differentiation between calling `get_activations()` for th eregistry path and `get_activations()` for the repository path such that we get different system credentials cache handling server side.

rjschwei (changes requested):

lib/cloudregister/VERSION
jesusbv: Yes, Im not merging this PR yet
the extra param with whatever header name neded on the server side, will be added on this PR, but the version needs to be bumped anyway ? So when the PR is feature complete it can be merged and nothing else to do

rjschwei: WE should not have this PR grow with an additional unknown number of commits. The basic structure is good, remove the version bump, fix the exit conditions in the function of the utils and we can merge the first step.

bear454:
... and get the tests passing again ;-)

bear454 (approved):

PR #155, opened 2024-04-19
Add header for reauth purposes

jesusbv:
- When refreshing the credentials server side, add a header to explicitly set the registry cache file

### DIFF
--- a/tests/test_registerutils.py
+++ b/tests/test_registerutils.py
@@ -1121,23 +1121,23 @@ def test_get_activations_request_registry_header(
          region="antarctica-1"/>''')
     smt_server = SMT(etree.fromstring(smt_data_ipv46))
     mock_get_smt.return_value = smt_server
-    mock_get_creds.return_value = 'foo', 'bar'
-    mock_http_basic_auth.return_value = 'foobar'
+    mock_get_creds.return_value = 'username_foo', 'password_bar'
+    mock_http_basic_auth.return_value = 'http_basic_auth_from_userpass'
     mock_get_instance_data.return_value = 'super_instance_data'
     response = Response()
     response.status_code = 200
     json_mock = Mock()
-    json_mock.return_value = {"foo": "bar"}
+    json_mock.return_value = {"prod": "activated"}
     response.json = json_mock
     mock_request_get.return_value = response
-    assert utils.get_activations(True) == {'foo': 'bar'}
+    assert utils.get_activations(True) == {'prod': 'activated'}
     assert mock_logging.error.not_called
     mock_request_get.assert_called_once_with(
         'https://fantasy.example.com/connect/systems/activations',
-        auth='foobar',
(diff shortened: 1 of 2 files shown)


tests/test_registerutils.py
```diff
@@ -1095,6 +1095,53 @@ def test_get_activations_request_OK(
...
+    response.json = json_mock
+    mock_request_get.return_value = response
+    assert utils.get_activations(True) == {'foo': 'bar'}
+    assert mock_logging.error.not_called
+    mock_request_get.assert_called_once_with(
+        'https://fantasy.example.com/connect/systems/activations',
+        auth='foobar',
+        headers={
+            'X-Instance-Data': b'c3VwZXJfaW5zdGFuY2VfZGF0YQ==',
+            'X-Registry': 'foo'
```
schaefi, line 1140: nitpick, X-Registry is expected to be a username according to the code changes right ? Neither the key name nor the test data would point me to this. I think at least the test here should create clarity e.g

```
'X-Registry': 'some:user_name'
```

schaefi (commented):
Just nitpick on the test, the code looks good to me :+1: 

I'm missing details about the motivation of the new header field as well as where it's been used or what consumes it. I guess this information is in some other API documentation ? I again reviewed only in terms of code not in terms of functionality as a whole

tests/test_registerutils.py
jesusbv: Currently the content of `X-Registry` (or whatever name we decide on) does not matter, 
it is more a flag to know that the action of registry cache handling should be carried on the RMT server

I will update the test

jesusbv:
> On the server side we, i.e. people working on the registry as part of RMT project have currently 2 PRs pending, [SUSE/rmt#1124](https://github.com/SUSE/rmt/pull/1124) and [SUSE/rmt#1128](https://github.com/SUSE/rmt/pull/1128). As commented in 1128 I do not understand the need for it given 1124. Also `X-Registry` is too generic.
> 
> My understanding based on 1124 in the RMT project is as follows:
> 
> When the client, docker or podman, come to visit they will send a token. The code in 1124 will check if the `last_seen_at` time is more then 8 hours ago and if so the token will be rejected. Thus access is denied, meaning the user has to re-auth. When the `activation()` function is called the system via the repository path will not be in the cache and therefore a full verification will take place. This will update `last_seen_at` and thus the next access of the registry if it is less than 8 hours later will work.
> 
> Meaning this change is not needed.

I understood that the behavior on 1124 was correct until 

> The idea is to expire the token after 8 hours no matter how often the system comes to visit in those 8 hours

meaning we would need to force a reauth, depending on that comment or, in other words, whether we are OK having the expiration of the access updated by any access of `zypper` (again, I understood we are OK with that but the comment states different behavior) then this PR would be needed or not

jesusbv:
Clarified that `last_seen_at` is enough for granting access to the registry, we do not need this change

PR #156, opened 2024-05-09
Support systems with transactional updates

rjschwei:
On systems that have a read-only root filesystem and transactional updates `SUSEConnect` disables itself and we need to call the `transactional-update register` command in order to complete system registration. This is not true for transactional-update systems that have a read-write root filesystem. Therefore we need to look at the mount options of the filesystem. On a read-only setup zypper is called to work on a specific directory and we have to take this into account when handling credentials.


lib/cloudregister/registerutils.py
```diff
@@ -874,6 +878,21 @@ def get_zypper_pid_cache():
...
 
+# ----------------------------------------------------------------------------
+def get_zypper_target_root():
+    """Return the target root if zypper has the --root argument to
+       specify a target directory in which to operate.
+    """
+    zypper_cmd = get_zypper_command()
+    target_root = ''
+    for root_arg in ('-R', '--root'):
+        if zypper_cmd and root_arg in zypper_cmd:
```
jesusbv, line 889: shouldn't this be 
```python
if root_arg in zypper_cmd:
```
?

usr/sbin/registercloudguest
```diff
@@ -70,13 +110,23 @@ def register_modules(extensions, products, registered=[], failed=[]):
...
-                register_cmd,
-                '--url',
-                'https://%s' % registration_target.get_FQDN(),
-                '--product',
-                triplet
-            ]
+            if 'transactional' in register_cmd:
+                cmd = [
+                    register_cmd,
+                    sub_cmd,
```
jesusbv, line 116: Not sure if it is a GH issue but I do not see where `sub_cmd` is set ?

lib/cloudregister/registerutils.py
```diff
@@ -159,10 +159,13 @@ def clear_rmt_as_scc_proxy_flag():
...
     """Compare the base credentials files the the repo header and make
        sure they have the same values."""
-    credentials_location = '/etc/zypp/credentials.d/'
+
     if not repo_credentials or not isinstance(repo_credentials, str):
         return False
 
+    base_credentials_location = '/etc/zypp/credentials.d/'
+    target_root = get_zypper_target_root()
+    credentials_location = target_root + base_credentials_location
```
apozsuse, line 168: Would it be worth join the path fragments with `os.path.join`  to avoid possible '//' when adding the strings?   (I assume `os.path.userexpand` is not necessary for the paths we're processing)

apozsuse (approved):
Other than the minor comments, LGTM

usr/sbin/registercloudguest
rjschwei: Good catch, thanks. Fixed. should have been `register` as in the other case.

lib/cloudregister/registerutils.py
rjschwei: The path we get back from reading the process table is formed by the transactional-update command and does not end in a `/`. Also `os.path.join` has some weird rules and to make that work `get_zypper_target_root()` would need to return `\` instead of an empty string and `base_credentials_location` would have to start without a `/`. Meaning one has to be rather careful with the construction of the arguments passed to `os.path.join` . Besides a `//` does not hurt.

Since we will operate under the conditions of transactional-update I think we are safe also skipping `userexpand`. 

There are use cases where the underlying assumptions of operating in an environment constructed `transactional-update` will be violated and the code will fail. IF ever needed we'll come back and look at that specific use case.

PR #157, opened 2024-05-16
Add and clean up registry entry from /etc/hosts

jesusbv:


schaefi (approved):
Looks good to me :+1: 

nitpick: A bit more information in the commit message would be good. Information like that in addition to the hosts entry for the RMT server also an entry and proper cleanup for targeting the distribution registry server needs to be added

lib/cloudregister/registerutils.py
```diff
@@ -86,23 +88,30 @@ def add_region_server_args_to_URL(api, cfg):
 
 
 # ----------------------------------------------------------------------------
-def clean_hosts_file(domain_name):
+def clean_hosts_file(domain_name, registry_name):
```
rjschwei, line 91: The registry should have the same domain as the repository server. I don't think we want to support a feature where the repository is configured as 'foo.us` and the registry is setup as `bar.eu`. I think it is fair to expect that the domain name is the same. If someone ever needs these to be separate the we'll have that conversation at that time, for now let's enforce that the registry and repository exist on the same domain.

lib/cloudregister/registerutils.py
```diff
@@ -790,7 +799,7 @@ def get_smt(cache_refreshed=None):
                     '"%s"' % str((server.get_ipv4(), server.get_ipv6()))
                 )
                 # Assume the new server is in the same domain
-                clean_hosts_file(server.get_FQDN())
+                clean_hosts_file(server.get_FQDN(), server.get_registry_FQDN())
```
rjschwei, line 802: Ditto, let's not support the registry and repository to exist on different domains.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: I agree the domain should be the same, however, in https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/d91381c7b8dacb11e99c8a3726d8dc242a09d62b/lib/cloudregister/registerutils.py#L1297 the domain name is used but in this line you are pointing here the FQDN is used

note that `smt.get_domain_name()` != `smt.get_FQDN()`
                  `susecloud.net`                   != `smt-ec2.susecloud.net`
with that into account, with the current implementation of `clean_hosts_file` and using the domain name (`smt.get_domain_name()`) the registry entry does **not** get cleaned up

jesusbv: I will adapt the clean_hosts_file method to work _only_ with the domain and works to remove the registry entry as well, unless any objection

lib/cloudregister/registerutils.py
```diff
@@ -56,10 +56,12 @@ def add_hosts_entry(smt_server):
...
     if has_ipv6_access(smt_server):
         smt_ip = smt_server.get_ipv6()
-    entry = '%s\t%s\t%s\n' % (
+    entry = '%s\t%s\t%s\n%s\t%s\n' % (
         smt_ip,
         smt_server.get_FQDN(),
-        smt_server.get_name()
+        smt_server.get_name(),
+        smt_ip,
+        smt_server.get_registry_FQDN()
```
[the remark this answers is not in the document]

lib/cloudregister/registerutils.py

Updated the code to handle domain name only
Updated and added tests to check that behavior

usr/sbin/registercloudguest
```diff
@@ -214,6 +214,12 @@ argparse.add_argument(
     dest='user_smt_fqdn',
     help=help_msg
 )
+help_msg='The target registry server FQDN. '
+argparse.add_argument(
```
jesusbv, line 218: I will remove this option as it should not be available for the user

lib/cloudregister/registerutils.py
```diff
@@ -101,13 +107,21 @@ def clean_hosts_file(domain_name):
...
     for entry in content:
         if b'# Added by SMT' in entry:
             smt_announce_found = True
             continue
         if smt_announce_found and domain_name in entry:
             smt_announce_found = False
+            smt_domain_found = True
             continue
+        if smt_domain_found:
+            smt_domain_found = False
```
rjschwei, line 120: This makes the assumption that the registry entry is right after the repository entry. Which is what we create. However, users do stuff we don't want them to do. Therefore the flag needs to be set to `False` after we identify the line we want to remove. This means that we inspect every line after the announce flag is found if there is no registry entry.

For example
```
# Added by SMT registration do not remove, retain comment as well
2.2.2.2 smt-ec2.susecloud.net smt-ec2
1.1.1.1 my.specialhost.us
2.2.2.2 registry-ec2.susecloud.net smt-ec2
```
In this case the registry entry would not get cleaned up.

lib/cloudregister/registerutils.py
```diff
@@ -101,13 +107,21 @@ def clean_hosts_file(domain_name):
...
+            if b'registry' in entry and domain_name in entry:
+                # registry entry is optional
+                continue
+
```
rjschwei, line 124: In the diff it looks like the `new_hosts_content.append(entry)` moved outside the for loop, please double check this is not the case.

tests/test_smt.py
```diff
@@ -34,25 +34,29 @@
...
      SMTserverName="fantasy.example.com"
+     SMTregistryName="registry-fantasy.example.com"
      region="antarctica-1"/>''')
 
 smt_data_no_region = dedent('''\
     <smtInfo fingerprint="00:11:22:33"
      SMTserverIP="192.168.1.1"
-     SMTserverName="fantasy.example.com"/>''')
+     SMTserverName="fantasy.example.com"
+     SMTregistryName="registry-fantasy.example.com"/>''')
```
rjschwei, line 59: There should be a test setup without `SMTregistryName`

tests/test_registerutils.py
```diff
@@ -323,6 +326,7 @@ def test_clean_host_file_no_empty_bottom_lines():
 
 # Added by SMT, please, do NOT remove this line
 2.3.4.5   smt-entry.susecloud.net smt-entry
+2.3.4.5   registry-entry.susecloud.net
```
rjschwei, line 329: We should create a test case where the user interfered in the order see the example in a previous comment

rjschwei (changes requested):

tests/test_smt.py
jesusbv: You mean empty value or no entry at all ?
Because currently the presence of the entry is mandatory [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/157/files#diff-9811adbce77061bf864555ec5e53fec8f5e8b2e87fa91b1618dabe3ef564a139R43)
the registry entry presence on the RMT server (smt info) should be mandatory, shouldn't it ?

jesusbv: I will change it to non mandatory as maybe 3rd parties users have an RMT version excluding or not using the registry at all

lib/cloudregister/registerutils.py
jesusbv: I checked, it is inside the for loop in the code

lib/cloudregister/registerutils.py
```diff
@@ -105,9 +111,10 @@ def clean_hosts_file(domain_name):
         if b'# Added by SMT' in entry:
             smt_announce_found = True
             continue
-        if smt_announce_found and domain_name in entry:
-            smt_announce_found = False
-            continue
+        if smt_announce_found:
+            if domain_name in entry:
```
jesusbv, line 115: we could check whether `b'smt'` and `b'registry'` are inside `entry` in case there are more entries with the same domain added by the user, as we also ignore the [entries with the same domain before the comment added by SMT](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/cc2603406183c585bc6ae78d46f22a70a80af122/tests/test_registerutils.py#L325)

jesusbv: will resolve this conversation unless any objection

rjschwei (approved):

PR #159, opened 2024-06-05
Set registry credentials

jesusbv:
Set the registry auth token


lib/cloudregister/registerutils.py
```diff
@@ -509,6 +510,47 @@ def get_credentials(credentials_file):
     return (username, password)
 
 
+# ----------------------------------------------------------------------------
+def setup_registry(registry_fqdn, username, password):
```
bear454, line 514: What's the point of this method? Why not just call `set_registry_auth_token`, or rename it `s/set_registry_auth_token/setup_registry/` ?

lib/cloudregister/registerutils.py
```diff
@@ -509,6 +510,47 @@ def get_credentials(credentials_file):
...
+        # config file does not exist or "auths" key is not set
+        os.makedirs(os.path.dirname(REGISTRY_CREDENTIALS_PATH), exist_ok=True)
+        config_json.update({'auths': registry_credentials})
+    except json.decoder.JSONDecodeError:
+        logging.info(
+            'Error found when opening %s' % REGISTRY_CREDENTIALS_PATH
+        )
+        config_json.update({'auths': registry_credentials})
+
+    with open(REGISTRY_CREDENTIALS_PATH, 'w') as cred_json_file:
```
schaefi, line 546: If opening the file for writing or the actual write failed the application will fail with a stack trace. Do we want to handle this error condition ?

schaefi (commented):
Looks good to me, just one error condition that we might want to catch

lib/cloudregister/registerutils.py
jesusbv: in that method I plan to call more methods, so when `setup_registry` finishes, all the bits needed for registry are done and in place

lib/cloudregister/registerutils.py
```diff
@@ -509,6 +510,47 @@ def get_credentials(credentials_file):
...
+    try:
+        with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
+            config_json = json.load(cred_json)
+        # file exists
+        # set the new registry credentials,
+        # independently of what that content was
+        config_json['auths'].update(registry_credentials)
+    except (FileNotFoundError, KeyError):
+        # config file does not exist or "auths" key is not set
+        os.makedirs(os.path.dirname(REGISTRY_CREDENTIALS_PATH), exist_ok=True)
```
rjschwei, line 538: This is rather in-discriminant, if the file doesn't exist we also assume that the directory doesn't exist and then try to create it. If we are going to be in-discriminant about the operation and we know we need the directory why do we not set it up in the wrapper function i.e. in `setup_registry` there we could have 

```
if not os.path.exists(os.path.dirname(REGISTRY_CREDENTIALS_PATH)):
    os.makedirs(os.path.dirname(REGISTRY_CREDENTIALS_PATH)
```

The directory is one of the fundamental things for setting up the registry on the client as such it should be at the highest level possible. Then we also do not have to handle 2 different things here.

lib/cloudregister/registerutils.py
rjschwei: Yes, there shold be a try-except block here and further up the code where we handle file existence should be handled with `if os.path.exists(REGISTRY_CREDENTIALS_PATH):`

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: Creating the dir as early as possible does not eliminate the handling of the 2 different things in this case:

- if dir did not exist, and we created it, the file won't exist
- if dir did exist does not guarantee the file to be present anyway

I will check for the dir as early as possible as suggested but handling both exception

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,56 @@ def get_credentials(credentials_file):
...
+# ----------------------------------------------------------------------------
+def set_registry_auth_token(registry_fqdn, username, password):
+    """Set the auth token to pull images from SUSE registry."""
+    auth_token = base64.b64encode('{username}:{password}'.format(
+        username=username,
+        password=password
+    ).encode()).decode()
+    registry_credentials = {registry_fqdn: {'auth': auth_token}}
+    config_json = {}
+    try:
```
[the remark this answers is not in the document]
jesusbv: wouldn't that be a one time occurrence ? 
and after any successful subsequent de-registration + registration the file would be present, right ?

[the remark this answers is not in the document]
jesusbv: also, in the scenario of subsequent registrations, wouldn't it be better to not check with an if, in case the file gets removed between the check and the actual reading ?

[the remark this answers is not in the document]
jesusbv: Maybe it is not an exception per se, but depending on the use, the file being present is more likely to be the common case
Still think it is better to deal with files in a better way than an if but  the likelihood of the file being remove may be low, guess we can change it if that scenario ever showed up

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,59 @@ def get_credentials(credentials_file):
...
+        config_json.update({'auths': registry_credentials})
+    else:
+        try:
+            with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
+                config_json = json.load(cred_json)
+            # set the new registry credentials,
+            # independently of what that content was
+            config_json['auths'].update(registry_credentials)
+        except KeyError:
+            # file does not exist or "auths" key is not set
```
rjschwei, line 552: When we get here we know the file exists, we already checked that condition. So the comment is incorrect. Also why not use a way that avoids exceptions when we can. Exceptions are for exceptional cases that are hard to test or that we truly cannot expect.

We can do positive testing and avoid some exceptions as follows: (Also handles the case to preserve the file we couldn't parse, rather than clobbering it)

```
if os.path.exists(REGISTRY_CREDENTIALS_PATH):
    try:
        with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
            container_engine_config = json.load(cred_json)
        if 'auths' in container_engine_config:
            container_engine_config['auths'].update(registry_credentials)
    except Error:
        logging.info('Unable to parse existing %s, preserving file as %s.bak, writing new credentials') % (REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH)
        os.system('mv -Z %s %s.bak' % (REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH)
else:
    container_engine_config = {'auths': registry_credentials}
```

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: I appreciate the suggestion, wasn't trying to be difficult with the coding styles

I don't have a strong preference for any of the two styles, followed the one preferred by the community (EAFP) even if it felt more natural the other one (coming from other languages like C/C++) 
I can see benefits/pros in both and I can see cons in both as well

I guess or understand you have a preference for one of them, I will follow that style then (as I said, no strong preference), for files/dictionaries (like checking the key in this case but especially with files) I find the `try-except` approach handy, easier to read and to code than the suggested code. Still will use the suggested style as the main one and keep the `try-except` as the less common approach

I understand the suggestion is pseudo code and I will take the preservation of the file  into account

jesusbv: any reason about why using `mv` instead of `cp` ?

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,63 @@ def get_credentials(credentials_file):
...
+            logging.info(
+                'Unable to parse existing %s, preserving file as %s.bak, '
+                'writing new credentials' % (
+                    REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH
+                )
+            )
+            mv_file_cmd = 'mv -Z {} {}.bak'.format(
+                REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH
+            )
+            exec_subprocess(mv_file_cmd.split())
```
jesusbv, line 556: do we want to log that the preservation of the problematic file suceeded ?
something like
```python
failed = exec_subprocess(mv_file_cmd.split())
message = 'File not created' if failed else 'File created'
logging.info(message)
```

rjschwei: Good point. Also reminds me that the logic is not yet complete. If the file exists we should comprehend the file. It could be that the user created a golden image without cleaning up, in which case we would preserve a file with old credentials and nothings else, which we do not want to do. We only want to preserve a file that has content other than the authentication setting for our registry. Meaning we need to do a bit more checking.

Rest looks much better now, thanks

jesusbv: Thanks for the feedback

> we need to do a bit more checking

not sure we can ? we try to open the json file and it crashes when parsing with `JSONDecodeError`
how can we check its content ?

jesusbv: do you want the code to run `grep` or another command to check its content without accessing it from Python ?

rjschwei (approved):

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,90 @@ def get_credentials(credentials_file):
     return (username, password)
 
 
+# ----------------------------------------------------------------------------
+def setup_registry(registry_fqdn, username, password):
+    """Set all the necessary parts for the registry,
+       returns True if the setup completed, False otherwise."""
+    if not os.path.exists(os.path.dirname(REGISTRY_CREDENTIALS_PATH)):
+        os.makedirs(os.path.dirname(REGISTRY_CREDENTIALS_PATH))
```
schaefi, line 525: This can be done simpler with
```python
os.makedirs(os.path.dirname(REGISTRY_CREDENTIALS_PATH), exists_ok=True)
```

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,90 @@ def get_credentials(credentials_file):
...
+def get_registry_credentials():
+    """Read the registry credentials file
+       and return its content or an empty dict."""
+    config_json = {}
+    failed = False
+    if os.path.exists(REGISTRY_CREDENTIALS_PATH):
+        try:
+            with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
+                config_json = json.load(cred_json)
+        except json.decoder.JSONDecodeError:
```
schaefi, line 543: We do not catch open errors this way, intentionally ?

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,90 @@ def get_credentials(credentials_file):
...
+    """Update the registry credentials file with the value of 'content'."""
+    try:
+        with open(REGISTRY_CREDENTIALS_PATH, 'w') as cred_json_file:
+            json.dump(content, cred_json_file)
+        logging.info(
+            'Credentials for the registry set in %s' %
+            REGISTRY_CREDENTIALS_PATH
+        )
+        return True
+    except Exception as error:
```
schaefi, line 574: we catch any Exception, intentionally ?

lib/cloudregister/registerutils.py
```diff
@@ -516,6 +517,90 @@ def get_credentials(credentials_file):
...
+
+
+# ----------------------------------------------------------------------------
+def write_registry_credentials(content):
+    """Update the registry credentials file with the value of 'content'."""
+    try:
+        with open(REGISTRY_CREDENTIALS_PATH, 'w') as cred_json_file:
+            json.dump(content, cred_json_file)
+        logging.info(
+            'Credentials for the registry set in %s' %
```
schaefi, line 570: I suggest to use f-strings 

```python
logging.info(
    f"Credentials for the registry set in '{REGISTRY_CREDENTIALS_PATH}'"
)
```

schaefi (commented):
Thanks for your effort. The implementation looks complete :+1: See my comments on error handling and simplification. Thanks

lib/cloudregister/registerutils.py
jesusbv: I'd like to use f-string but this code needs to be able to run on SLES 12 SP5, and IIRC, that means be compatible with Python 3.4, I think f-string are available in Python 3.6

schaefi: oh, ok sorry didn't thought about that

lib/cloudregister/registerutils.py
rjschwei: The path exists, we are root. 2 issues remain, the system is out of file handles or the disk is bad. For neither there is anything we can do about it, It's OK to blow up.

lib/cloudregister/registerutils.py
rjschwei: Yes, if we cannot write the registry credentials, for whatever reason we still want to proceed and at least preserve the  registration of the repositories.

lib/cloudregister/registerutils.py
jesusbv: Done in [93bf1aa](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/159/commits/93bf1aa8e92b4b7acee4aab51034087fbeb84149)

lib/cloudregister/registerutils.py
rjschwei: Well 12 SP5 is almost EOL and we do not have containers that require authentication that are part of any SLE 12 product.  We can freeze cloud-regionsrv-client at 10.2.0 in SLE 12 and submit these changes only to SLE 15.

lib/cloudregister/registerutils.py
jesusbv: > if we cannot write the registry credentials, for whatever reason we still want to proceed and at least preserve the registration of the repositories.

I agree with that, but the "whatever reason" part could be more specific, right ?
either the file fails to open or the dump failed, do we want to capture a general `Exception` ?

lib/cloudregister/registerutils.py
jesusbv: SLES 12 SP5 GS ends on 10 Oct 2024 and LTSS on the same date of 2027, are we sure the "not have containers that require authentication...on SLES12" part will not change on that time ?

lib/cloudregister/registerutils.py
rjschwei: I do not think we need to be that pedantic and catching anything that shows up at this point here is fine.

schaefi (approved):

PR #160, opened 2024-06-05
Some changes to make building possible

mcepl:
Trying to catch a bug in `reauth` branch, but I needed to do some work to even build the thing.


mcepl:
@dirkmueller Do you have any sudden idea why `pkg-config --libs python3` on my system is empty?

dirkmueller:
> @dirkmueller Do you have any sudden idea why `pkg-config --libs python3` on my system is empty?
> 
> In the end I filed https://bugzilla.suse.com/1226015.

I think we maybe want to use `pkg-config --libs python3-embed` with newer interpreters. it just doesn't exist on sle15 yet (or I am blind).

dirkmueller:
I suggest to split the mock removal patch from the other one because the mock-removal should go in globally imho.

mcepl:
Yes, `python3-embed` is the proper name, just came up with this as well.

Makefile
```diff
@@ -22,7 +22,7 @@ tar:
 
 exec:
 	mkdir -p usr/bin
-	gcc  $$(pkg-config --cflags python3) -Wno-unused-result -Wsign-compare -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -DNDEBUG -fmessage-length=0 -grecord-gcc-switches -g -DOPENSSL_LOAD_CONF -fwrapv -fno-semantic-interposition src/reauth.c $$(pkg-config --libs python3) -o usr/bin/cloudguestregistryauth
+	gcc  $$(pkg-config --cflags python3-embed) -Wno-unused-result -Wsign-compare -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -DNDEBUG -fmessage-length=0 -grecord-gcc-switches -g -DOPENSSL_LOAD_CONF -fwrapv -fno-semantic-interposition src/reauth.c $$(pkg-config --libs python3-embed) -o usr/bin/cloudguestregistryauth
```
rjschwei, line 25: Fails on 15 SP2.
```
~> pkg-config --libs python3-embed
Package python3-embed was not found in the pkg-config search path.
Perhaps you should add the directory containing `python3-embed.pc'
to the PKG_CONFIG_PATH environment variable
No package 'python3-embed' found
```
cloud-regionsrv-client still has to support 15 SP2.

I agree with @dirkmueller the build questions should be addressed in a separate PR than the improvements for unit testing.

rjschwei (changes requested):

Makefile
mcepl: No problem, I will split this.

mcepl:
Just building stuff. And if we don’t have `python3-embed.pc` in Python 3.6 on SLE, then it is a problem.

Makefile
mcepl: Hmm, concerning `python3-embed.pc` … it arrived only in https://github.com/python/cpython/pull/13500 (for Python 3.8; first released with tag `v3.8.0b1`). Now, the question is whether we really care about this functionality enough to make our Python non-standard. OTOH, it is just one file, which shouldn’t make much difference. @dirkmueller , opinions?

rjschwei: It looks to me the only way to make this work in a more generic way is by passing in variable from the spec file, something like `CFLAGS_PYTHON` and `LIBS_PYTHON`. Then in the spec file when/if we switch the build to Python 3.11 we could change the value of the variables. Which is equally "cumbersome" as touching the makefile. The advantage of doing it in the spec file is that the change is front an center, while having the change in the Makefile means the developer has to remember that it's there without a hint. As such using variable in the Makefile and setting those through the spec file does have an advantage.

mcepl:
On Mon Jun 10, 2024 at 8:43 PM CEST, Robert Schweikert wrote:
> It looks to me the only way to make this work in a more generic way is by passing in variable from the spec file, something like `CFLAGS_PYTHON` and `LIBS_PYTHON`. Then in the spec file when/if we switch the build to Python 3.11 we could change the value of the variables. Which is equally "cumbersome" as touching the makefile. The advantage of doing it in the spec file is that the change is front an center, while having the change in the Makefile means the developer has to remember that it's there without a hint. As such using variable in the Makefile and setting those through the spec file does have an advantage.

Well, passing of environmental variable would help, but I don’t think it is the only way:

1. If you give me a good business reason why embedding of the 100% exact same code on Python 3.6 and 3.11 is such necessary thing, I can add yet another symlinks to Python 3.6 packages, so that `python3-embed.pc` would be available everywhere. Problem solved.

2. Otherwise, yes it would an environmental variable, but it would say `"python3"` on Python 3.6 and `"python3-embed"` on Python 3.11. Problem solved.

Much better than fiddling with `CFLAGS` and `LDFLAGS`, possibly in the way incompatible with what our interpreters are built for.

rjschwei:
I have no strong preference. This package is for SLE only and when/if it gets switched to Python 3.11 the Makefile can be updated accordingly. From my perspective this PR can be closed.

mcepl:
On Mon Jun 10, 2024 at 9:15 PM CEST, Robert Schweikert wrote:
> I have no strong preference. This package is for SLE only and
> when/if it gets switched to Python 3.11 the Makefile can be
> updated accordingly. From my perspective this PR can be closed.

Well, if it is for Python 3.6 only, then s/python-embed/python/g and you are done.

rjschwei:
> On Mon Jun 10, 2024 at 9:15 PM CEST, Robert Schweikert wrote: I have no strong preference. This package is for SLE only and when/if it gets switched to Python 3.11 the Makefile can be updated accordingly. From my perspective this PR can be closed.
> Well, if it is for Python 3.6 only, then s/python-embed/python/g and you are done.

Which we already have and why I propose to close this PR.

PR #162, opened 2024-06-06
Remove auth registry credentials

jesusbv:


lib/cloudregister/registerutils.py
```diff
@@ -567,6 +567,52 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+        logging.info('No auth key present. Nothing to do.')
+        return True
+    except json.decoder.JSONDecodeError:
+        logging.info(
+            'Could not unset the registry credentials: '
+            'Error found when opening %s' % REGISTRY_CREDENTIALS_PATH
+        )
+        return
+
+    try:
```
jesusbv, line 603: this `try/except` block of updating the `/etc/containers/config.json` file can has its own method so it can be reused in `set_registry_auth_token()` as well

I plan to do that once the registry PRs are merged

jesusbv: Created the method in #159, no reason to wait

lib/cloudregister/registerutils.py
```diff
@@ -601,6 +602,45 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+            entry = smt.get_registry_FQDN()
+        else:
+            logging.info('Unsetting the auth entry based on the token')
+            auth_token = __generate_registry_auth_token()
+            entries = config_json.get('auths', {}).items()
+            entry = [entry for entry, auth in entries if auth == auth_token]
+            entry = ''.join(entry)
+
+        config_json.get('auths', {}).pop(entry, {})
+        logging.info('Registry auth entry unset or not present on the file')
```
rjschwei, line 634: We should check if our content is the only content in the file and if it is simply `unlink` the file.

tests/test_registerutils.py
```diff
@@ -3412,7 +3414,275 @@ def test_setup_registry_content_write_error(
...
+
+# ---------------------------------------------------------------------------
+@patch('cloudregister.registerutils.get_credentials')
+@patch('cloudregister.registerutils.os.path.exists')
+@patch('cloudregister.registerutils.json.dump')
+@patch('cloudregister.registerutils.logging')
+@patch('cloudregister.registerutils.get_smt_from_store')
+@patch('cloudregister.registerutils.__get_registered_smt_file_path')
+@patch('cloudregister.registerutils.json.load')
+def test_remove_auth_entry_content_write_no_smt_toke_based(
```
rjschwei, line 3607: s/toke/token/

lib/cloudregister/registerutils.py
```diff
@@ -601,6 +602,45 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+
+    if config_json:
```
rjschwei, line 636: If our entry was the only entry config_json contains no useful information and we'll write `{'auths':{}}` which is not meaningful.

lib/cloudregister/registerutils.py
```diff
@@ -531,31 +531,35 @@ def setup_registry(registry_fqdn, username, password):
...
-            logging.info(message)
+    failed = None
+    if not os.path.exists(REGISTRY_CREDENTIALS_PATH):
+        logging.info('Credentials file does not exist. Nothing to do')
+        return config_json, failed
+
+    try:
+        with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
+            config_json = json.load(cred_json)
+    except json.decoder.JSONDecodeError:
```
schaefi, line 546: This is not catching opening errors, e.g permission denied. Intentional ?

lib/cloudregister/registerutils.py
```diff
@@ -531,31 +531,35 @@ def setup_registry(registry_fqdn, username, password):
...
+        logging.info(message)
+        # we want to remove the possible
+        # auth token present in the file
+        # before preserving it
+        mv_file_cmd = 'mv -Z {} {}.bak'.format(
+            REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH
+        ).split()
+        failed = exec_subprocess(mv_file_cmd)
+        message = 'File not preserved' if failed else 'File preserved'
+        logging.info(message)
```
schaefi, line 562: We need our own exec_subprocess to move a file ? How about ```shutil.move```. It's also unclear to me why the selinux fattr needs to be reset to the default. It actually feels wrong to do that if there is a selinux context set on the file. If this is wanted the comment should rather state the reason for the use of the mv options instead of saying in text what the next line does :)

lib/cloudregister/registerutils.py
```diff
@@ -567,28 +571,25 @@ def write_registry_credentials(content):
...
     except Exception as error:
-        logging.error('Could not set the registry credentials: %s' % error)
+        logging.error('Could not update the registry credentials: %s' % error)
 
 
 # ----------------------------------------------------------------------------
 def set_registry_auth_token(registry_fqdn, username, password):
     """Set the auth token to access the SUSE registry."""
-    config_json, preserve_failed = get_registry_credentials()
+    config_json, preserve_failed = get_registry_credentials(True)
```
schaefi, line 585: I would use ```set_new=True``` for better readability

lib/cloudregister/registerutils.py
schaefi: I would also delete the file instead of changing the content. The other logic for creating the credentials is also based on the presence of the file and iirc the logic above could leave a valid file with empty auths which confuses the creation of a new credentials file

schaefi (commented):
Thanks for your work :+1: I think the logic in remove_auth_entry needs to be refactored. Other than that only nitpick on error handling and readability :)

lib/cloudregister/registerutils.py
jesusbv: > We need our own exec_subprocess to move a file ? How about shutil.move

the suggested method to use was `os.system` and call a cmd, happy to use `shutil.move`, 
if we are calling a cmd, yes  it is my opinion to use the same method for that `exec_subprocess` in this case

lib/cloudregister/registerutils.py
jesusbv: Yes, it will get the [code](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L543) from #159 once merged into this branch

lib/cloudregister/registerutils.py
jesusbv: good catch, was treating `config_json` as if `auths` had content

lib/cloudregister/registerutils.py
jesusbv: if the file has other content, I would not remove the file

The file gets removed when
- empty content -> `{}`
- not relevant content -> `{"auths": {}`
- only our entry -> `{"auths": {our_fqdn: our_auth_token}`

otherwise, files does not get removed

lib/cloudregister/registerutils.py
jesusbv: master branch merged and conflicts fixed, tests updated

lib/cloudregister/registerutils.py
```diff
@@ -530,73 +530,80 @@ def setup_registry(registry_fqdn, username, password):
...
-                'Unable to parse existing %s, preserving file as %s.bak, '
-                'writing new credentials' % (
-                    REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH
-                )
-            )
-            file_error = True
+    failed = None
+    if not os.path.exists(REGISTRY_CREDENTIALS_PATH):
+        logging.info('Credentials file does not exist. Nothing to do')
+        return config_json, failed
```
rjschwei, line 540: The function is called `get_registry_credentials()` in this case we cannot return the credentials, so I would consider that a failure, but we return `None` maybe indicating no failure? The intend here is not clear and we need to sort out what it should be.

lib/cloudregister/registerutils.py
rjschwei: `os.system` was an example, anyway I agree we already have a subprocess wrapper we should use it.

With respect to the SELinux context, I am not sure how that works when we create the file or if a user created the file. We get here when we cannot read the existing config file, for whatever reason and since we create a backup maybe we do not care for the SELinux context. That said, this was part of the previous PR and is an artifact in this diff because this branch has not yet been rebased. An artifact of chaining branches and then not waiting until the start of the chain is merged and rebasing before sending a new PR. Which makes review more difficult.

lib/cloudregister/registerutils.py
```diff
@@ -605,10 +612,90 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+        auth_token = __generate_registry_auth_token()
+        entries = config_json.get('auths', {}).items()
+        entry = [entry for entry, auth in entries if auth == auth_token]
+        entry = ''.join(entry)
+
+    config_json.get('auths', {}).pop(entry, {})
+    logging.info('Registry auth entry unset or not present on the file')
+    # file was not empty or
+    # file could not be parsed and the remove cmd did not fail
+    # and there is content worht updating the credentials file
```
rjschwei, line 675: s/worht/worth/

lib/cloudregister/registerutils.py
```diff
@@ -605,10 +612,90 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+    # we could open the credentials file
+    # and it is not empty
+    # unset the registry credentials
+    entry = None
+    smt = get_smt_from_store(__get_registered_smt_file_path())
+    if same_registry_auth_content(config_json, smt):
+        # if the content of the registry auth file is only
+        # our auth info, remove the file
+        logging.info(
+            'The content of the registry auth is our credentials only. '
```
rjschwei, line 653: s/The content of the registry auth is our credentials only./Registry authentication config only contains managed content/

lib/cloudregister/registerutils.py
jesusbv: I applied the changes from master on this branch yesterday

> Which makes review more difficult

Yes, sorry about that, it's a side effect of the time constraint we have for this feature
Hopefully, it is not that much more difficult

lib/cloudregister/registerutils.py
```diff
@@ -605,10 +612,90 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+    if smt and smt.get_registry_FQDN():
+        logging.info(
+            'Unsetting the auth entry for %s' % smt.get_registry_FQDN()
+        )
+        entry = smt.get_registry_FQDN()
+    else:
+        logging.info('Unsetting the auth entry based on the token')
+        auth_token = __generate_registry_auth_token()
+        entries = config_json.get('auths', {}).items()
+        entry = [entry for entry, auth in entries if auth == auth_token]
```
jesusbv, line 668: I think in [this case,](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/162/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R696) checking the content on a dict like `{"auths": {"xxxxx?": "auth_token_we_can_match"}}`

I think for this case, it would be better to check the parts we know of the FQDN, will modify it for that

lib/cloudregister/registerutils.py
jesusbv: Yes, the returning of `None` is intentional to differentiate the case of the non existing file with the case of existing file and failing and failing to open it or accessing the file and being empty, for example

lib/cloudregister/registerutils.py

lib/cloudregister/registerutils.py

lib/cloudregister/registerutils.py
jesusbv: Moved the check outside the method, I think that makes the intent clear

lib/cloudregister/registerutils.py
jesusbv: > anyway I agree we already have a subprocess wrapper we should use it.

@rjschwei I understand you prefer `exec_subprocess` than using `os.rename` for this case, correct ?
or maybe not prefer as a strong word but maybe more "let's make fewer changes"

schaefi: I'm fine either way. If you still need the SELinux context related move, let's stick to how it is right now

schaefi (approved):
From my perspective this looks good

lib/cloudregister/registerutils.py
```diff
@@ -530,73 +530,78 @@ def setup_registry(registry_fqdn, username, password):
...
+    file_error = False
+    message = ''
+    try:
+        with open(REGISTRY_CREDENTIALS_PATH, 'r') as cred_json:
+            config_json = json.load(cred_json)
+    except OSError as error:
+        message = (
+            'Unable to open %s: %s, preserving file as %s.bak'
+            % (REGISTRY_CREDENTIALS_PATH, error, REGISTRY_CREDENTIALS_PATH)
+        )
```
rjschwei, line 547: Something is not right here, either the merge didn't work or the diff is off, what is shown as the diff that we try to read the config file and if we cannot then we log a message that we could not preserve the file. Given such inconsistencies it is basically impossible to review the PR.

jesusbv: I see
```python
    except OSError as error:
        message = (
            'Unable to open %s: %s, preserving file as %s.bak'
            % (REGISTRY_CREDENTIALS_PATH, error, REGISTRY_CREDENTIALS_PATH)
        )
```
which is part of the original intent
open the credentials file to update the credentials, 
opening failed, preserve the file <-- that is the part of that exception
create a new net credentials file with the info
  
 I can squash the commits into one, see if that help ?

rjschwei: But the preserving of the file is not there, line 547 in the diff has a `)` from the log message, then line 548 has `if set_new`. There is no attempt to create `.bak`

jesusbv: yes, there is ? if there was an [error](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/162/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R560), we do the preserve-file op

rjschwei: Ok there has to be a way to handle this with reasonable code locality.

jesusbv: we do the attempt to move the file when there are two exceptions, so we do it once instead of twice and log the message accordingly, we have used that style in our code base before

lib/cloudregister/registerutils.py
```diff
@@ -531,40 +530,187 @@ def setup_registry(registry_fqdn, username, password):
...
+            error_msg += ', writing new credentials'
+        logging.info(error_msg)
+        mv_file_cmd = 'mv -Z {} {}.bak'.format(
+            REGISTRY_CREDENTIALS_PATH, REGISTRY_CREDENTIALS_PATH
+        ).split()
+        failed = exec_subprocess(mv_file_cmd)
+        message = 'File not preserved.' if failed else 'File preserved.'
+        logging.info(message)
 
+    return config_json, failed
```
rjschwei, line 564: Since `failed` has 3 states, which does not make sense for a variable named `failed` we should rename it to `status`.

rjschwei (changes requested):

rjschwei (approved):
Please squash before merging.

PR #163, opened 2024-06-11
Handle registry env vars

jesusbv:
Add and remove registry environment variables in the bashrc file


lib/cloudregister/registerutils.py
```diff
@@ -598,14 +587,56 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
         list(registry_credentials.items())            # new content
     )
+    logging.info('Writing new credentials')
     updated = write_registry_credentials(config_json)
     return updated
 
 
+# ----------------------------------------------------------------------------
+def set_container_engines_env_vars():
+    """Set the environment variables needed available for
```
rjschwei, line 597: s/available//

lib/cloudregister/registerutils.py
```diff
@@ -598,14 +587,56 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+    return update_bashrc(export_registry_env_vars, 'a') and source_env()
+
+
+# ----------------------------------------------------------------------------
+def update_bashrc(content, mode):
+    """Update (add or remove) the env vars for the container engines
+    with the location of the config file to the bashrc local file."""
+    try:
+        with open(BASHRC_LOCAL_PATH, mode) as basrhc_file:
+            basrhc_file.write(content)
```
rjschwei, line 616: I don't see where the file is getting updated. We simply write the content that is passed in. If the user built their own golden image with a custom `/etc/bash.bashrc.local` file then we just clobbered the users content.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: We add the two environment vars to the file, whatever the file had

I will modify it accordingly

lib/cloudregister/registerutils.py
```diff
@@ -1912,3 +2043,14 @@ def __generate_registry_auth_token(username=None, password=None):
         username=username,
         password=password
     ).encode()).decode()
+
+
+# ----------------------------------------------------------------------------
+def __mv_file_backup(filename):
```
jesusbv, line 2049: I created the method to move the files in this PR, a while ago,

I can rename it to `__create_config_backup(file_name)` in a refactor, in a different PR

lib/cloudregister/registerutils.py
```diff
@@ -717,6 +783,79 @@ def same_registry_auth_content(content, smt):
...
+    """Remove the registry environment variables."""
+    env_vars = ['REGISTRY_AUTH_FILE', 'DOCKER_CONFIG']
+    if not os.path.exists(BASHRC_LOCAL_PATH):
+        logging.info('%s file does not exist' % BASHRC_LOCAL_PATH)
+        # remove the enviroment variables from the env, if present
+        [os.environ.pop(env_var, {}) for env_var in env_vars]
+        logging.info(
+            '%s variables removed from the environment, if present'
+            % ' '.join(env_vars)
+        )
```
rjschwei, line 797: Not necessary to access `os.environ`. How would the variables get set if the file from which they are set does not exists. Also it makes the assumption that it would matter in the parent process from where the clean up is executed. Something we cannot know and cannot predict. If the file is not there then there is nothing to do.

lib/cloudregister/registerutils.py
```diff
@@ -717,6 +783,79 @@ def same_registry_auth_content(content, smt):
...
+    bashrc_local_new_lines = []
+    modified = False
+    for bashrc_line in bashrc_local_lines:
+        skip = False
+        for env_var in env_vars:
+            if not bashrc_line.startswith('#') and env_var in bashrc_line:
+                skip = True
+                modified = True
+        if not skip:
+            bashrc_local_new_lines.append(bashrc_line)
```
rjschwei, line 854: We do not need the skip flag, set modified and continue.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
```diff
@@ -613,10 +616,73 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+            logging.info('Could not open %s: %s' % (BASHRC_LOCAL_PATH, error))
+            failed = __mv_file_backup(BASHRC_LOCAL_PATH)
+            if failed:
+                return False
+
+    present = {}
+    for env_var_key in list(env_vars_dict):
+        for bashrc_line in bashrc_local_lines:
+            if env_var_key in bashrc_line:
+                present[env_var_key] = True
```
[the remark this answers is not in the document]
schaefi: Please also remember ```in``` is a substring match in python. Maybe not the best choice when matching key names

lib/cloudregister/registerutils.py
```diff
@@ -46,6 +46,7 @@
 REGISTERED_SMT_SERVER_DATA_FILE_NAME = 'currentSMTInfo.obj'
 RMT_AS_SCC_PROXY_MARKER = 'rmt_is_scc_proxy'
 REGISTRY_CREDENTIALS_PATH = '/etc/containers/config.json'
+BASHRC_LOCAL_PATH = '/etc/bash.bashrc.local'
```
schaefi, line 49: Is our standard bashrc provided by the distro really sourcing this file ?

lib/cloudregister/registerutils.py
```diff
@@ -613,10 +616,73 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+
+
+# ----------------------------------------------------------------------------
+def source_env():
+    """Source the updated bashrc local file for the changes to take effect."""
+    source_cmd = 'source {}'.format(BASHRC_LOCAL_PATH).split()
+    sourced_env_vars_failed = exec_subprocess(source_cmd)
+    if sourced_env_vars_failed:
+        logging.info('Source %s failed' % BASHRC_LOCAL_PATH)
+    return not sourced_env_vars_failed
```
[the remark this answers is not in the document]
schaefi: I agree we should not source the env file, sourcing should be done by the bash and the main bashrc should do this for us

schaefi (commented):
Thanks for starting this effort. I think the implementation does more than expected. imho the registration should only write the env file but sourcing is a responsibility of the shell itself. That's why I think it's important to check the suse default bashrc setup if there is already a file that gets sourced as custom extension and use and extend that one.

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
jesusbv: if you mean we should source the environment, I'm not sure I agree

the expectation for the user is to run `registercloudguest` and they expect to be able to run Podman or Docker
right after, if registration suceeded

if we do not source the file for the variables to take effect, running Podman or Docker would not work

jesusbv:
> I agree we should not source the env file, sourcing should be done by the bash and the main bashrc should do this for us

I'm not sure I agree, the expectation for the user is to run `registercloudguest` and they expect to be able to run Podman or Docker right after, if registration suceeded

if we do not source the file for the variables to take effect, running Podman or Docker would not work

lib/cloudregister/registerutils.py
jesusbv: from the `/etc/bash.bashrc`:

```
PLEASE DO NOT CHANGE /etc/bash.bashrc There are chances that your changes
# will be lost during system upgrades.  Instead use /etc/bash.bashrc.local
# for bash ....
```

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
jesusbv: I will remove the sourcing of the env as I think you both are right but do we not have the same issue ?

the variables in the file won't be present unless that file is sourced

lib/cloudregister/registerutils.py
jesusbv: > How would the variables get set if the file from which they are set does not exists

Could the user not set the environment variables, if they wanted to, manually ?

lib/cloudregister/registerutils.py
```diff
@@ -613,10 +616,73 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+        )
+        new_keys_to_add.append(env_var)
+
+    if new_keys_to_add:
+        return update_bashrc(new_keys_to_add, 'a') and source_env()
+
+
+# ----------------------------------------------------------------------------
+def update_bashrc(content, mode):
+    """Update (add or remove) the env vars for the container engines
```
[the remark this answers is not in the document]
jesusbv: I removed both `add or remove` intent should be implied from `update`

lib/cloudregister/registerutils.py
```diff
@@ -613,10 +616,58 @@ def set_registry_auth_token(registry_fqdn, username, password):
...
+    if keys_to_add:
+        return update_bashrc(keys_to_add, 'a')
+
+
+# ----------------------------------------------------------------------------
+def update_bashrc(content, mode):
+    """Update the env vars for the container engines
+    with the location of the config file to the bashrc local file."""
+    try:
+        with open(BASHRC_LOCAL_PATH, mode) as basrhc_file:
```
rjschwei, line 656: s/basrhc_file/bashrc_file/

lib/cloudregister/registerutils.py
rjschwei: The loop looks at every line and checks both variables, no skip flag needed.

```
for bashrc_line in bashrc_local_lines:
    for env_var in env_vars:
        if not bashrc_line.startswith('#') and env_var in bashrc_line:
            modified = True
            break
    else:
        bashrc_local_new_lines.append(bashrc_line)
```

lib/cloudregister/registerutils.py
```diff
@@ -717,6 +768,73 @@ def same_registry_auth_content(content, smt):
...
+    bashrc_local_new_lines = []
+    for bashrc_line in bashrc_local_lines:
+        skip = False
+        for env_var in env_vars:
+            if not bashrc_line.startswith('#') and env_var in bashrc_line:
+                skip = True
+        if not skip:
+            bashrc_local_new_lines.append(bashrc_line)
+
+    modified = bashrc_local_lines != bashrc_local_new_lines
```
rjschwei, line 834: This potentially long and therefore expensive string comparison can be skipped, we already know when we modified the, which happens when we do not re-add the entries for the environment variables.

rjschwei (changes requested):

rjschwei (approved):
One thing to think about is whether or not we need to have a separate function to clean up registry setting. I cannot think of a use case where we would clean up the registry but not the repository setup. as such we can probably drop `clean_registry_setup()` and add the calls to the `cleanup` function in `registercloudguest`. (Separate PR)

rjschwei:
And squash before merging please. @schaefi can you please take another look. Thanks

schaefi (approved):
Thanks @jesusbv this looks great now :+1:

jesusbv:
> One thing to think about is whether or not we need to have a separate function to clean up registry setting. I cannot think of a use case where we would clean up the registry but not the repository setup. as such we can probably drop `clean_registry_setup()` and add the calls to the `cleanup` function in `registercloudguest`. (Separate PR)

We added the `clean_registry_setup` to the `cleanup` method already ?
in https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L167

PR #164, opened 2024-06-12
Handle registry search order

jesusbv:
Add and remove registries conf info to search in the private and public SUSE registries

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -610,6 +610,7 @@ def set_container_engines_env_vars():
     ])
     return update_bashrc(export_registry_env_vars, 'a') and source_env()
 
+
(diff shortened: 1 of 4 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -30,6 +30,7 @@
 import subprocess
 import sys
 import time
+import toml
```
schaefi, line 33: Starting with Python 3.11, a TOML reader is part of the standard library. That said, a perfectly good module for reading and modifying TOML is tomlkit (which was created by the Poetry developers and used by Poetry itself).

schaefi (commented):
just one comment so far, it's hard to review the complete code in the current state. I'll wait until it leaves the draft mode :)

jesusbv:
close in favor of #165

PR #165, opened 2024-06-17
Handle search order for the registry

jesusbv:
- Add registries conf registry content

- Remove registries conf registry content

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -30,6 +30,7 @@ import stat
 import subprocess
 import sys
 import time
+import toml
 
 from lxml import etree
 from pathlib import Path
@@ -47,6 +48,8 @@ REGISTERED_SMT_SERVER_DATA_FILE_NAME = 'currentSMTInfo.obj'
 RMT_AS_SCC_PROXY_MARKER = 'rmt_is_scc_proxy'
 REGISTRY_CREDENTIALS_PATH = '/etc/containers/config.json'
 BASHRC_LOCAL_PATH = '/etc/bash.bashrc.local'
+REGISTRIES_CONF_PATH = '/etc/containers/registries.conf'
+DOCKER_CONFIG_PATH = '/etc/docker/daemon.json'
 
 requests.packages.urllib3.disable_warnings(
     requests.packages.urllib3.exceptions.InsecureRequestWarning
@@ -529,6 +532,8 @@ def setup_registry(registry_fqdn, username, password):
     )
     if setup_registry_succeed:
         setup_registry_succeed = set_container_engines_env_vars()
+    if setup_registry_succeed:
+        set_registries_conf(registry_fqdn)
     return setup_registry_succeed
 
 
@@ -649,6 +654,45 @@ def set_container_engines_env_vars():
 
 
 # ----------------------------------------------------------------------------
+def set_registries_conf(registry_fqdn):
+    """Add the registry container engines search configuration."""
+    return (
+        __set_registries_conf_podman(registry_fqdn) and
+        __set_registries_conf_docker(registry_fqdn)
+    )
+
+
+# ----------------------------------------------------------------------------
+def get_registry_conf_file(container_path, container):
+    registries_conf = {}
+    try:
+        with open(container_path, 'r') as registries_conf_file:
+            if container == 'podman':
+                registries_conf = toml.load(registries_conf_file)
+            if container == 'docker':
+                registries_conf = json.load(registries_conf_file)
+            return registries_conf, False
+    except IOError as error:
+        logging.info(str(error))
+        action = 'open'
+    except json.decoder.JSONDecodeError:
+        action = 'parse'
+
+    logging.info(
+        'Could not %s %s, preserving file as %s.bak' % (
+            action, container_path, container_path
+        )
+    )
+    mv_file_cmd = 'mv -Z {} {}.bak'.format(
+        container_path, container_path
+    ).split()
+    failed = exec_subprocess(mv_file_cmd)
+    message = 'File not preserved.' if failed else 'File preserved.'
+    logging.info(message)
+    return {}, failed
+
+
+# ----------------------------------------------------------------------------
 def update_bashrc(content, mode):
     """Update the env vars for the container engines
     with the location of the config file to the bashrc local file."""
@@ -668,6 +712,7 @@ def clean_registry_setup():
     """Remove the data previously set to make the registry work."""
     clean_registry_auth()
     unset_env_vars()
+    clean_registries_conf()
 
 
 # ----------------------------------------------------------------------------
@@ -710,7 +755,6 @@ def clean_registry_auth():
             return True
 
         if smt and smt.get_registry_FQDN():
-
             logging.info(
                 'Unsetting the auth entry for %s' % smt.get_registry_FQDN()
             )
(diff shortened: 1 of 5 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -836,6 +866,100 @@ def clean_bashrc_local(env_vars):
...
+
+            if modified:
+                unqualified_search_reg = new_unq_search_reg
+
+        pub_registry_fqdn = 'registry.suse.com'
+        if pub_registry_fqdn in unqualified_search_reg:
+            modified = True
+            unqualified_search_reg.pop(
+                unqualified_search_reg.index(pub_registry_fqdn)
+            )
```
rjschwei, line 913: Nitpick, the modified flag should be set after we do the modification.

lib/cloudregister/registerutils.py
```diff
@@ -836,6 +866,100 @@ def clean_bashrc_local(env_vars):
...
+
+        registries_conf['unqualified-search-registries'] = \
+            unqualified_search_reg
+
+    registries_list = registries_conf.get('registry', [])
```
rjschwei, line 918: No hungarian notation please.

rjschwei (changes requested):

lib/cloudregister/registerutils.py
```diff
@@ -648,6 +652,33 @@ def set_container_engines_env_vars():
         return update_bashrc(keys_to_add, 'a')
 
 
+# ----------------------------------------------------------------------------
+def set_registries_conf(registry_fqdn):
+    """Add the registry container engines search configuration."""
+    # only for Podman for now, Docker does not support search
```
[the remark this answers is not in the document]
jesusbv: yes, I included the Docker search in the closed first PR (long ago) but left it out because of the mislead search term

will include it back

lib/cloudregister/registerutils.py
```diff
@@ -648,6 +652,33 @@ def set_container_engines_env_vars():
...
+            'Could not open %s: %s, preserving file as %s.bak' % (
+                REGISTRIES_CONF_PATH, err, REGISTRIES_CONF_PATH
+            )
+        )
+        mv_file_cmd = 'mv -Z {} {}.bak'.format(
+            REGISTRIES_CONF_PATH, REGISTRIES_CONF_PATH
+        ).split()
+        failed = exec_subprocess(mv_file_cmd)
+        message = 'File not preserved.' if failed else 'File preserved.'
+        logging.info(message)
```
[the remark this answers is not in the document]
jesusbv: yes, that is intentional, at least to bring up this conversation

maybe it is better to have a PR after this for the refactor of replacing that code with the wrapper/method that handles that ?

rjschwei (changes requested):
Didn't quite get all the way through but at least we have an incremental review. Need to get off the computer now.

lib/cloudregister/registerutils.py
```diff
@@ -836,6 +917,199 @@ def clean_bashrc_local(env_vars):
...
+            if private_registry_fqdn in unqualified_search_reg:
+                unqualified_search_reg.pop(
+                    unqualified_search_reg.index(private_registry_fqdn)
+                )
+                modified = True
+            if pub_registry_fqdn in unqualified_search_reg:
+                unqualified_search_reg.pop(
+                    unqualified_search_reg.index(pub_registry_fqdn)
+                )
+                modified = True
```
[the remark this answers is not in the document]
jesusbv: no, because the 

```python

            if pub_registry_fqdn in unqualified_search_reg:
                unqualified_search_reg.pop(
                    unqualified_search_reg.index(pub_registry_fqdn)
                )
                modified = True
                
````

was outside that check, but given that there is a [loop](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/165/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R967)

I decided to check in the loop for `registry.suse.com` instead of loop + _another_ operation to check

File: lib/cloudregister/registerutils.py [hunk 6]
```diff
@@ -648,6 +653,82 @@ def set_container_engines_env_vars():
...
+        container_path, container_path
+    ).split()
+    failed = exec_subprocess(mv_file_cmd)
+    message = 'File not preserved.' if failed else 'File preserved.'
+    logging.info(message)
+    return {}, failed
+
+
+# ----------------------------------------------------------------------------
+# def write_registry_conf_file(info, container_path, container):
```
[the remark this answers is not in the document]
jesusbv: it is a duplicate that I forgot to remove

lib/cloudregister/registerutils.py
```diff
@@ -2042,3 +2279,118 @@ def __mv_file_backup(filename):
...
+            unqualified_search_reg.insert(0, fqdn) for fqdn in
+            [public_registry_fqdn, private_registry_fqdn]
+        ]
+
+    private_registry = {'location': private_registry_fqdn, 'insecure': False}
+    # a list of dictionaries
+    registry_mirrors = registries_conf.get('registry', [])
+    present = False
+    if private_registry in registry_mirrors:
+        # exact match
```
schaefi, line 2327: matching with ```in``` is a substring match not an exact match as the comment indicates, is this correct ?

lib/cloudregister/registerutils.py
```diff
@@ -2042,3 +2279,118 @@ def __mv_file_backup(filename):
...
+            priv_index = unqualified_search_reg.index(private_registry_fqdn)
+        if public_registry_fqdn in unqualified_search_reg:
+            pub_index = unqualified_search_reg.index(public_registry_fqdn)
+
+        if not priv_index == 0 or not pub_index == 1:
+            if priv_index > 0:
+                unqualified_search_reg.pop(priv_index)
+            if pub_index > 0:
+                unqualified_search_reg.pop(pub_index)
+            modified = True
```
schaefi, line 2314: This logic in the ```__set_``` method drops an entry under certain conditions prior adding data. I wonder if it would be possible to just call ```clean_registries_conf_podman``` in ```__set_registries_conf_podman``` and keep only one code path that actually drops entries ?

lib/cloudregister/registerutils.py
jesusbv: but `registry_mirrors` is a list of dictionaries ?
so that `if` checks the following

`{'location': registry-ec2.susecloud.net', 'insecure': False} in [{'location': 'foo'}, {'location': registry-ec2.susecloud.net', 'insecure': False}]`

lib/cloudregister/registerutils.py
jesusbv: so instead of 

- open file
- check content
- modify (drop and add) content if needed
- write file

you suggest

- open file 
- drop content
- write file
- open file
- add content
- write file
?

jesusbv: having one code path to drop entries seems fine to me, just wonder if it is not that cheap ?

lib/cloudregister/registerutils.py
schaefi: ok for this datatype match it's exact and correct. All good sorry for the noise

lib/cloudregister/registerutils.py
schaefi: I was making the suggestion just for logic and consistency reasons. I agree with you that your current implementation is more performant. The part that matters most should be followed and I leave this decision up to you. I'm fine either way. In this particular case I believe less code makes it easier to prevent pitfalls due to the many ```if``` conditions and I think performance is not too much important at this point. So I looked at it more from a transaction perspective in a way that you can set what is needed when you start from a clean state

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
rjschwei: But this code does loop + another operation. If we have the fqdn for the update infrastructure we check to remove it and we check to remove registry.suse.com, if we do not have the fqdn of the update server we check for the domain name of the update infrastructure **and** we check for registry.suse.com. Meaning we check for registry.suse.com in the `if` and the `else` block. Meaning we do the same operation in any condition and as such that operation should then happen unconditionally.

lib/cloudregister/registerutils.py
rjschwei: At first glance my gut feel is that the logic can be simplified. But I do not have enough peace an quiet to think about what that would look like. We will need some refactoring paths over the code anyway. As such I'd tend to leave this as is at the moment.

rjschwei (approved):
Some changes needed that can come in a re-factoring effort.

Please squash before merging.

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
jesusbv: yes, the operation happening unconditionally was as it was, thought to piggy back on the for-loop

will change it in the refactor, as you mentioned

schaefi (approved):
Ok, as it seems the path forward is the agreement that there is room for improvement on a refactoring I'm fine with the current state too.

PR #166, opened 2024-06-19
Handle SUMA registry file

jesusbv:
- Add SUSE registry FQDN to the SUMA if not present

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -31,6 +31,7 @@ import subprocess
 import sys
 import time
 import toml
+import yaml
 
 from lxml import etree
 from pathlib import Path
@@ -50,6 +51,7 @@ REGISTRY_CREDENTIALS_PATH = '/etc/containers/config.json'
 BASHRC_LOCAL_PATH = '/etc/bash.bashrc.local'
 REGISTRIES_CONF_PATH = '/etc/containers/registries.conf'
 DOCKER_CONFIG_PATH = '/etc/docker/daemon.json'
+SUMA_REGISTRY_CONF_PATH = '/etc/uyuni/uyuni-tools.yaml'
 
 requests.packages.urllib3.disable_warnings(
     requests.packages.urllib3.exceptions.InsecureRequestWarning
@@ -656,10 +658,15 @@ def set_container_engines_env_vars():
 # ----------------------------------------------------------------------------
 def set_registries_conf(registry_fqdn):
     """Add the registry container engines search configuration."""
-    return (
+    container_engines_conf_set = (
         __set_registries_conf_podman(registry_fqdn) and
         __set_registries_conf_docker(registry_fqdn)
     )
+    suma_registry_conf_set = True
+    if is_suma_instance():
+        suma_registry_conf_set = __set_registry_fqdn_suma(registry_fqdn)
+
+    return container_engines_conf_set and suma_registry_conf_set
 
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -2027,6 +2035,41 @@ def has_network_access_by_ip_address(server_ip):
     return True
 
 
+# ----------------------------------------------------------------------------
+def is_suma_instance():
+    """Return a list containing the SUMA product if present."""
+    products_list = glob.glob('/etc/products.d/*.prod')
+    suma_prod = [
+        product for product in products_list if 'suse' in product.lower() and
+        'manager' in product.lower()
```
schaefi, line 2044: Let's hope that suse has no product that contains 'suse' and 'manager' in its name. Call me pedantic but couldn't we match on a clear name ? I'm not too deep into suma but I can imagine it has a unique identifier ?

lib/cloudregister/registerutils.py
```diff
@@ -2394,3 +2437,90 @@ def __set_registries_conf_docker(private_registry_fqdn):
...
+        return True
+
+    suma_yaml_content = {}
+    suma_yaml_content, failed = get_suma_registry_content()
+    if failed:
+        return False
+
+    if not suma_yaml_content or suma_yaml_content == {'registry': ''}:
+        # empty content or no other content than registry key and empty value
+        os.unlink(SUMA_REGISTRY_CONF_PATH)
```
schaefi, line 2493: Would removing that file harm SUMA ?

lib/cloudregister/registerutils.py
```diff
@@ -2394,3 +2437,90 @@ def __set_registries_conf_docker(private_registry_fqdn):
...
+            # registry value does not contain registry nor susecloud.net
+            return True
+
+    if suma_yaml_content.pop('registry', False):
+        if suma_yaml_content:
+            logging.info('Updating %s after removing registry entry')
+            return __write_suma_conf(suma_yaml_content)
+
+    # no other content than the removed registry key
+    os.unlink(SUMA_REGISTRY_CONF_PATH)
```
schaefi, line 2521: Same here, would removing the file harm the functionality of SUMA itself ?

schaefi (commented):
Thanks Jesus :+1: Implementation looks good but I'm unsure if there is maybe a more stable way to detect the product and I don't know if we can remove /etc/uyuni/uyuni-tools.yaml without harming SUMA itself ? Maybe something to check prior merge ?

lib/cloudregister/registerutils.py
jesusbv: we can check the exact match which is `SUSE-Manager-Server.prod` as of today

lib/cloudregister/registerutils.py
jesusbv: When I asked that I ws told if the file is not present SUMA would use the builtin defaults, however, the image
should contain that file and it should not get removed

Will edit the code to reflect that

jesusbv: if the file should be removed and the file should be baked in with the image + containing the registry FQDN
maybe should only add it if not present and let it be, meaning not removing that entry when deregistering 

Thoughts ?

lib/cloudregister/registerutils.py
jesusbv: Updated the check for SUMA instance

lib/cloudregister/registerutils.py
jesusbv: I have removed the SUMA file cleanup, we set the value if not present as the value should be present and `cloud-regionsrv-client` does not touch that file once value is set

lib/cloudregister/registerutils.py
```diff
@@ -2027,6 +2034,42 @@ def has_network_access_by_ip_address(server_ip):
...
+            suma_reg_config_data = yaml.safe_load(suma_reg_config)
+        return suma_reg_config_data, False
+    except IOError as error:
+        logging.info(str(error))
+        action = 'open'
+    except yaml.YAMLError:
+        action = 'parse'
+
+    logging.info('Could not %s %s' % (action, SUMA_REGISTRY_CONF_PATH))
+    failed = __mv_file_backup(SUMA_REGISTRY_CONF_PATH)
```
jesusbv, line 2069: I'm not 100 % sure we should move the file to a backup, that would make `cloud-regionsrv-client` the owner of that file and IMO it should be owned by SUMA

that would mean we log that file could not be opened or parsed and no further action

lib/cloudregister/registerutils.py
schaefi: perfect, so that's clear then. Thanks

schaefi: Another point that we need to clarify. When we write that file, what needs to be done that SUMA reads it ? Maybe we need to send a signal or restart a service, or at least document the needed steps regarding suma after the registration. @jesusbv you mentioned a meeting with the SUMA people, can you ask that question there ? Thanks

lib/cloudregister/registerutils.py
schaefi: I agree, the backup strategy in this case might be unexpected. Just the logging is imho enough

lib/cloudregister/registerutils.py
schaefi: I like @rjschwei feedback regarding this matching

schaefi: my gut feeling is, I would just check for the exact .prod file to identify the product and that's it

schaefi: given the suse madness about names I might be wrong here, but in that case matching for 'suse' and 'manager' could also fail when they decide to name it suse-mgr or s.u.s.e.-mgr or maybe some other crazy name to make live of developers hard :-))

jesusbv: the [code](href="https://confluence.suse.com/display/publiccloud/Roadmap">https:/) has changed ? 
can you see the new code, @schaefi ?

rjschwei (approved):

jesusbv:
will squash the commits and merge

PR #167, opened 2024-06-20
Update and extend test matrix

schaefi:
Make sure we test our code also against python version 3.11 which has landed in SLES too. Also fix the tox setup as well as add a proper pytest config and add one missing test for the code path if no network connection is possible.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -4,11 +4,16 @@ on: [push]
 
 jobs:
   unit_tests:
-    name: Linter checks
-    runs-on: ubuntu-20.04
+    name: Unit tests for cloud regionsrv client
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: ["3.11"]
+        os: [ubuntu-latest]
+        include:
+          # Use ubuntu 20.04 for python 3.6 (dropped from later versions)
+          - python-version: 3.6
+            os: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v3
@@ -20,9 +25,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox
-      - name: Run code checks
-        run: |
-          tox -e check
-          tox -e "unit_py${PY_VER/./_}" -- -n auto
+      - name: Run linter and tests
+        run: tox -e "check,unit_py${PY_VER/./_}"
         env:
           PY_VER: ${{ matrix.python-version }}
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,4 @@
+[tool:pytest]
+norecursedirs = .git build .tox/ .tmp/
+addopts = --ignore=.tmp/ --ignore=.git/ --ignore=.tox/ -p no:warnings
+testpaths = tests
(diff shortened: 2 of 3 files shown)


tests/test_registerutils.py
```diff
@@ -3239,6 +3239,12 @@ def test_remove_service(
     mock_logging.info.not_called()
 
 
+@patch('socket.create_connection')
+def test_network_access_error(mock_socket_create_connection):
```
jesusbv, line 3243: I'm not against this explicit test

Just mentioning that that code path is tested in `test_fetch_smt_data_api_no_valid_ip` 
Maybe we want to extend that test to check the log message ?

.github/workflows/ci-testing-legacy.yml
```diff
@@ -0,0 +1,30 @@
+name: CI-Code-Style
+
+on:
+  push:
+    branches:
+      - "master"
+  pull_request:
+
```
rjschwei, line 8: I think we want to run the CI on PRs. There are generally not that many

.github/workflows/ci-testing.yml
```diff
@@ -1,14 +1,18 @@
...
 jobs:
   unit_tests:
-    name: Linter checks
-    runs-on: ubuntu-20.04
+    name: Unit tests for cloud regionsrv client
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: ["3.11"]
```
rjschwei, line 15: cloud-regionsrv-client still only builds for Python 3.6. I think we should make this change when/if we switch the package build to Python 3.11

jesusbv: my understanding when we talked about the PR, was that a new test env would be aded
so one could run `tox -e unit_py3_6` or `tox -e unit_py3_11` not to remove 3.6 which we still need AFAIK

tests/test_registerutils.py
schaefi: I think I'm going to drop this test from here as you covered it in another PR that you did after I created this one

.github/workflows/ci-testing-legacy.yml
schaefi: I can't follow you ? This is what the workflow definition does, it's called on any pull request and it operates on master after merge. You can see that it does run on pull requests, because it ran on this pull request :)

.github/workflows/ci-testing.yml
schaefi: nothing got removed. With this PR you have now two workflow files, one for legacy (3.6) and one for the future 3.11 onwards. I was not able to handle 3.6 + 3.11 + more in one workflow file because the ubuntu version that is used to run the containers for the workflow run does not offer all this python versions. In fact py3.6 only exists for an older ubuntu container which we select in the workflow definition to test the 3.6 run and for 3.11 we have to select a newer ubuntu container but that one does not longer provide 3.6. I think it is not possible to select a different container per python version in one workflow file

jesusbv: got it, it is in legacy

tests/test_registerutils.py
schaefi: done

jesusbv (approved):
LGTM

PR #168, opened 2024-06-20
Mock all socket connections

jesusbv:


schaefi (approved):
Looks good :+1: I think this one also covers the test I wrote in #167

rjschwei (approved):

PR #169, opened 2024-06-20
Refactor profile reading

schaefi:
The current implementation to point tools like docker, podman, etc to our infrastructure registry is based on exporting a number of environment variables which are read by the individual tools. The file written so far was placed to ```/etc/bash.bashrc.local```. However, there is no mechanism in the system which would automatically source this file on a new login. So far only the console which has started registercloudguest or the respective service at boot time has the right environment set up. This commit changes the location of the env file to ```/etc/profile.local```. This is a system wide profile file for custom settings in addition to /etc/profile. The file is automatically sourced on every login or ssh login which allows us to make sure our settings are active in every login session to the instance.


rjschwei (approved):

jesusbv (approved):
LGTM

PR #171, opened 2024-06-21
Refactor to use the backup creation method

jesusbv:


lib/cloudregister/registerutils.py
```diff
@@ -2271,7 +2244,8 @@ def __generate_registry_auth_token(username=None, password=None):
 
 
 # ----------------------------------------------------------------------------
-def __mv_file_backup(filename):
+def __create_file_backup(filename):
+    """Create a backup of the file filename, return the status of the cmd."""
     message = ('Preserving file as %s.bak' % filename)
     logging.info(message)
     mv_file_cmd = 'mv -Z {} {}.bak'.format(filename, filename).split()
```
jesusbv, line 2251: I know we agreed to use `mv -Z` but maybe not all backups need the attribute and maybe some can be
done with `shutil.move`, if so, the method can be like:

```python
def __create_file_backup(filename, preserve_attribute=False):
    if not preserve_attribute:
        shutil.move(....)
    else:
        cmd = 'mv -Z....'
    ...
```

rjschwei: If we are on a system that has SELinux enabled then the use of `-Z` is not file dependent. The context is system wide.

schaefi:
@jesusbv this one has conflicts which should be resolved. Are you still up for this PR or should this refactoring be done later ? Thanks

schaefi:
I guess the answer is later :) closing for later inspection

PR #175, opened 2024-06-25
Add the call to set up the registry

jesusbv:
Re adding the call in registercloudguest


schaefi (approved):
Looks good :+1: 

I would add a bit more details in the commit message, saying that on accident the setup_registry() call was removed and that this patch fixes it.

PR #176, opened 2024-06-25
Add registry info to be checked to the integration test

jesusbv:


integration_test-process.txt
```diff
@@ -15,22 +15,44 @@ After installing the test package with "zypper in"
   + registration success
     - check log
   + zypper lr has repos
+  + podman search registry-<csp>.susecloud.net returns some value
```
jesusbv, line 18: Maybe it is easier to check the info is present on the system instead of installing Podman and running the command ?

schaefi: I'd say both

```
+ check `/etc/containers/registries.conf` for registry-<csp>.susecloud.net
+ read catalog `podman search registry-<csp>.susecloud.net:/`
```

rjschwei: But we expect the user to not have to know about `registry-<csp>.susecloud.net` which is why we added significant amount of code to setup podman configuration. I am fine with having the search as part of the test procedure but then it should be `podman search $SOMECONTAINER`. 

That said, the problem is that the containers we currently mirror to the update infrastructure are all associated with a given subscription and at present there are no such containers for a general SLES or SLES4SAP subscription, as such the search `podman search $SOMECONTAINER` would by necessity have to be for a container that is free and on registry.suse.com. At the very least it will test that we didn't break the podman configuration.

schaefi:
@jesusbv @rjschwei this one has conflicts and seems to be stuck. Can I get an estimate how you would like to proceed ? Thanks

jesusbv:
> @jesusbv @rjschwei this one has conflicts and seems to be stuck. Can I get an estimate how you would like to proceed ? Thanks

My opinion is that, yes there are registry operation on the test now (`podman pull`) but I think the search command must be tested as well
like regular PAYG instance should have access to one product and PAYG + LTSS activated should have access to the same product + LTSS containers for example

rjschwei:
I stick with what I stated previously, we extended a bunch of effort to make `podman search $CONTAINER_NAME` work, and it does:
```
# podman search bci-base
NAME                                 DESCRIPTION
registry.suse.com/bci/bci-base       
registry.suse.com/bci/bci-base-fips  
```
What is the goal of the integration test?

1.) Do we want to check that the search setup for podman via the configuration written by the client is correct, as suggested by @jesusbv in that case the integration test procedure should include just this

```
- podman search bci-base  should return results from registry.suse.com
```
And ultimately, with this test procedure we are testing the client, as such this should be sufficient.

2.) Do we want to check that the "right" containers are mirrored?
This IMHO is not the job of the client integration test process and should be tested in some other fashion. But could be done with a `podman search suse/ltss/sle15.4/sle15` (I have not tested if that works but IMHO it should.) However that requires additional steps in the LTSS testing path.

The proposed change does not work for me, as stated we extended the effort to hide `registry-$CSP.susecloud.net` and therefore we should test that this effort is rewarded. If we expect that users know the registry then we can pull out code staring here: https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L910

schaefi:
This one is almost a year old, has conflicts and no agreement was achieved. I don't think we will be making any significant success steps with this PR. Thus I'm going to close this one and we will revise the topic either in an issue or via a conversation within the team.

Thanks

PR #177, opened 2024-06-27
Update infrastructure registry re-auth

rjschwei:
Switch back to using Python as a driver and create a sudoers rule for the script such that a non root user can execute the re-authentication.


jesusbv (approved):
Added the copyright and updated the string due to a typo

LGTM

mgerstner:
You might want to add some transparent sudo reexecution logic to the script like:

```
if os.geteuid() != 0:
    res = subprocess.call(["sudo", sys.argv[0]])
    sys.exit(res)
```

This way the command can be used without having to deal with sudo explicitly.

jesusbv:
> You might want to add some transparent sudo reexecution logic to the script like:
> 
> ```
> if os.geteuid() != 0:
>     res = subprocess.call(["sudo", sys.argv[0]])
>     sys.exit(res)
> ```
> 
> This way the command can be used without having to deal with sudo explicitly.

Good call, I added the root check back and also the message to inform the user

schaefi:
I agree to @mgerstner suggestion. As it's expected that the command is managed through sudo we should make this explicit. I would do it as follows:

```
#!/usr/bin/sudo /usr/bin/python3

import os

print(f'EUID: {os.geteuid()}')
```

schaefi (approved):
Given we have a transparent execution through sudo this looks good :+1:

rjschwei:
Thanks @jesusbv  and @schaefi for handling the remaining details. @jesusbv please build the package and submit to Maintenance.

mgerstner:
> I agree to @mgerstner suggestion. As it's expected that the command is managed through sudo we should make this explicit. I would do it as follows:
> 
> ```
> #!/usr/bin/sudo /usr/bin/python3
> 
> import os
> 
> print(f'EUID: {os.geteuid()}')
> ```

By doing this via the shebang line, the sudoers rule no longer matches the actual execution of the script. Sudo will be called like this `/usr/bin/sudo /usr/bin/python3 /usr/bin/cloudregistryauth`, and as a result, the root password will be required for execution.

The shebang approach also means if the caller is already `root`, that the script will still go through `sudo`, needlessly. The shebang line also has restrictions on the way parameters are passed to the "interpreter". I'm not sure that the shebang approach is a good idea. A caller might also want to pass additional `sudo` arguments in some circumstances. Thus keepin the sudoers rule as it is currently seems the better solution to me.

PR #179, opened 2024-06-28
Fix misnamed method

jesusbv:


schaefi (approved):
consistent naming, I like it :+1:

PR #180, opened 2024-06-28
Proper sudo call

jesusbv:
- Remove local from sudoers content path
- Call command with 'usod' prefix if not sudo

### DIFF
--- a/usr/bin/cloudguestregistryauth
+++ b/usr/bin/cloudguestregistryauth
@@ -1,4 +1,4 @@
-#!/usr/bin/sudo /usr/bin/python3
+#!/usr/bin/python3
 
 # Copyright (c) 2024, SUSE LLC, All rights reserved.
 #
@@ -20,12 +20,21 @@ feature.
 The script is used in a sudo rule to allow its use by a non root user
 """
 
+import os
 import sys
 
-from cloudregister.registerutils import get_activations
+from cloudregister.registerutils import (
+    get_activations, exec_subprocess
+)
 
-
-if not get_activations():
-    sys.exit('Could not refresh credentials')
+error_message = 'Could not refresh credentials'
+if os.geteuid():
+    if exec_subprocess(['sudo', sys.argv[0]]):
+        # geteuid is not root and
+        # the command was called with sudo
+        # and failed
+        sys.exit(error_message)
+elif not get_activations():
+    sys.exit(error_message)
 
 print('Credentials refreshed')
--- a/tox.ini
+++ b/tox.ini
@@ -47,4 +47,6 @@ skip_install = True
 usedevelop = True
 commands =
     flake8 --statistics -j auto --count {toxinidir}/lib/cloudregister
+    flake8 --statistics -j auto --count \
+        {toxinidir}/usr/bin/cloudguestregistryauth
     flake8 --statistics -j auto --count {toxinidir}/tests
--- a/usr/etc/sudoers.d/cloudguestregistryauth
+++ b/usr/etc/sudoers.d/cloudguestregistryauth
@@ -1,1 +1,1 @@
-ALL ALL=(root) NOPASSWD: /usr/local/bin/cloudguestregistryauth ""
+ALL ALL=(root) NOPASSWD: /usr/bin/cloudguestregistryauth ""


jesusbv:
@mgerstner ^^

Also, we missed `/usr/local/bin/cloudguestregistryauth` in #177 review, fixed here + placed that file under /etc

usr/bin/cloudguestregistryauth
```diff
@@ -20,12 +20,21 @@ feature.
...
+    get_activations, exec_subprocess
+)
+
+error_message = 'Could not refresh credentials'
+if os.geteuid():
+    if exec_subprocess(['sudo', sys.argv[0]]):
+        # geteuid is not root and
+        # the command was called with sudo
+        # and failed
+        sys.exit(error_message)
```
rjschwei, line 36: @mgerstner this is based on https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/177#issuecomment-2194352748 .Can you please confirm my understanding.

A user that is not setup as a sudo user runs `cloudguestregistryauth`, we then detect the executing user not to be root and trigger `sudo cloudguestregistryauth` in a sub process. sudo in turn will find the rule for this script and allow execution in a root user context?

Is this correct?

rjschwei (approved):
Conditionally approved pending comment from @mgerstner

usr/bin/cloudguestregistryauth
mgerstner: Mostly correct, yes, except for:

> A user that is not setup as a sudo user runs [...]

Any user can run the script. Only if the caller is not already root, then reexecution using `sudo` happens.

The sudoers rule you currently have allows _any_ user to run this script as root, therefore no special configuration is needed. As I outlined in the related Bugzilla bug, you could consider requiring membership in a dedicated group for hardening purposes. The sudoers rule would then need to look like this:

```
%cloudgroup ALL=(root) NOPASSWD: /usr/bin/cloudguestregistryauth ""
```

Otherwise the script and rule in this PR# look good to me now.

jesusbv:
@rjschwei @mgerstner I plan to include the change in this PR, checking with you first

having 
```bash
> cat /usr/etc/sudoers.d/cloudguestregistryauth
ALL ALL=(root) NOPASSWD: /usr/bin/cloudguestregistryauth ""
> sudo useradd foo
> sudo passwd foo
> su -l foo
foo@hostname > /usr/bin/cloudguestregistryauth
We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

For security reasons, the password you type will not be visible.

[sudo] password for root: 
```
asks for the root password

Maybe I missed some step, moving the file from `/usr/etc/sudoers.d` to `/etc/sudoers.d` works as expected and intended, meaning no asking for root password but running the command

mgerstner:
> Maybe I missed some step, moving the file from `/usr/etc/sudoers.d` to `/etc/sudoers.d` works as expected and intended, meaning no asking for root password but running the command

Are running this on SLE-15? Then I guess this is my fault, I suggested the path in `/usr/etc` thinking of Tumbleweed, where packaged configuration files should all go to `/usr/etc` by now; but SLE-15 doesn't support that yet.

mgerstner:
> Meaning, AFAIU even if the system is immutable, we are on the safe side adding the file under `/etc`

I'm not sure about the exact context of this piece of documentation. For Tumbleweed and newer products, though, the idea is that everything coming from the OS vendor is placed under /usr/etc, so that the system admin can place any overrides into /etc. This way during package updates anything /usr/etc can be replaced unconditionally, while the local site configuration / override remains intact in /etc.

jesusbv:
> > Meaning, AFAIU even if the system is immutable, we are on the safe side adding the file under `/etc`
> 
> I'm not sure about the exact context of this piece of documentation. For Tumbleweed and newer products, though, the idea is that everything coming from the OS vendor is placed under /usr/etc, so that the system admin can place any overrides into /etc. This way during package updates anything /usr/etc can be replaced unconditionally, while the local site configuration / override remains intact in /etc.

@mgerstner yes, I understand that but currently locating the file under `/usr/etc/sudoers.d` has no effect and running the command with the file under `/usr/etc/sudoers.d` path requires root password

mgerstner:
> @mgerstner yes, I understand that but currently locating the file under `/usr/etc/sudoers.d` has no effect and running the command with the file under `/usr/etc/sudoers.d` path requires root password

No worries, as I said before, on SLE codestreams this feature has not yet been implemented. `sudo` has likely been patched on Tumbleweed to look in /usr/etc.

I just wanted to clarify the semantics of the /usr/etc approach, didn't want to say you need to do it this way on SLE based systems.

PR #181, opened 2024-07-01
Clean for BYOS fails

jesusbv:
When registering or deregistering a BYOS unregistered instance, the cache SMT object does not have the attribute for registry FQDN

### DIFF
--- a/lib/cloudregister/smt.py
+++ b/lib/cloudregister/smt.py
@@ -107,7 +107,7 @@ class SMT:
     # --------------------------------------------------------------------
     def get_registry_FQDN(self):
         """Return the fully qualified domain registry name"""
-        return self._registry_fqdn
+        return self._registry_fqdn if hasattr(self, '_registry_fqdn') else ''
 
     # --------------------------------------------------------------------
     def get_name(self):


jesusbv:
@rjschwei @mgerstner I plan to include the change in this PR, checking with you first

having 
```bash
> cat /usr/etc/sudoers.d/cloudguestregistryauth
ALL ALL=(root) NOPASSWD: /usr/bin/cloudguestregistryauth ""
> sudo useradd foo
> sudo passwd foo
> su -l foo
foo@hostname > /usr/bin/cloudguestregistryauth
We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

For security reasons, the password you type will not be visible.

[sudo] password for root: 
```
asks for the root password

Maybe I missed some step, moving the file from `/usr/etc/sudoers.d` to `/etc/sudoers.d` works as expected and intended, meaning no asking for root password but running the command

schaefi (approved):
LGTM

PR #182, opened 2024-07-24
Fix invalid env

schaefi:
**Add registercloudguest wrapper**
    
To enforce that we are running in a proper environment wrap the call of the ```registercloudguest python``` program into a ```registercloudguest``` shell program which ensures to spawn a new shell env after the job is done and also make sure we don't end in a shell process that holds in improper environment.


schaefi:
@jesusbv @rjschwei The overall diff is hard to read because of the rename of the file. Please review commit based which makes it easy because the real change is small. Thanks

schaefi:
I tested this change in my container based dev infrastructure as it's not cloud specific and it worked there

usr/sbin/registercloudguest
```diff
@@ -1,682 +1,19 @@
-#!/usr/bin/python3
+#!/bin/bash
+set -e
 
-# Copyright (c) 2024, SUSE LLC, All rights reserved.
```
jesusbv, line 3: why is the copyright block removed ?

jesusbv:
LGTM but it has a side effect, when the command finishes it ends up in `sudo` even when the original caller was a regular user using `sudo cmd`

usr/sbin/registercloudguest
```diff
@@ -1,682 +1,19 @@
...
-# Lesser General Public License for more details.
-# You should have received a copy of the GNU Lesser General Public
-# License along with this library.
+# store pid of parent shell
+parent=$PPID
 
-"""This script obtains information from the configured region server in the
-   cloud environment and uses the information to register the guest with
-   the SMT server based on the information provided by the region server.
+registercloudguest.py "$@"
```
jesusbv, line 7: should we be explicit on the path ?
`/usr/sbin/registercloudguest.py`

jesusbv:
OK with [06d5161](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/182/commits/06d51618f1d66d3e73d1e71b66bf52d052880c5c) the sudo issue is gone and it works as expected when called from a regular user, however, if one runs `sudo su` and calls the cmd, after completion it lands on non sudo user
I think that is confusing

Also, in both scenarios, the new shell (as expected) has no history, which, taking into account the approach, is the expected behavior but, IMHO, it would definitely feel strange for the user

usr/sbin/registercloudguest
schaefi: good point, now added

usr/sbin/registercloudguest
schaefi: Hmm, it's unlikely that we change the location, ok let's make it sticky. done

schaefi:
>Also, in both scenarios, the new shell (as expected) has no history

yes, I think this is inevitable as we intentionally start a new login shell to read all env files and that starts with an empty history. We have to live with that but I think it's acceptable

schaefi:
@jesusbv something to discuss. With this change we don't want ```/usr/sbin/registercloudguest.py``` to be called by users. How about changing the permission mode to non-exec for this file and adapt ```/usr/sbin/registercloudguest``` to call the python code as

```
python3 /usr/sbin/registercloudguest.py "$@"
```

Thoughts ?

jesusbv:
> @jesusbv something to discuss. With this change we don't want `/usr/sbin/registercloudguest.py` to be called by users. How about changing the permission mode to non-exec for this file and adapt `/usr/sbin/registercloudguest` to call the python code as
> 
> ```
> python3 /usr/sbin/registercloudguest.py "$@"
> ```
> 
> Thoughts ?

I like that idea, yes

schaefi:
> I like that idea, yes

ok done

schaefi:
Also added a check if stdin/stdout/stderr are open and connected to a terminal. Only in this case a new shell is opened. We need that to avoid a new interactive shell session when we are not coming from an interactive terminal session

rjschwei:
What if we approach the problem differently?

Let's say we modify the early exit condition https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L409, which we have to do anyway, see #183 . That means a call of `registercloudguest` will setup the registries. As such we can then add a `registercloudguest` call to `%post` in  the sepc file and the since that happens during upgrade all is well. Yes the user still doesn't get the environment in the same shell where "zypper up" was run, but this applies to any shell environment that is running. I think we are working too hard for a case that we cannot completely solve anyway. It may be better to simply write a message when we add the env settings, i.e. net new setting.

"""
The shell environment was changed please restart any open shell session.
"""

schaefi:
Your look from a different angle is surely correct but it cannot invalidate the following conditions:

* registercloudguest is a command that can be called by a user from any type of shell session
* the latest implementations to registercloudguest were based on creating custom environment files which implies the software and/or the components configured by the software can only work in this environment

From that perspective I would like to make sure that IF `registercloudguest` is called from a connected console that we also make sure that we don't leave the user in a non working environment. As you cannot prevent this situation as long as `registercloudguest` stays a user tool I vote for the proposed solution.

> I think we are working too hard for a case that we cannot completely solve anyway

I think the change is maintainable and with the current code I do not see a way how users would end up in a broken env

schaefi:
I will build an intermediate cloud-regionsrv-client package and check if everything continues to work

rjschwei:
> Your look from a different angle is surely correct but it cannot invalidate the following conditions:
> 
> * registercloudguest is a command that can be called by a user from any type of shell session

Well it is only in the root users path as it lives in `/usr/sbin` and it requires root permission, i.e. has to be executed as sudo or when logged in directly as root. We only made the re-auth script executable by all users and it lives in `/usr/bin` for that reason.

Whether or not it is a good idea that root runs podman command is a different conversation but we can expect that some users will, as such I am not completely opposed to this approach I am just not convinced yet that fixing one shell, i.e. the one where the root user ran `zypper up cloud-regionsrv-client` is worth the we are undertaking. Never mind having to remember to change the called Python interpreter when it gets switched or where /usr/bin/python3 may longer exist. If we keep the executable script to be the python code that gets set automatically for us during package build. If we tale that away we end up with if conditions in the spec file. It gets complicated and I am not convinced all that complication is worth the benefit of fixing 1 shell session.

> * the latest implementations to registercloudguest were based on creating custom environment files which implies the software and/or the components configured by the software can only work in this environment
> 
> From that perspective I would like to make sure that IF `registercloudguest` is called from a connected console that we also make sure that we don't leave the user in a non working environment. As you cannot prevent this situation as long as `registercloudguest` stays a user tool I vote for the proposed solution.
> 
> > I think we are working too hard for a case that we cannot completely solve anyway
> 
> I think the change is maintainable and with the current code I do not see a way how users would end up in a broken env

Well if the sysadmin runs "zypper up cloud-regionsrv-client" and then tells all the other system users that are currently on the system "you now have access to SUSE containers" they will all be broken. I am not convinced that fixing the root users shell is worth the complication in maintaining the package.

schaefi:
>Well if the sysadmin runs "zypper up cloud-regionsrv-client" and then tells all the other system users that are currently on the system "you now have access to SUSE containers" they will all be broken.

Where in the current implementation do you see that we do this ? Sorry I can't follow you. Given the current implementation would be in place and that %post you mentioned too. What happens if you call that zypper up command is this

```
zypper up cloud-regionsrv-client
... the thing gets installed and registercloudguest gets called, all information from the python code of registercloudguest lands in /var/log/cloudregister
... on this console and on all other consoles a wall message appears:

Broadcast message from root@....

registercloudguest modified the environment, please re-login

... and on this console you will also see

This shell will automatically update the environment...

... back at your prompt

```

So where do you see a problem ?

schaefi:
> Never mind having to remember to change the called Python interpreter when it gets switched or where /usr/bin/python3 may longer exist.

we call `python3 some_code` and you expect this to go away ? There will never be a python4 according to the author and python2 is dead so what is the alternative to calling a python3 program via the python3 interpreter ?

schaefi:
ok so there is a problem with this approach that we probably don't want.

If there is a program or script connected to a terminal and it calls `registercloudguest` it will now be a blocking call which is a behavior change compared to the current code. We don't want that so let's not follow my idea and use a simple message from the python code

I will close this one and bring back the message you proposed in the other PR

PR #183, opened 2024-07-25
Preparation for ltss

schaefi:
This patch is two fold

**Fix a missing exit condition**
    
In case an alternative update server was found and the system is properly registered there is not more to do. We found this over a conversation with Robert and I thought let's fix it here

**Update the container registry code path**
    
Change the code that runs the container registry registration process  to check if the registry has already been setup. At the moment  there is no check and the only reason why it's not called multiple times are the early exit conditions which we have
to modify for the LTSS use cases.

This is in line with the dev plan as one of the first actions to prepare the code such that we can add the LTSS required additional code without causing too many bad after effects

The show can begin ;)

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1104,6 +1104,36 @@ def get_credentials_file(update_server, service_name=None):
 
 
 # ----------------------------------------------------------------------------
+def get_hosts():
+    """
+    Read /etc/hosts file into the following structure
+
+    {
+        hostname: {
+            'address': some,
+            'ipv6': true|false,
+            'ipv4': true|false
+        }
+    }
+    """
+    with open(HOSTSFILE_PATH, 'r') as hosts_fd:
+        hostlines = hosts_fd.readlines()
+    hostlines = [line.strip() for line in hostlines
+                 if not line.startswith('#') and line.strip() != '']
+    hosts = {}
+    for line in hostlines:
+        hostnames = line.split('#')[0].split()[1:]
+        hostaddress = line.split('#')[0].split()[0]
+        for hostname in hostnames:
+            hosts[hostname] = {
+                'address': hostaddress,
+                'ipv6': True if '::' in hostaddress else False,
+                'ipv4': False if '::' in hostaddress else True
+            }
+    return hosts
+
+
+# ----------------------------------------------------------------------------
 def get_current_smt():
     """Return the data for the current SMT server.
        The current SMT server is the server against which this client
@@ -1619,6 +1649,37 @@ def import_smt_cert(smt):
             )
 
     return 1
+
+
+# ----------------------------------------------------------------------------
+def get_profile_env_var(varname, profile_file):
+    shell_command = [
+        'bash', '-c', 'source {0}; echo ${1}'.format(
+            profile_file, varname
+        )
+    ]
+    variable_content, error = exec_subprocess(shell_command, True)
+    if not error:
+        return variable_content.decode().strip()
+
+
+# ----------------------------------------------------------------------------
+def is_registry_registered(registry_server_name):
+    """
+    Check if the instance is registered for accessing the
+    container registry server on the update server
+    """
+    # A good registry setup exists if we can find the entry in the
+    # hosts file and a proper environment definition
+    hosts = get_hosts()
+    registry_auth_file = get_profile_env_var(
+        'REGISTRY_AUTH_FILE', PROFILE_LOCAL_PATH
+    )
+    if registry_server_name in hosts.keys() \
+       and registry_auth_file == REGISTRY_CREDENTIALS_PATH \
+       and os.path.exists(registry_auth_file):
+        return True
+    return False
 
 
 # ----------------------------------------------------------------------------
--- a/tests/data/hosts
+++ b/tests/data/hosts
@@ -1,0 +1,25 @@
+#
+# hosts         This file describes a number of hostname-to-address
+#               mappings for the TCP/IP subsystem.  It is mostly
+#               used at boot time, when no name servers are running.
+#               On small systems, this file can be used instead of a
+#               "named" name server.
+# Syntax:
+#    
+# IP-Address  Full-Qualified-Hostname  Short-Hostname
+#
+
+127.0.0.1   localhost
+
+# special IPv6 addresses
+::1             localhost ipv6-localhost ipv6-loopback
+
+fe00::0         ipv6-localnet
+
+ff00::0         ipv6-mcastprefix
+ff02::1         ipv6-allnodes
+ff02::2         ipv6-allrouters
+ff02::3         ipv6-allhosts
+
+# other hosts
+192.0.2.1   some.box
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -449,7 +449,7 @@ if registration_smt:
                 new_target
             )
             utils.update_rmt_cert(new_target)
-            break
+            sys.exit(0)
         else:
             msg = 'Configured update server is unresponsive. Could not find '
             msg += 'a replacement update server in this region. '
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -46,11 +46,14 @@ from cloudregister import smt
 from lxml import etree
 from requests.auth import HTTPBasicAuth
 
+LOG_FILE = '/var/log/cloudregister'
+
 # Disable the urllib warnings
 # We have server certs that have no subject alt names
 # We have to check the server state API without certificate validation
 urllib3.disable_warnings()
 registration_returncode = 0
+
 
 # ----------------------------------------------------------------------------
 def get_register_cmd():
@@ -76,7 +79,7 @@ def get_register_cmd():
             # transactional-update command
             if 'ro' in fsoptions.split(','):
                 cmd_name = 'transactional-update'
-                for path in ['/sbin/','/usr/sbin/']:
+                for path in ['/sbin/', '/usr/sbin/']:
                     exec_path = path + cmd_name
                     if os.path.exists(exec_path):
                         register_cmd = exec_path
@@ -90,7 +93,8 @@ def get_register_cmd():
                     sys.exit(1)
 
     return register_cmd
-    
+
+
 # ----------------------------------------------------------------------------
 def register_modules(extensions, products, registered=[], failed=[]):
     """Register modules obeying dependencies"""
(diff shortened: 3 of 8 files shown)


schaefi:
Please note I saw that the hosts file is being read at several places and I believe the tool deserves one method that provides a proper representation. That's why I wrote one also with the idea that other code parts could benefit from it later.

usr/sbin/registercloudguest
```diff
@@ -170,6 +173,23 @@ def cleanup():
...
+# ----------------------------------------------------------------------------
+def register_registry(
+    registration_target: smt.SMT, user: str = '', password: str = ''
+) -> None:
+    """
+    Run the registration process for the container registry
+    """
+    registry_fqdn = registration_target.get_registry_FQDN()
+    if not utils.is_registry_registered(registry_fqdn):
+        if not utils.setup_registry(registry_fqdn, user, password):
```
jesusbv, line 185: can those two `ifs` be one `if` with two conditions ?

usr/sbin/registercloudguest
```diff
@@ -649,12 +669,10 @@ if registration_returncode:
...
 
-if not utils.setup_registry(
-    registration_target.get_registry_FQDN(), user, password
-):
-    cleanup()
-    sys.exit('Registration failed, see /var/log/cloudregister for details')
+# Run the container registry registration process
+register_registry(registration_target, user, password)
 
+# Jobs done
```
jesusbv, line 675: s/Jobs/Repository and registry registrations/

jesusbv:
Other than minor nitpicks, LGTM

lib/cloudregister/registerutils.py
```diff
@@ -1621,6 +1654,20 @@ def import_smt_cert(smt):
...
+# ----------------------------------------------------------------------------
+def is_registry_registered(registry_server_name: str) -> bool:
+    """
+    Check if the instance is registered for accessing the
+    container registry server on the update server
+    """
+    # The indicator for the registry access is provided
+    # by the entry in the hosts file.
+    hosts = get_hosts()
+    if registry_server_name in hosts.keys():
```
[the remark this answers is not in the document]
schaefi: The credentials file will exist if the user has run any other sort of `podman login` on this system. I don't think it's a valuable check. I agree to the env check and will add that part

usr/sbin/registercloudguest
```diff
@@ -170,6 +173,23 @@ def cleanup():
     utils.clear_new_registration_flag()
     utils.clean_framework_identifier()
 
+# ----------------------------------------------------------------------------
+def register_registry(
```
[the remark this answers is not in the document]
schaefi: ok :)

lib/cloudregister/registerutils.py
```diff
@@ -1103,6 +1106,36 @@ def get_credentials_file(update_server, service_name=None):
     return credentials_file
 
 
+# ----------------------------------------------------------------------------
+def get_hosts() -> Dict[str, Dict[str, Union[str, bool]]]:
```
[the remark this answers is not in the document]
schaefi: Uhm, that's really a major drawback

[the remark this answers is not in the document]
schaefi: Also fstrings are not supported in py3.4, that also needs to be reverted :(((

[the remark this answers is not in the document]
schaefi: done

usr/sbin/registercloudguest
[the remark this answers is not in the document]
schaefi: done

lib/cloudregister/registerutils.py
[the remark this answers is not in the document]
schaefi: Added check for the environment

usr/sbin/registercloudguest
```diff
@@ -649,12 +669,10 @@ if registration_returncode:
     )
     sys.exit(registration_returncode)
 
-if not utils.setup_registry(
-    registration_target.get_registry_FQDN(), user, password
-):
-    cleanup()
-    sys.exit('Registration failed, see /var/log/cloudregister for details')
+# Run the container registry registration process
+register_registry(registration_target, user, password)
```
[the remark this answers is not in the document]
schaefi: To summarize:

```
The behavior we want is that if the system is connected to the update infrastructure but the registries are missing then we want them set up. Due to the check on Line 409 we will not reach this code if the system is already connected.
```

yes fully agreed and this change will come next I promise, but as a followup and in line with the dev-plan.
Thanks

schaefi:
Thanks Jesus for the review. @rjschwei I think I covered the topics you brought up. Could you do another review ? If we are good the next PRs will focus on the mentioned behavior changes. Thanks

PS: I tried to add a workflow run for python 3.4 but the github matrix has no such old disitribution available anymore. So if we want 3.4 testing we would need to add our own hosted sle12 container or something. Would be nice to be honest because I think nobody really tests the code on py3.4 but I have not enough time to set that up, maybe we have someone in the team who has experience with private github runners

usr/sbin/registercloudguest
```diff
@@ -170,6 +173,21 @@ def cleanup():
...
     utils.clean_framework_identifier()
 
+# ----------------------------------------------------------------------------
+def setup_registry(registration_target, user, password):
+    """
+    Run the registration process for the container registry
+    """
+    registry_fqdn = registration_target.get_registry_FQDN()
+    if not utils.is_registry_registered(registry_fqdn)
+       and not utils.setup_registry(registry_fqdn, user, password):
```
rjschwei, line 183: I simplified this in #187, also pyflakes 3.6 complained about the current code being invalid syntax.

usr/sbin/registercloudguest
```diff
@@ -170,6 +173,21 @@ def cleanup():
...
+        cleanup()
+        print(
+            'Registration failed(registry), see {0} for details'.format(
+            ), file=sys.stderr
```
rjschwei, line 187: Syntax error, the substitution is missing, also fixed in #187

rjschwei (changes requested):
I know you have a refactor branch for the early exit condition as such I didn't pay much attention to avoid duplication in #187 . That said I think it makes sense to have the addition of the registry in this PR in the early exit path to make this functionally complete and then use the ltss_refactor_early_exit branch to do the cleanup/streamlining.

In any event the logic triggering the syntax error in Python 3.6 and the missing format substitution have to be fixed even if #187 does not get merged.

usr/sbin/registercloudguest
schaefi: flake8 is not even called for this code... my bad this should not have landed in any PR. Thanks

schaefi: Thanks I merged your PR and added the code path to flake8 which complained at several places and I fixed those in another commit

tests/data/hosts
```diff
@@ -0,0 +1,25 @@
...
+
+fe00::0         ipv6-localnet
+
+ff00::0         ipv6-mcastprefix
+ff02::1         ipv6-allnodes
+ff02::2         ipv6-allrouters
+ff02::3         ipv6-allhosts
+
+# other hosts
+192.168.178.1   some.box
```
pdostal, line 25:
```suggestion
192.0.2.1   some.box
```
Let's use a bit more neutral address ([rfc5737](https://datatracker.ietf.org/doc/rfc5737/))

schaefi: ok, makes sense. Thanks I will change that

schaefi: done

schaefi:
I'm going to zip a package with this PR and test the initial registration as well as the processing if the system is already registered. Stay tuned :)

schaefi:
For your reference. There is a package including this patch available here:

In addition there is a container that uses this package as part of my infrastructure development env here:

* https://build.suse.de/package/show/home:mschaefer/RMTClient

If you would like to do your own testing you can setup a local infrastructure, container based and only for the general use cases (not cloud specific), using the following documentation:

* https://confluence.suse.com/display/publiccloud/Setup+Public+Cloud+Development+Environment

I did use this env for testing and here is what I did:

# Initial registration

- [ ] Run the RMTClient container, login, expect the registration to be done
- [ ] Checked the journal of ```systemctl status guestregister```
- [ ] Checked access via ```zypper refresh```

# subsequent registration

- [ ] Store the current ```/var/log/cloudregister``` and recall registercloudguest
- [ ] Compare the two logs
- [ ] Checked access via ```zypper refresh```

Result of the log diff

```
diff -u /var/log/cloudregister cloudregister 
+++ /var/log/cloudregister      2024-07-29 07:43:43.774677913 +0000
--- cloudregister       2024-07-29 07:43:27.014675166 +0000
@@ -62,12 +62,3 @@
 2024-07-29 07:37:44,813 DEBUG:Starting new HTTPS connection (1): localhost:44300
 2024-07-29 07:37:44,830 DEBUG:https://localhost:44300 "GET /api/health/status HTTP/1.1" 200 None
 2024-07-29 07:37:44,831 INFO:Current update server will be used: "('localhost:44300', None)"
+2024-07-29 07:43:43,679 INFO:Using API: regionInfo
+2024-07-29 07:43:43,692 INFO:Getting update server information, attempt 1
+2024-07-29 07:43:43,692 INFO:  Using region server: localhost:44301
+2024-07-29 07:43:43,719 INFO:Check for cert update
+2024-07-29 07:43:43,720 INFO:Using API: regionInfo
+2024-07-29 07:43:43,747 INFO:No cert change
+2024-07-29 07:43:43,767 INFO:Instance is registered, and update server is reachable, nothing to do
+2024-07-29 07:43:43,774 INFO:No credentials entry for "*localhost:44300"
+2024-07-29 07:43:43,777 INFO:Instance is setup to access container registry, nothing to do
```

schaefi:
So far this looks good to me

@rjschwei Can you take a look to the code, the tests and maybe run your own or let me know what I forgot. If we are fine, your approval would be great such that we can continue with the next steps that are closer to end goal, LTSS support

Thanks much

schaefi:
I also did several tests with breaking the system by e.g deleting /etc/profile.local, unset profile env, delete some repo from /etc/zypp/repos.d ... So there are plenty of ways to destroy the connection to the update infrastructure. A simple ```registercloudguest``` does not fix this, however the cleanup implementation seems to do a good job 

```
registercloudguest --clean
registercloudguest
```

always worked.

As we have decided to print a message regarding the update of the environment. That message is currently part of the logfile.

```
2024-07-29 08:27:27,694 INFO:Adding registry setup to instance, all existing shell sessions must be restarted to use the registry.
```

I think we want that information on the console.

That one also brought me to the question why not all logging.info messages are printed on the console. When started via systemd all this info will be placed to the journal and the logfile. I opened #188 to address this in the future

As a quick fix for the issue described above I will use the standard ```print``` method as it is done in several other places of the code

rjschwei (approved):

PR #184, opened 2024-07-25
Fix spec file

schaefi:
There is code that imports the yaml module but no
package requirement is setup in the spec file


jesusbv (approved):
Odd how the testing package does not complain about it when getting installed

PR #186, opened 2024-07-26
Update early exit condition and registry setup

schaefi:
Update early exit condition and registry setup
    
* Reduce code duplication and prevent duplicate checking for  utils.is_registry_registered, increase readability and prepare better for the LTSS use case additions in the future
        
* Refactor code to identify the responding update servers into its own function for better readibility.
    
* Move the setup of the container registry in front of the setup of the repositories(SUSEConnect), as it could now be done directly after checking of the early exit condition

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -178,16 +178,23 @@ def cleanup():
 
 
 # ----------------------------------------------------------------------------
-def setup_registry(registration_target, user, password, clean='all'):
+def setup_registry(registration_target, clean='registry'):
     """
     Run the registration process for the container registry
 
     clean == all -> cleans repository and registry setup
     clean == registry -> clean only registry artifacts
     """
+    user, password = utils.get_credentials(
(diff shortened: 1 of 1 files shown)


schaefi:
Important: This change can only be merged if #183 got merged

schaefi:
Important: This change requires integration testing on test servers or in my provided development container setup. This changes the behavior as described in the initial comment

usr/sbin/registercloudguest
```diff
@@ -178,16 +178,23 @@ def cleanup():
...
     clean == all -> cleans repository and registry setup
     clean == registry -> clean only registry artifacts
     """
+    user, password = utils.get_credentials(
+        utils.get_credentials_file(registration_target)
+    )
     registry_fqdn = registration_target.get_registry_FQDN()
     if utils.is_registry_registered(registry_fqdn):
         return True  # already setup nothing to do
+    logging.info(
```
jesusbv, line 194: I would move the log message to _after_ the action

PR #187, opened 2024-07-26
Add registry to already registered systems

rjschwei:
For systems that are already registered we do not want to force the user to re-register the system to get access to the registries. When we detect, on boot or manual execution that the system already points to the update infrastructure add the registry setup to the system.


schaefi (approved):
makes sense and thanks for fixing the syntax issues. We have to add this code to the flake8 checks

PR #189, opened 2024-07-29
Sle12 ports

schaefi:
For SLE12 some patches to the registration client are needed. I moved the existing patch to the git and also added a new one which we need because we don't support the registry registration on SLE12

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -44,7 +44,9 @@ Requires:       python3-lxml
 Requires:       python3-requests
 Requires:       python3-urllib3
 Requires:       python3-zypp-plugin
+%if 0%{?suse_version} > 1315
 Requires:       python3-toml
+%endif
 Requires:       regionsrv-certs
 Requires:       sudo
 Requires:       zypper
@@ -63,7 +65,9 @@ BuildRequires:  python3-lxml
 BuildRequires:  python3-requests
 BuildRequires:  python3-setuptools
 BuildRequires:  python3-zypp-plugin
+%if 0%{?suse_version} > 1315
 BuildRequires:  python3-toml
+%endif
 BuildRequires:  sudo
 BuildRequires:  systemd-rpm-macros
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
@@ -132,8 +136,10 @@ Enable/Disable Guest Registration for Microsoft Azure
 %prep
 %setup -q
 %if 0%{?suse_version} == 1315
-%patch0
-%patch1
+%patch -P 0
+%patch -P 1
+test -e usr/sbin/registercloudguest.orig && rm usr/sbin/registercloudguest.orig
+test -e lib/cloudregister/registerutils.py.orig && rm lib/cloudregister/registerutils.py.orig
 %endif
 
 %build
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -27,6 +27,8 @@ URL:            http://www.github.com/SUSE-Enceladus/cloud-regionsrv-client
 Source0:        %{name}-%{version}.tar.bz2
 # PATCH-FIX-SLES12 bsc#1203382 fix-for-sles12-disable-ipv6.patch
 Patch0:         fix-for-sles12-disable-ipv6.patch
+# PATCH-FIX-SLES12 fix-for-sles12-disable-registry.patch
+Patch1:         fix-for-sles12-disable-registry.patch
 Requires:       SUSEConnect > 0.3.31
 Requires:       ca-certificates
 Requires:       cloud-regionsrv-client-config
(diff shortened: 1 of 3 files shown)


cloud-regionsrv-client.spec
```diff
@@ -131,6 +133,7 @@ Enable/Disable Guest Registration for Microsoft Azure
 %setup -q
 %if 0%{?suse_version} == 1315
 %patch0
+%patch1
```
rjschwei, line 136: The `patch#` syntax is deprecated, please switch to `%patch -P 0` and `%patch -P 1`

fix-for-sles12-disable-registry.patch
```diff
@@ -0,0 +1,10 @@
+--- usr/sbin/registercloudguest.orig
++++ usr/sbin/registercloudguest
+@@ -185,6 +185,7 @@ def setup_registry(registration_target, clean='registry'):
+     clean == all -> cleans repository and registry setup
+     clean == registry -> clean only registry artifacts
+     """
++    return None
+     user, password = utils.get_credentials(
+         utils.get_credentials_file(registration_target)
+     )
```
rjschwei, line 10: This patch should also contain
```
def is_registry_registered(registry_server_name):
    return True
```

https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/183/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R1671

rjschwei: And we need to patch out the hosts entry in SLE 12, i.e.
https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L75 needs to be skipped

rjschwei (changes requested):

fix-for-sles12-disable-registry.patch
schaefi: Hmm, are you sure. If we disable everything related to the registry setup (which is what the patch does) we will never reach the mentioned code paths

cloud-regionsrv-client.spec
schaefi: ok

schaefi: done

schaefi:
For reference, package build for SLE12-SP5 can be found here:

schaefi:
We have more issues with SLE12. As you can see from the above build there is no python3-toml in SLE12 but recently an `import toml` was added to the python code I think due to the reading of that suse manager files

schaefi:
Missing python-toml for SLE12 added here:

* https://build.suse.de/package/show/home:mschaefer/python-toml

This is a link from the SLE15 version.

schaefi:
SLE12 package builds with patching results 

```
[   29s] + cd cloud-regionsrv-client-10.3.0
[   29s] + /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
[   29s] + echo 'Patch #0 (fix-for-sles12-disable-ipv6.patch):'
[   29s] Patch #0 (fix-for-sles12-disable-ipv6.patch):
[   29s] + /usr/bin/cat /home/abuild/rpmbuild/SOURCES/fix-for-sles12-disable-ipv6.patch
[   29s] + /usr/bin/patch -p0 --fuzz=0
[   29s] patching file lib/cloudregister/smt.py
[   29s] Hunk #1 succeeded at 128 (offset 19 lines).
[   29s] + exit 0

```

Testing on a SLE12 registered system would be nice... I don't have one

fix-for-sles12-disable-registry.patch
rjschwei: @schaefi yes. The `add_hosts_entry` function is called to make `smt-$SCP.susecloud.net` resolvable and with the addition of the registry it now also makes the registry resolvable. But on SLE 12 we do not support containers so we do not want a registry entry in the `/etc/hosts/file. The SMT object is created with information from the region server and the region server will send the registry information since it does not differentiate between SLE12 or SLE15 systems and as such `if smt_server.get_registry_FQDN()` will also be true on SLE 12 clients.

schaefi: Hrm, yes I see. There is this code 

```
if smt_server.get_registry_FQDN():
        entry += '%s\t%s\n' % (
            smt_ip,
            smt_server.get_registry_FQDN()
        )
```

in `add_hosts_entry()` and it is called unconditionally, thus also for SLE12

ok yes this needs patching or a proper refactoring. imho it would be nice to have a generic add_hosts_entry method and two specific methods that says `add_rmt_hosts_entry` and `add_registry_hosts_entry` 

But I will not do this refactoring now and just adapt the SLE12 patch... we don't have time for this now

schaefi: done

schaefi:
> We have more issues with SLE12. As you can see from the above build there is no python3-toml in SLE12 but recently an `import toml` was added to the python code I think due to the reading of that suse manager files

I solved this by not importing toml, it was only needed for registry code which got disabled by the patch

rjschwei (approved):

PR #190, opened 2024-07-29
Deduplicate code that calls SUSEConnect

schaefi:
Instead of duplicating the code to prepare for a SUSEConnect (or the respective transactional) command, provide a common method named run_SUSEConnect which only receives the relevant parameters. Apply the new method to the main registration call as well as to the module registration. This refactor was done also in preparation to an additional LTSS regcode process which requires another SUSEConnect call.

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -152,7 +152,12 @@ def run_SUSEConnect(
     if regcode:
         cmd += ['--regcode', regcode]
 
-    logging.info('Registration: {0}'.format(' '.join(cmd)))
+    log_information = ' '.join(cmd)
+    if regcode:
+        # registration codes should not end up in the log
+        log_information.replace(regcode, 'XXXX')
+
+    logging.info('Registration: {0}'.format(log_information))
     call = subprocess.Popen(
         cmd,
         stdout=subprocess.PIPE,
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -42,6 +42,7 @@ import uuid
 
 import cloudregister.registerutils as utils
 
+from collections import namedtuple
 from cloudregister import smt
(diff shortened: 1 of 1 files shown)


jesusbv (approved):
LGTM

usr/sbin/registercloudguest
```diff
@@ -95,11 +96,80 @@ def get_register_cmd():
...
+    # set registration code for given product
+    if regcode:
+        cmd += ['--regcode', regcode]
+
+    logging.info('Registration: {0}'.format(' '.join(cmd)))
+    call = subprocess.Popen(
+        cmd,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
```
rjschwei, line 160: @Johannes should we, do we need to sanitize the log, i.e. do not log the registration code?

rjschwei (approved):
Let's wait to hear back from Johannes about logging the reg code before we merge. Other than that LGTM

usr/sbin/registercloudguest
johannes: Funny, I am working on Python code right now using Popen and doing something not too different, but still you highlighted the wrong person :)

Good luck with your project!

rjschwei: Funny, sorry @johannes couldn't find the Johannes I was looking for and was only going for the name without realizing there was a user reference just for Johannes.

rjschwei: Per Slack conversation with another Johannes, we should not log the reg code. @schaefi please update accordingly. Thanks

schaefi: All right, I will adapt the PR

schaefi: done

rjschwei (approved):

PR #192, opened 2024-08-01
Add SUSEConnect command for testing

jesusbv:
Those commands use the same method that registry engine uses on RMT to verify cache access, so they must work as expected


jesusbv:
For reference, please take a look at [bsc#1228774](https://bugzilla.suse.com/show_bug.cgi?id=1228774)

rjschwei (approved):

PR #194, opened 2024-08-12
Add option -p/--product for regcode products

schaefi:
Allow to specify the product triplet on the commandline. This option can be used to register an add-on product which is not of the same base product.

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -318,7 +318,7 @@ def get_product_triplet(product_tree):
 
 # ----------------------------------------------------------------------------
 def setup_ltss_registration(
-    registration_target, regcode, instance_data_filepath
+    registration_target, regcode, instance_data_filepath, product_name=''
 ):
     """
     Run SUSEConnect to register LTSS for this instance
@@ -345,12 +345,16 @@ def setup_ltss_registration(
     else:
         base_product = get_product_triplet(product)
 
-        # The name of the product to register is the LTSS variant for the
-        # registered base product. In theory other combinations are possible
-        # but we don't allow this for the moment
-        ltss_product_triplet = '{0}-LTSS/{1}/{2}'.format(
-            base_product.name, base_product.version, base_product.arch
-        )
+        if not product_name:
+            # The name of the product to register, if not specified differently
+            # is the LTSS variant for the registered base product.
+            ltss_product_triplet = '{0}-LTSS/{1}/{2}'.format(
+                base_product.name, base_product.version, base_product.arch
+            )
+        else:
+            # The product name was specified per cmdline option.
+            # We leave it in the hand of the server side to validate it
+            ltss_product_triplet = product_name
 
         suseconnect = run_SUSEConnect(
             registration_target=registration_target,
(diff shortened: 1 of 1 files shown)


schaefi:
Note: This is outside of the dev plan and I know we have agree to not allow product registrations that are not based on the base product. With this option you can specify any product name and I only excluded it from being used in the base product registration.

Why am I proposing this new option:

When testing the LTSS registrations I got a regcode from Jesus for `SLES-LTSS/15.4/x86_64` There is no LTSS regcode for 15.5 so far and therefore the only option to test the LTSS process was in combining a SLES15-SP5 registered instance with a SLES-LTSS-SP4 support registration. I was told this is a valid setup and such it seems to be a customer facing situation that can appear with any LTSS offering that SUSE makes.

schaefi:
This needs clarification if a product combination like

`SLES-15-SP5` + `SLES-LTSS-15-SP4` is allowed

usr/sbin/registercloudguest
```diff
@@ -341,12 +341,16 @@ def setup_ltss_registration(registration_target, regcode):
     else:
         base_product = get_product_triplet(product)
```
jesusbv, line 342: This should be moved to line 347 of this PR, inside the no product name block, if we are going to have a default name

IMO, we should have both regcode + product_name from the cmdline options, instead of a default product_name, the user knows which product they got the regcode for

if extensions to be activated with a regcode will have to match the base product version/arch maybe we dont need the triplet ? will we support different versions than the base product for the extensions ?

jesusbv (changes requested):
getting the base product triplet method call should move to inside the if block, see comment

usr/sbin/registercloudguest
schaefi: > This should be moved...
I rebased the change here with master and that move has already been done

> IMO, we should have both regcode + product_name from the cmdline options, instead of a default product_name, the user knows which product they got the regcode for
Well this is not the scope of this PR. With this PR we give the user an opportunity to specify the product on the cmdline which did not exist before. Using the base product name extended by 'LTSS' as the default product name if none was specified was a decision we wrote into the implementation details document and is not touched nor questioned in this PR. I believe the default is good and having an opportunity to specify it on the commandline is still a question I'd like to discuss.

So is there a real use case for this option in our scope ?

After all the LTSS testing I think there is no use case for allowing alternative product names and I believe LTSS should follow the base product only.

schaefi:
@rjschwei @jesusbv I tend to close this one without merge. See my comments in the conversation thread Jesus opened

usr/sbin/registercloudguest
jesusbv: > So is there a real use case for this option in our scope ?

I think there is, right now we have LTSS but the moment we have one more non free extension available to be activated, we would need to specify the product somehow, having `-p` in the cli is the right option, as it is the pattern we have been using

jesusbv: True, you assumed that, I did assume if more products are added to be activated with `registercloudguest -r regcode` same way as LTSS, I do not know if that would be a use case in the future (I assumed other products may or may not follow) and if so, how they would be activated

schaefi: Yes this is also a fair assumption. However, I suggest to wait adding this functionality when we really have this use case. I'm going to close this PR but keep the branch for later.

I believe it's better not to provide an option which could be easily used to try a registration against a bogus product name and I also believe customers would require documentation to pass along the correct value. For the later a real use case is needed so let's stick to what we have right now.

Thanks

schaefi:
closed as `keep branch for later` when there is a real use case

rjschwei:
There is a related jira ticket PCT-475. It was decided that we will not move forward with supporting other add-ons in the same way we support LTSS. This is consistent with the closure of this request.

PR #196, opened 2024-08-14
Fixed setup of registry hosts entry

### LINKED ISSUE #195: Adding registry infor upon re-run of registercloudguest misses hosts update
# registercloudguest 
Instance registry setup done, sessions must be restarted !
# grep registry /etc/hosts

there should be a registry-ec2.susecloud.net entry in the hosts file in this case as the test was run on EC2.

Package was taken from the test build in the SUSE build service: cloud-regionsrv-client-10.3.0-150500.12.1

The region server does return the registry information as expected:

>>> r = requests.get('https://52.79.82.165/regionInfo?regionHint=us-east-1', verify='/usr/lib/regionService/certs/52.79.82.165.pem')
>>> r.text
'<regionSMTdata>\n<smtInfo SMTserverIP="54.197.240.216" SMTserverName="smt-ec2.susecloud.net" SMTregistryName="registry-ec2.susecloud.net" fingerprint="A0:8A:3E:90:1A:FE:DF:A4:DB:A4:B0:61:A4:86:DE:48:8A:B9:D8:75" region="us-east-1"/>\n<smtInfo SMTserverIP="54.225.105.144" SMTserverName="smt-ec2.susecloud.net" SMTregistryName="registry-ec2.susecloud.net" fingerprint="A0:8A:3E:90:1A:FE:DF:A4:DB:A4:B0:61:A4:86:DE:48:8A:B9:D8:75" region="us-east-1"/>\n<smtInfo SMTserverIP="107.22.231.220" SMTserverName="smt-ec2.susecloud.net" SMTregistryName="registry-ec2.susecloud.net" fingerprint="A0:8A:3E:90:1A:FE:DF:A4:DB:A4:B0:61:A4:86:DE:48:8A:B9:D8:75" region="us-east-1"/>\n</regionSMTdata>'

The environment setting is applied.

# cat /etc/profile.local

export REGISTRY_AUTH_FILE=/etc/containers/config.json

export DOCKER_CONFIG=/etc/containers

schaefi:
There is a logic error in the code that sets up the hosts entry. One method named add_hosts_entry() is responsible for setting up the RMT and the registry hosts entries. Both entries targets the same server IP and can only be differentiated by their name and not by their IP. Next to this design there is a method named has_rmt_in_hosts() which checks by the IP of the server if the hosts entry exists or not. Obviously this condition check is not sufficient if there is the need to check individually for the hosts entry of the RMT target or the registry target.

At the time we added support for the container registry target, we require a more specific host entry checking which is specific to the target system we want to add the hosts entry for. This is because there are instances which can run an old version of cloud-regionsrv-client that only has support for the RMT target and therefore has added a proper hosts entry for this target. If a customer now updates cloud-regionsrv-client, the next reboot or manual call for registercloudguest adds support for the registry but did not update the hosts file because due to the checking above the code thought everything is ok because the IP based check for the hosts entry said everything is good.

To fix this I changed the way how has_rmt_in_hosts() works by using the new get_hosts() interface as well as checking for the FQDN for RMT and not the IP. In addition I added a new method called has_registry_in_hosts() which checks for the FQDN for the registry. has_registry_in_hosts() is then used to make sure a hosts entry for the registry exists independently of the case that there is a RMT hosts entry already present. This Fixes #195

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1104,7 +1104,7 @@ def get_credentials_file(update_server, service_name=None):
 
 
 # ----------------------------------------------------------------------------
-def get_hosts():
+def get_hosts(with_loopback=False):
     """
     Read /etc/hosts file into the following structure
 
@@ -1124,6 +1124,12 @@ def get_hosts():
     for line in hostlines:
         hostnames = line.split('#')[0].split()[1:]
         hostaddress = line.split('#')[0].split()[0]
+        loopback = False
+        if hostaddress == '::1' or hostaddress == '127.0.0.1':
+            loopback = True
+        if not with_loopback and loopback:
+            # no loopback addresses wanted
+            continue
         for hostname in hostnames:
             hosts[hostname] = {
                 'address': hostaddress,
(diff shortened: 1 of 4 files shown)


schaefi:
I have tested this code in my container based development environment. While doing the testing I also found that we might not want get_hosts() to include the loopback addresses for ipv4/6 because the names associated with the loopback interface could interfere with other names for real remote locations, hence the small change on the get_hosts interface

schaefi:
@rjschwei for testing I created a package that includes this PR at the usual place:

rjschwei (approved):

PR #197, opened 2024-08-15
Fix docker config file mirror settings

rjschwei:
Per docker documentation setting a registry mirror requires a url. Previously we wrote the fqdn of the update server to the daemon.json which breaks docker daemon start up with the 'unsupported scheme "" ' message. Convert the fqdn into a url.


schaefi:
Oh very good catch :+1: Thanks

schaefi (approved):

PR #199, opened 2024-08-23
Handle process structure change

rjschwei:
During migration the various process get spawned. When looking for the main zypper executable (Zypp-main) we also get information about the child process. However we are only interested in the main process PID. Change the logic to extract the PID such the we only look at the PID for the main process.

### DIFF
--- a/integration_test-process.txt
+++ b/integration_test-process.txt
@@ -125,6 +125,26 @@ After installing the test package with "zypper in"
   + Cloud based RMT server is the target
 - registercloudguest --clean
 
+Migration testing
+SLE 15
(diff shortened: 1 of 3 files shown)


jesusbv (approved):
LGTM

shall we update that [method test](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/tests/test_registerutils.py#L1974) ?

rjschwei:
> LGTM
> 
> shall we update that [method test](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/tests/test_registerutils.py#L1974) ?

Well, a modification would not be sufficient, added a new test.

PR #200, opened 2024-08-26
Send instance data when activating LTSS

jesusbv:
- Move the block to get the instance data written in a file to before LTSS block

- Update the methods to incude instance data file path

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -318,7 +318,9 @@ def get_product_triplet(product_tree):
 
 
 # ----------------------------------------------------------------------------
-def setup_ltss_registration(registration_target, regcode):
+def setup_ltss_registration(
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -351,7 +353,9 @@ def setup_ltss_registration(registration_target, regcode):
         suseconnect = run_SUSEConnect(
             registration_target=registration_target,
             product=ltss_product_triplet,
-            regcode=regcode
+            regcode=regcode,
+            instance_data_filepath=instance_data_filepath
+            if os.path.exists(instance_data_filepath) else ''
```
rjschwei, line 358: This already happens in the called function, no need to to the same operation twice. Fixed, force pushed.

rjschwei (approved):

PR #201, opened 2024-08-26
Do not lower the product name text

jesusbv:
use SLES-LTSS instead of sles-LTSS,for example

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -311,7 +311,7 @@ def get_product_triplet(product_tree):
         'product_type', ['name', 'version', 'arch']
     )
     return product_type(
-        name=product_tree.find('name').text.lower(),
+        name=product_tree.find('name').text,
         version=product_tree.find('version').text,
         arch=product_tree.find('arch').text
     )


usr/sbin/registercloudguest
```diff
@@ -311,7 +311,7 @@ def get_product_triplet(product_tree):
         'product_type', ['name', 'version', 'arch']
     )
     return product_type(
-        name=product_tree.find('name').text.lower(),
+        name=product_tree.find('name').text,
```
schaefi, line 314: I'd like to get @rjschwei input here. I added the method `get_product_triplet` in one of the former pull requests and the code to lookup `name`, `version` and `arch` was just copied from the existing implementation. I'm sure Robert had a reason why the name attribute was lowered.

As we are calling the method for the base product registration and for any later other product registration we should make sure no longer lowering the value does not harm the base product triplet

@rjschwei can you please double check on that ?

Thanks

rjschwei: Well this code has existed for a long time and the `.lower()` was most likely inserted to address a case mismatch between what was delivered in a .prod file and what was expected server side.  As such there are 2 questions on the subject:

1.) @digitaltom do you remember when/where we had a mismatch between product name as it was encoded in the `.prod` file and what was expected in SCC or on SMT or RMT when the product triplet was passed? @mcalmer maybe something in SMT?
2.) @digitaltom for the base product, i.e. SLES both work

`SUSEConnect -p sles/15.6/x86_64 -r foo`

 and

 `SUSEConnect -p SLES/15.6/x86_64 -r foo`

 why is this not true for LTSS products?

`SUSEConnect -p SLES-LTSS/15.3/x86_64 -r bar`

works but

`SUSEConnect -p sles-ltss/15.3/x86_64 -r bar`

fails.

rjschwei: > I don't know if it should be lower 

Then why do you post a PR requesting a merge of the change when you do not know and by extension do not know the effects of the PR?

digitaltom: 1) I can't remember such a mismatch. 

2) Please use the string as it is returned by `zypper products` (.prod files from /etc/products.d). You can see the matching string in SCC by browsing to the product (eg. https://scc.suse.com/admin/products/2696) -> 'Show technical details'. In case of 15SP4 LTSS, that indeed is "SLES-LTSS" and the `lower()` looks like it will break. 

This is not a limitation of SCC, because we use the string case insensitive (see https://github.com/SUSE/happy-customer/blob/separate-rancher-prime/glue/app/models/product.rb#L256), but an issue when installing the release package (as quoted above), which needs the exact identifier string. For SLES, i suppose the lowercase approach works, because the release package is already installed.

rjschwei (approved):

usr/sbin/registercloudguest
mcalmer: Originally this tripple was case insensitiv (NCC times), thats why SMT  and SUMA store these values lowercase.
Is it this what you mean?

jesusbv: > Then why do you post a PR requesting a merge of the change when you do not know and by extension do not know the effects of the PR?

As mentioned in the same comment, the tests show that the `lower` did not work and removing it did work
Now that there is a PR, there was an opportunity to discuss that

In any case, if the current tests done were not enough, we are still on the test phase, so if there any other side effect, we would see it

PR #203, opened 2024-08-26
Make network access check message less alarming

rjschwei:
At present when we test for network access by IP protocol version the message in the log file makes it hard to differentiate whether this is hard error or for information. While the message is set as info it still contains the word "error" in the message. Make the message less alarming in an attempt to not train users to ignore errors.

Enhance the testing of the function checking the network access. Previously only the happy path was tested.

Make the network acces test function private to the utils, we do not want function to be used outside the utils. Client code should use the test methods for IPv4 and IPv6 access.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -246,7 +246,7 @@ def fetch_smt_data(cfg, proxies, quiet=False):
         except Exception as e:
             logging.error('=' * 20)
             logging.error(str(e))
-            logging.error('Unable to obtain SMT server information, exiting')
+            logging.error('Unable to obtain update server information, exiting')
             sys.exit(1)
         smt_info = json.loads(response.text)
         expected_entries = ('fingerprint', 'SMTserverIP', 'SMTserverName')
@@ -1485,6 +1485,25 @@ def get_zypper_target_root():
 
 
 # ----------------------------------------------------------------------------
+def has_network_access_by_ip_address(server_ip):
+    """Check if we can connect to the given server"""
+    try:
+        connection = socket.create_connection((server_ip, 443), timeout=2)
+    except OSError:
+        try:
+            ip_ver = 'IPv%d' % ipaddress.ip_address(server_ip).version
+        except ValueError:
+            ip_ver = 'UNKNOWN'
+        logging.info(
+            'Skipping %s protocol version, no network configuration' % ip_ver
+            )
+        return False
+
+    connection.close()
+    return True
+
+
+# ----------------------------------------------------------------------------
 def has_rmt_ipv6_access(smt):
     """IPv6 access is possible if we have an SMT server that has an IPv6
        address and it can be accessed over IPv6"""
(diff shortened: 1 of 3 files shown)


jesusbv (changes requested):
The renamed method is used [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L472)

Keep the old name or rename the method in all the places it is used

rjschwei:
> The renamed method is used [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L472)
> 
> Keep the old name or rename the method in all the places it is used

Thanks for catching this. fixed

usr/sbin/registercloudguest
```diff
@@ -454,8 +454,11 @@ if args.user_smt_ip or args.user_smt_fqdn or args.user_smt_fp:
         print(msg, file=sys.stderr)
         sys.exit(1)
 
+    ip_addr_version = 'IPv4'
     try:
-        ipaddress.ip_address(args.user_smt_ip)
+        ip_addr = ipaddress.ip_address(args.user_smt_ip)
```
jesusbv, line 459: Instead of introducing two new variables here:
-  `ip_addr_version`
- `ip_addr`

consider keep the code as it is and use `version` [1] method

```python3
def has_network_access_by_ip_address(server_ip):
    """Check if we can connect to the given server"""
    try:
    except OSError:
        logging.info(
            'Skipping IPv%s protocol version, no network configuration' % ipaddress.ip_address(server_ip).version
            )
        return False

    connection.close()
    return True
```

i.e., using test IP values:
```bash
>>> foo = ipaddress.ip_address('8.8.8.8')
>>> foo.version
4
>>> foo = ipaddress.ip_address('2001:4860:4860::8888')
>>> foo.version
6
```
[1] https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.version

rjschwei: Yeah I considered putting the detection into the `has_network_access_by_ip_address` function then decided against it. I'll think about it some more, it would minimize the overall change set and we already depend on the ipaddress module in the code base as such adding it to this file as a dependency is not really that big of a deal.

Thanks for pointing this out.

rjschwei: Pushed the IP protocol version check into `has_network_access_by_ip_address`

lib/cloudregister/registerutils.py
```diff
@@ -246,7 +246,7 @@ def fetch_smt_data(cfg, proxies, quiet=False):
         except Exception as e:
             logging.error('=' * 20)
             logging.error(str(e))
-            logging.error('Unable to obtain SMT server information, exiting')
+            logging.error('Unable to obtain update server information, exiting')
```
jesusbv, line 249: seems this change is unrelated to this PR or to make network message less alarming ?

jesusbv (approved):
LGTM, thanks for the changes

lib/cloudregister/registerutils.py
```diff
@@ -1484,6 +1484,25 @@ def get_zypper_target_root():
...
 
 
+# ----------------------------------------------------------------------------
+def has_network_access_by_ip_address(server_ip):
+    """Check if we can connect to the given server"""
+    try:
+        connection = socket.create_connection((server_ip, 443), timeout=2)
+    except OSError:
+        try:
+            ip_ver = 'IPv%d' % ipaddress.ip_address(server_ip).version
```
jesusbv, line 1494: looks much better, IMO

lib/cloudregister/registerutils.py
rjschwei: Yes it is, stumbled over it and having a PR to change 1 word is a bit overkill on process.

PR #204, opened 2024-08-27
Remove superfluous condition

rjschwei:
Do not check for the instance metadata file at the point of call, this is already handled by the function implementation.
    
Only create the instance metadata file if we have instance metadata to write,
 this ensures we do not create an empty file that then gets passed to
    downstream invocations hoping that the downstream code copes properly
    with an empty file.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1485,6 +1485,21 @@ def get_zypper_target_root():
 
 
 # ----------------------------------------------------------------------------
+def has_network_access_by_ip_address(server_ip, ip_ver='IPv4'):
+    """Check if we can connect to the given server"""
+    try:
+        connection = socket.create_connection((server_ip, 443), timeout=2)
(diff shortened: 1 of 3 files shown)


jesusbv (changes requested):
I think commit [395f52a](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/204/commits/395f52a5b599fbc154a59e7de4413c50d45e5d9a) does not belong to this PR ? 
as clarified, #203 and #204 are independent

rjschwei:
> I think commit [395f52a](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/204/commits/395f52a5b599fbc154a59e7de4413c50d45e5d9a) does not belong to this PR ? as clarified, #203 and #204 are independent

Yes, I rebased the branch on master before pushing. I guess github has other ideas. I'll start over to make this easier to review and appease GitHub, grmbl.....

PR #205, opened 2024-08-27
Remove superfluous condition

rjschwei:
Do not check for the instance metadata file at the point of call, this is already handled by the function implementation.

Only create the instance metadata file if we have instance metadata to write, this ensures we do not create an empty file that then gets passed to downstream invocations hoping that the downstream code copes properly with an empty file.


jesusbv (approved):
LGTM

PR #207, opened 2024-08-28
LTSS registration user message

rjschwei:
The LTSS registration take a bit longer as such we want the user to know that we are doing something. Print a message to give the user some confidence that we are doing something useful.


jesusbv (approved):
LGTM

usr/sbin/registercloudguest
```diff
@@ -324,6 +324,7 @@ def setup_ltss_registration(
     Run SUSEConnect to register LTSS for this instance
     """
     logging.info('Running LTSS registration...')
```
jesusbv, line 326: we could use `message = 'Running LTSS registration...this takes a little longer'` in the log and print

PR #208, opened 2024-09-04
Remove non free extensions files from /etc/products.d/ directory

jesusbv:
When activating LTSS, a file is created in /etc/products.d:

/etc/products.d/SLES-LTSS.prod

That file makes re registration fail, leaving the system in an inconsistent state because 'registercloudguest --clean' command does not clean that file up

In order to fix that we need to implement a de-registration procedure for the installed non-free product(s)

This Fixes jira#PCT-471


schaefi:
ok when I said I tested it, I should also have said that I tested with the following patch applied

```
diff --git a/lib/cloudregister/registerutils.py b/lib/cloudregister/registerutils.py
index b65b5b5..52cf09e 100644
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -180,14 +180,14 @@ def clean_payg_extensions():
     extensions = get_extensions()
     products_paths = glob.glob('/etc/products.d/*.prod')
 
-    return remove_extension_file(extensions, products_paths)
+    return remove_extension_file(extensions or [], products_paths)
 
 
 def remove_extension_file(extensions, products_paths):
     for extension in extensions:
-        if extensions.get('recommended'):
-            remove_extension_file(extensions.get('extensions'), products_paths)
+        if extension.get('recommended'):
+            remove_extension_file(extension.get('extensions') or [], products_paths)
 
         extension_path = os.path.join(
             'etc', 'products.d', '{}.prod'.format(extension.get('identifier'))
@@ -249,7 +249,7 @@ def get_product_data(registration_target=None):
     url_system_products = 'https://%s/connect/systems/products?%s' % (
         registration_target.get_FQDN(), query_args
     )
-    res = requests.get(url_system_products), auth=auth_creds, headers=headers)
+    res = requests.get(url_system_products, auth=auth_creds, headers=headers)
     if res.status_code != 200:
         err_msg = 'Unable to obtain product information from server "%s"\n'
         err_msg += '\t%s\n\t%s\nUnable to register modules, exiting.'
diff --git a/usr/sbin/registercloudguest b/usr/sbin/registercloudguest
index 69385e6..6198d8c 100755
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -231,7 +231,7 @@ def cleanup():
         utils.clean_payg_extensions()
     except Exception as err:
         print(
-            'Could check the system product data: %s' % err
+            'Could not check the system product data: {0}'.format(err)
         )
     utils.clean_registry_setup()
     utils.remove_registration_data()
```

schaefi:
and this works nicely, I have checked the many combinations of register, with ltss, without, cleanup, re-register, etc etc. It did the right thing

jesusbv:
> ok when I said I tested it, I should also have said that I tested with the following patch applied
> 
> ```diff
> diff --git a/lib/cloudregister/registerutils.py b/lib/cloudregister/registerutils.py
> index b65b5b5..52cf09e 100644
> --- a/lib/cloudregister/registerutils.py
> +++ b/lib/cloudregister/registerutils.py
> @@ -180,14 +180,14 @@ def clean_payg_extensions():
>      extensions = get_extensions()
>      products_paths = glob.glob('/etc/products.d/*.prod')
>  
> -    return remove_extension_file(extensions, products_paths)
> +    return remove_extension_file(extensions or [], products_paths)
>  
>  
>  # ----------------------------------------------------------------------------
>  def remove_extension_file(extensions, products_paths):
>      for extension in extensions:
> -        if extensions.get('recommended'):
> -            remove_extension_file(extensions.get('extensions'), products_paths)
> +        if extension.get('recommended'):
> +            remove_extension_file(extension.get('extensions') or [], products_paths)
>  
>          extension_path = os.path.join(
>              'etc', 'products.d', '{}.prod'.format(extension.get('identifier'))
> @@ -249,7 +249,7 @@ def get_product_data(registration_target=None):
>      url_system_products = 'https://%s/connect/systems/products?%s' % (
>          registration_target.get_FQDN(), query_args
>      )
> -    res = requests.get(url_system_products), auth=auth_creds, headers=headers)
> +    res = requests.get(url_system_products, auth=auth_creds, headers=headers)
>      if res.status_code != 200:
>          err_msg = 'Unable to obtain product information from server "%s"\n'
>          err_msg += '\t%s\n\t%s\nUnable to register modules, exiting.'
> diff --git a/usr/sbin/registercloudguest b/usr/sbin/registercloudguest
> index 69385e6..6198d8c 100755
> --- a/usr/sbin/registercloudguest
> +++ b/usr/sbin/registercloudguest
> @@ -231,7 +231,7 @@ def cleanup():
>          utils.clean_payg_extensions()
>      except Exception as err:
>          print(
> -            'Could check the system product data: %s' % err
> +            'Could not check the system product data: {0}'.format(err)
>          )
>      utils.clean_registry_setup()
>      utils.remove_registration_data()
> ```

There may be a cleaner way to handle your suggestion

```python
    return prod_data.get('extensions', [])
```
[here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/208/files#diff-84460adc47c95f16ae220daac1882a7bc19b595d1ba01650e286b666ef14f2a4R231)

and

```python
extensions.get('extensions', [])
```
for the others

schaefi:
sure whatever you prefer, just go ahead and push it :) I was just pasting what was needed to fix it, feel free to make it nicer :)

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,90 @@ def clear_rmt_as_scc_proxy_flag():
...
+
+    for extension in extensions:
+        if not extension.get('free'):
+            extension_path = os.path.join(
+                os.sep, 'etc', 'products.d', '{}.prod'.format(
+                    extension.get('identifier')
+                )
+            )
+            if extension_path in products_paths:
+                os.unlink(extension_path)
```
jesusbv, line 192: maybe we should log here the file(s) we are removing

schaefi: yes please add a log message for the removed .prod files

schaefi:
I tested the latest code and it works for me too :+1: 

Please note it requires the latest version of the RMTServer that has the `free` patch applied. Without that change on the server the code in `clean_payg_extensions()` which checks `if not extension.get('free'):` will not apply.

So this looks good

schaefi:
From my perspective this is no longer a work in progress and can be opened for review and merge. Thanks

lib/cloudregister/registerutils.py
jesusbv: added it just right after adding the suggestion in [be49517](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/208/commits/be495177d3726f5fe4e68c424c8d0bba741218ed)

schaefi (approved):
Looks great :+1: tested and works for me. logging is fine, thanks for the inline comment about the server side capability this helps when reading the code.

From a logic point of view we remove all "non-free" marked extensions if the base product is gone. On repeat of the registration all non-free extensions e.g LTSS must re-register too. Unless there are objections to this action we are good.

I'd like to wait for Robert's feedback too. From my perspective this is good

jesusbv:
> Looks great 👍 tested and works for me. logging is fine, thanks for the inline comment about the server side capability this helps when reading the code.
> 
> From a logic point of view we remove all "non-free" marked extensions if the base product is gone. On repeat of the registration all non-free extensions e.g LTSS must re-register too. Unless there are objections to this action we are good.
> 
> I'd like to wait for Robert's feedback too. From my perspective this is good

Tobe on the same page and precise
> if the base product is gone

we remove the non free extension _before_ de-registering/remving the base product, otherwise, there is no way to fetch that info from the server side

jesusbv:
@rjschwei the PR is ready for review, no longer a draft

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
...
+    return product_type(
+        name=product_tree.find('name').text,
+        version=product_tree.find('version').text,
+        arch=product_tree.find('arch').text
+    )
+
+
+# ----------------------------------------------------------------------------
+def get_extensions():
+    """Get the extensions of the base product on RMT."""
```
jesusbv, line 231: s/RMT/update server ?

schaefi: yeah maybe the better wording. We have some inconsistencies in the code regarding smt/rmt/update-server but I like the idea to expose messages that talks about "update server" rather than RMT.

jesusbv:
@rjschwei PR is ready for review

rjschwei:
I am not sure how a PR can be ready for review when we have not agreed to what we are going to do in the corresponding feature request. Converted to draft until such agreement is reached.

schaefi:
> I am not sure how a PR can be ready for review when we have not agreed to what we are going to do in the corresponding feature request. Converted to draft until such agreement is reached.

Because a tool that can setup something but not properly cleaning it up is not missing a feature, it's simply buggy and should be fixed, which is what this PR suggests. The part I don't understand is why a conversation to reach an agreement how the cleanup should happen cannot be done based on an implementation suggestion ? hence here in this PR

like @jesusbv, I don't understand it. Frankly speaking this is questionable behavior. You were asked to be a reviewer, instead you implicitly declined a review and turned it into a draft which is an action that I personally think only the author of code is eligible to perform, hence in my world this is considered disrespectful to the people who did the work. I'm sure this was not the intention ?

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
         pass
 
 
+# ----------------------------------------------------------------------------
+def clean_payg_extensions():
```
rjschwei, line 179: While the extensions we are removing are paid extensions they are not PAYG technically, using the `payg` moniker here can be confusing to the reader, I think we should stick with the `non_free` term, i.e. `clean_non_free_extensions`

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
...
+            extension_path = os.path.join(
+                os.sep, 'etc', 'products.d', '{}.prod'.format(
+                    extension.get('identifier')
+                )
+            )
+            if extension_path in products_paths:
+                os.unlink(extension_path)
+                logging.info(
+                    'Non free extension file %s removed' % extension_path
+                )
```
rjschwei, line 198: We cannot just remove the `.prod` file. The product file was put there by an RPM package. By simply removing the prod file we put the system into an inconsistent state. The rpm database will say that the package that put the prod file there is installed but the file will be missing making it more difficult to understand any issues that may arise. We have to keep the system in a consistent state, meaning we have to remove the corresponding `-release` package that put the product file in place.

Should the user ever run an RPM DB consistency check they will get an error because of a problem we created, that is not an option.

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
         pass
 
 
+# ----------------------------------------------------------------------------
+def clean_payg_extensions():
+    """Remove non free extensions files from /etc/products.d/."""
+    extensions = get_extensions()
+    products_paths = glob.glob('/etc/products.d/*.prod')
```
rjschwei, line 182: We should let zypper tell us what products are on the system. zypper already has a reliable way to do this and verifies things along the way. If changes are made to this type of identification there will be a feature ticket that will get implemented in zypper but we will be forgotten and our code will break. We should depend on teh distribution implementation in this case.

And we already have an implementation for this ;) `get_installed_products()` in registerutils.

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
...
+            )
+            if extension_path in products_paths:
+                os.unlink(extension_path)
+                logging.info(
+                    'Non free extension file %s removed' % extension_path
+                )
+
+
+# ----------------------------------------------------------------------------
+def get_product_tree(product_file=BASE_PRODUCT_PATH):
```
rjschwei, line 202: What does moving the `get_product_tree` and `get_product_triplet` functions from the command implementation have to do with the cleanup of the non free extensions? This adds an unnecessary distraction to the task ad hand.

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,96 @@ def clear_rmt_as_scc_proxy_flag():
...
+
+# ----------------------------------------------------------------------------
+def get_extensions():
+    """Get the extensions of the base product on the update server."""
+    prod_data = get_product_data()
+    return prod_data.get('extensions', [])
+
+
+# ----------------------------------------------------------------------------
+def get_product_data(registration_target=None):
```
rjschwei, line 237: This is being moved from the command implementation for no apparent reason related to the clean up but then also not used in the command implementation....?????

rjschwei (changes requested):

lib/cloudregister/registerutils.py
jesusbv: good catch

lib/cloudregister/registerutils.py
jesusbv: > What does moving the get_product_tree and get_product_triplet functions from the command implementation have to do with the cleanup of the non free extensions?

`clean_payg_extensions` method calls `get_extensions` method which calls `get_product_data` method which calls `get_product_triplet`

lib/cloudregister/registerutils.py
jesusbv: as the method s/`clean_payg_extensions`/`clean_non_free_extensions` uses these methods
tests would have failed

I guess we can try to mock those calls but seemed good to have them tested as, at the time of this PR, there were no tests for `registercloudguest`

lib/cloudregister/registerutils.py
schaefi: yes this is true. I overlooked that in the review and testing.

```
registercloudguest --regcode ...

ls -l /etc/products.d/
total 16
-rw-r--r-- 1 root root 2679 Aug  1 12:52 SLES-LTSS.prod
...
```

As we only call SUSEConnect I thought that SLES-LTSS.prod was created by SUSEConnect, but in fact a package was installed

```
rpm -qf /etc/products.d/SLES-LTSS.prod
sles-ltss-release-15.4-150400.13.8.1.x86_64
```

ok, so yes we cannot simply remove the .prod file, thanks for catching this one.

It actually impacts some parts of the implementation.

lib/cloudregister/registerutils.py
schaefi: and as I have learned SUSEConnect installs and de-installs packages the correct way to de-register would then be another SUSEConnect call

```
SUSEConnect --de-register --product SLES-LTSS/15.4/x86_64
```

schaefi:
@rjschwei 
>  What does moving the get_product_tree and get_product_triplet functions from the command implementation have to do with the cleanup of the non free extensions? This adds an unnecessary distraction to the task ad hand.

I don't think so, as Jesus mentioned the move happened due to the nesting of the new methods and is also required for proper unit testing. That's why I didn't had any issue with this move.

lib/cloudregister/registerutils.py
jesusbv: also, `get_extensions` can get re-used in `registercloudguest`

lib/cloudregister/registerutils.py
jesusbv: Yes, that is the approach I took

lib/cloudregister/registerutils.py
```diff
@@ -173,6 +175,110 @@ def clear_rmt_as_scc_proxy_flag():
...
+            triplet = get_product_triplet(extension)
+            triplet = '{name}/{version}/{arch}'.format(
+                name=triplet.name,
+                version=triplet.version,
+                arch=triplet.arch
+            )
+            if triplet in installed_products:
+                return_code = exec_subprocess(
+                    ['SUSEConnect', '-d', '-p', triplet]
+                )
```
schaefi, line 198: For calling SUSEConnect I wrote a wrapper method named `run_SUSEConnect`. I believe it would be good if any SUSEConnect call is handled by the wrapper. One reason to do so is because there is a difference on `SUSEConnect` vs `transactional register`. I guess there is also a `transactional deregister`

jesusbv: I thought to use it but refrained myself to move another method to `registerutils.py`

jesusbv: will adapt the code (move the method implementation and more changes) to use it in `registerutils`

jesusbv: Done in [f2e9acc](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/208/commits/f2e9acc78ba8cdcced60e2dd0cd9281707abac17) following the doc:

```bash
# De-register the system and base product
transactional-update register -d
 
# De-register a single product/extension
transactional-update register -d -p SUSE-Manager-Server/5.0/x86_64
```
Failed because of the linter

schaefi: nice :+1:

lib/cloudregister/registerutils.py
schaefi: @rjschwei I'm about to close this thread as resolved now as I believe Jesus provided the needed information. As I said from my perspective the involved refactoring/code-moves are good for the requested change and finally also adds tests for crucial code like product triplet or SUSEConnect  invocation.

If this change should be split into a refactor first, let's discuss at the workshop. I would be good as I think the code move plus tests is still good to review

lib/cloudregister/registerutils.py
schaefi: @rjschwei I'm about to close this thread as resolved now as I believe Jesus provided the needed information. As I said from my perspective the involved refactoring/code-moves are good for the requested change

If this change should be split into a refactor first, let's discuss at the workshop. I would be good as I think the code move plus tests is still good to review

lib/cloudregister/registerutils.py
```diff
@@ -225,6 +333,125 @@ def exec_subprocess(cmd, return_output=False):
...
+    cmd = [register_cmd]
+    target_url = registration_target.get_FQDN()
+
+    # distinguish command between standard and read-only system(transactional)
+    if 'transactional' in register_cmd:
+        cmd += ['register']
+
+    cmd += ['--url', 'https://{0}'.format(target_url)]
+
+    if de_register:
```
jesusbv, line 414: I think we should check if `de_register` option is used with a product, 
otherwise, it would fail as `registercloudguest` is present

schaefi:
run integration test procedures

```
registercloudguest -r LTSS-code
registercloudguest --clean
registercloudguest
registercloudguest -r LTSS-code
```

All working :+1:

PR #209, opened 2024-09-10
package build update

rjschwei:
- The way Python is being built has changed for SUSE distributions. Python 3.6 is, by design no longer available in products SLE Micro and later. We need to use macros that are defined in the build service to pick up the interpreter we want to use for building our code and to find the proper Python module packages we depend on.
- Add a previoulsy forgotten patch for the SLE 12 build that we need to maintain until the EOL of that code stream.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -15,6 +15,12 @@
 # Please submit bugfixes or comments via https://bugs.opensuse.org/
 #
 
+%if 0%{?suse_version} >= 1600
+%define pythons %{primary_python}
+%else
+%define pythons python3
+%endif
+%global _sitelibdir %{%{pythons}_sitelib}
 
 %define base_version 10.3.4
 Name:           cloud-regionsrv-client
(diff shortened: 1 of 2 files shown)


rjschwei:
Sticking with the Python 3.6 build for SLE 15. When SLE 15 SP3 reaches EOL we can consider switching this to a Python 3.11 build and creating a branch in the build service. Until then I'd rather not have a branch in the build service to maintain a Python 3.6 build for SP1, SP2, and SP3 and a Python 3.11 build for SP4 and later.

cloud-regionsrv-client.spec
```diff
@@ -29,6 +35,8 @@ Source0:        %{name}-%{version}.tar.bz2
 Patch0:         fix-for-sles12-disable-ipv6.patch
 # PATCH-FIX-SLES12 fix-for-sles12-disable-registry.patch
 Patch1:         fix-for-sles12-disable-registry.patch
+# PATCH-FIX-SLES12 fix-for-sles12-no-trans_update.patch
+Patch2:         fix-for-sles12-no-trans_update.patch
```
schaefi, line 39: The patch file `fix-for-sles12-no-trans_update.patch` is not part of the git repo. Can you please add it like the other patch files. Thanks

schaefi: I tested the package build for all maintained targets. One thing I noted is that for SLE15 the package builds against python 3.6, I guess this is intentional ?

schaefi (changes requested):
Looks good, I tested the package build except for SLE12 due to the missing patch file in the git. Thanks

schaefi (approved):
LGTM

PR #211, opened 2024-09-20
Add registercloudguest tests and coverage

jesusbv:
It happens (and it has happened) that some errors/bugs come from registercloudguest due to 
either missing things or having things not correctly set which would get caught if registercloudguest had tests

Some corrections were done even in this PR as they were found during the tests addition

This PR aims to reduce the amount of missing things in this project

- Modify registercloudguest to call `main()`
- Add tests to registercloudguest
- Add a check to keep the coverage for registercloudguest at 100%
- Add linter check for registercloudguest

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -37,8 +37,10 @@ passenv =
 deps = {[testenv]deps}
 commands =
     python3 setup.py develop
-    py.test --no-cov-on-fail --cov=lib/cloudregister \
-        --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc
+    py.test --no-cov-on-fail \
+    --cov=lib/cloudregister \
+    --cov=usr/sbin \
+    --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc
 
 
 [testenv:unit_py3_11]
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -193,7 +193,6 @@ def register_modules(
         identifier = extension.get('identifier')
         version = extension.get('version')
         triplet = '/'.join((identifier, version, arch))
-        print(triplet in products and triplet not in registered)
         if triplet in products and triplet not in registered:
             registered.append(triplet)
             suseconnect = run_SUSEConnect(
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -453,7 +453,6 @@ argparse.add_argument(
 )
 
 
-
 def main(args):
     global registration_returncode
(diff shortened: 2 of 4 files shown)


tox.ini
```diff
@@ -37,8 +37,10 @@ passenv =
 deps = {[testenv]deps}
 commands =
     python3 setup.py develop
-    py.test --no-cov-on-fail --cov=lib/cloudregister \
-        --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc
+    py.test --no-cov-on-fail \
+    --cov=lib/cloudregister \
+    --cov=usr/sbin \
+    --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc
```
schaefi, line 43: :+1:

usr/sbin/registercloudguest
```diff
@@ -451,425 +453,440 @@ argparse.add_argument(
 )
 
 
-args = argparse.parse_args()
+def main(args):
+    global registration_returncode
```
schaefi, line 457: while this is hard to review, I'm just asking. The change here was adding a main() method and all code was moved into it right ? or did you also change code paths that you want us to review ?

Thanks

schaefi (approved):
I really like we test all code, thanks much for this effort

I was a bit concerned about the big changeset in registercloudguest but I guess it's only caused due to the new main() function and I also assume you did that to allow for testing it better. Thus it looks good but please confirm the main code is only moved code correct ?

jesusbv:
> I really like we test all code, thanks much for this effort

Thanks
> 
> I was a bit concerned about the big changeset in registercloudguest 

the change in registercloudguest is to include

```python
if __main__:
  main()
  ```
  then args as params and some one line changes because of bugs found during testing
  
> but I guess it's only caused due to the new main() function and I also assume you did that to allow for testing it better. Thus it looks good but please confirm the main code is only moved code correct ?

the big change is the moving, yes, inside the `def main` method, then indentation needed
then after that I caught some errors as I was implementing the tests, like we were calling a method that has a parameter without parameters at all for example, or using a non-declared variable, things like that

99% of the changes are the tests
1% is the moving and small bug fixes

usr/sbin/registercloudguest
jesusbv: correct, introduction to `main(args)` method

schaefi (changes requested):
After a closer conversation on the changes it seems there are more code changes which are hard to review due to the changes done for the tests. I agreed with @jesusbv that we aim for a commit set which allows to review the code changes found by the testing effort individually

jesusbv:
Closing to create a new one with cleaner commit set as agreed

PR #212, opened 2024-09-23
Registercloudguest coverage

jesusbv:
It happens (and it has happened) that some errors/bugs come from registercloudguest due to
either missing things or having things not correctly set which would get caught if registercloudguest had tests

Some corrections were done (each of them has its own commit) even in this PR as they were found during the tests addition

This PR aims to reduce the amount of missing things in this project

- Modify registercloudguest to call main()
- Add tests to registercloudguest
- Add a check to keep the coverage for registercloudguest at 100%
- Add linter check for registercloudguest

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -174,7 +174,10 @@ def run_SUSEConnect(
 
 
 # ----------------------------------------------------------------------------
-def register_modules(extensions, products, registered=[], failed=[]):
+def register_modules(
+    extensions, products, registration_target, instance_data_filepath,
+    registered=[], failed=[]
+):
     """Register modules obeying dependencies"""
     global registration_returncode
     for extension in extensions:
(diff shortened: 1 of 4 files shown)


schaefi (commented):
Thanks much, nice to review. So far I have only one question with regards to sle12 compatibility.

There are also patch files
* fix-for-sles12-disable-ipv6.patch
* fix-for-sles12-disable-registry.patch
* fix-for-sles12-no-trans_update.patch

Can you check if those still apply after your changes ?

Thanks

usr/sbin/registercloudguest
```diff
@@ -833,7 +833,7 @@ def main(args):
         logging.error(err_msg % (ips, res.reason, res.content.decode("UTF-8")))
         sys.exit(1)
 
-    prod_data = json.loads(res.text)
+    prod_data = res.json()
```
schaefi, line 836: Is this compatible with python 3.4 which we have on SLE12 ?

jesusbv: Yes, it is

jesusbv: checked with the documentation and also tested (double check) that works for `Python 3.4.10`

schaefi (approved):
Looks good, thanks

PR #213, opened 2024-10-01
Keep free registry url

jesusbv:
When cleaning up the registry info, we remove bith the private and the public registry url, we should keep the public url

Otherwise, a command like

podman search bci

would not produce any results when it should show at least any match on registry.suse.com


schaefi (approved):
Yep, let's not mess with the public registries and only cleanup what we added ourselves. Look good :+1:

PR #216, opened 2024-10-23
Tidy up transactional-update integration

rjschwei:
The transactional-update command integration was done inside the run_SUSEConnect function as such creating a naming mismatch since the function no longer only called SUSEConnect. Rename the function to a generic name register_product and factor the detection about use of the transactional-update command out into a separate function to be re-used.

### DIFF
--- a/tests/test_registercloudguest.py
+++ b/tests/test_registercloudguest.py
@@ -3,6 +3,7 @@ import json
 import os
 import requests
 
+from io import StringIO
 from collections import namedtuple
 from importlib.machinery import SourceFileLoader
 from lxml import etree
@@ -982,7 +983,7 @@ def test_register_cloud_guest_force_reg_cert_import_failed(
 
 
 @patch('cloudregister.registerutils.set_proxy')
-@patch('cloudregister.registerutils.run_SUSEConnect')
+@patch('cloudregister.registerutils.register_product')
 @patch('cloudregister.registerutils.import_smt_cert')
 @patch('cloudregister.registerutils.get_installed_products')
 @patch('register_cloud_guest.logging')
@@ -1029,7 +1030,7 @@ def test_register_cloud_guest_force_baseprod_registration_failed(
     mock_clean_hosts_file, mock_add_hosts_entry, mock_has_registry_in_hosts,
     mock_update_rmt_cert, mock_get_register_cmd, mock_os_access,
     mock_set_as_current_smt, mock_logging, mock_get_installed_products,
-    mock_import_smt_cert, mock_run_SUSEConnect, mock_set_proxy
+    mock_import_smt_cert, mock_register_product, mock_set_proxy
 ):
     smt_data_ipv46 = dedent('''\
         <smtInfo fingerprint="AA:BB:CC:DD"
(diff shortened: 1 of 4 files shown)


rjschwei:
@jesusbv thanks for the effort on improving the unit testing. That said they also need to be implemented such that they are de-bugable on failure, which a test with 40+ mocks is IMNSHO not. Anyway, I'd appreciate it if you could sort out the test failure, otherwise I am fine dropping that test. This is ultimately covered in integration testing.

rjschwei:
@schaefi no version bump after this merge please, there are a couple of more things I am working on.

schaefi:
> @schaefi no version bump after this merge please, there are a couple of more things I am working on.

ok, I would not have done this anyway. I would bump versions when we are about to do the next round of releases. Thus no worries ;)

schaefi:
@rjschwei I like the change :+1: 

The code looks good, but please no merge without working tests. I can assist with the tests too if you like. Since you directly asked @jesusbv I'll wait for him to comment

jesusbv:
> a test with 40+ mocks is IMNSHO not [de-bugable]

I completely agree, after sorting the unit tests I've been working on split the responsabilities of `registercloudguest` code, similar to what is done in this PR, aiming to help implementing tests, among other benefits

> I'd appreciate it if you could sort out the test failure

No problem, as @schaefi mentioned, I also prefer to have the test, even if tedious to implement
I have no problem sorting the test failure out

schaefi:
@jesusbv I found a quick fix to the test issue and suggested a change in code for which I added a commit. @jesusbv I hope you are ok with it, as it does not only touch the tests but also proposes a small change in @rjschwei implementation I was so free to push this commit. Let me know what you think 

Thanks

tests/test_registercloudguest.py
```diff
@@ -2108,14 +2108,18 @@ def test_register_cloud_baseprod_ok_recommended_extensions_failed(
         'plugin:/susecloud?credentials=Basesystem_Module_x86_64&'
         'path=/repo/SUSE/Updates/SLE-Module-Basesystem/15-SP4/x86_64/update/'
     )
-    mock_exec_subprocess.side_effect = [True, False]
+    findmnt_return = b'{"filesystems": [{"target": "/","source": ' + \
+        b'"/dev/sda3","fstype": "xfs","options": "rw"}]}'
+    mock_exec_subprocess.return_value = findmnt_return, b'', 0
```
schaefi, line 2113: I added a test based on the real output of findmnt

schaefi: interesting that it works on py 3.11 but seems to fail on 3.6

jesusbv: Thanks for the changes, AFAICT it's missing the lines inside the if
```python
        if utils.is_transactional_system():
            activate_prod_cmd = 'transactional-update register -p {} '
            activate_prod_cmd += '-r ADDITIONAL REGCODE'
```
because the options of filesystems need a `ro`

```python
else:
        fsinfo = json.loads(output.decode())
        fsdata = fsinfo.get('filesystems')
        if fsdata:
            fsoptions = fsdata[0].get('options')
            # If we are on a RO system we need to use the
            # transactional-update command
            if 'ro' in fsoptions.split(','):
```

jesusbv: will take a look at that test

jesusbv: the `ro` change works

rjschwei:
@schaefi @jesusbv thanks for the help with sorting out the tests.

schaefi:
@rjschwei you are welcome, feel free to rebase -i the commits into one and force push, then this is ready for merge

schaefi (approved):

jesusbv (approved):
LGTM

PR #217, opened 2024-10-24
Use the region endpoint of IMDS in EC2

rjschwei:
A while back EC2 IMDS gained a "region" endpoint to determine the region in which the instance is running. Using this endpoint directly allows us to drop the code that determined the region based on the information from the "availability-zone" end point. This simplifies the code and testing. Further this is more robust should AWS expand the zone naming as has happened in the past.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -113,7 +113,7 @@ Requires:       cloud-regionsrv-client >= 6.0.0
 Guest registration plugin for images intended for Google Compute Engine
 
 %package plugin-ec2
-Version:        1.0.4
+Version:        1.0.5
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Amazon EC2
 Group:          Productivity/Networking/Web/Servers
--- a/cloud-regionsrv-client.changes
+++ b/cloud-regionsrv-client.changes
@@ -1,15 +1,20 @@
 -------------------------------------------------------------------
-Thu Oct  3 10:31:39 UTC 2024 - Marcus Schäfer <marcus.schaefer@suse.com>
+Upcoming 10.3.6
 
-- Deregister non free extensions at registercloudguest --clean
-- Fix registry cleanup at registercloudguest --clean, don't remove files
-- Prevent duplicate search entries in registry setup
+- Update to 10.3.6 (jsc#PCT-471)
+  + Deregister non free extensions at registercloudguest --clean
+  + Fix registry cleanup at registercloudguest --clean, don't remove files
+  + Prevent duplicate search entries in registry setup
+- Update EC2 plugin to 1.0.5
+  + Switch to using the region endpoint from IMDS to determine the region
+    instead of deriving the data from the availability zone
 
 -------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


cloud-regionsrv-client.spec
```diff
@@ -113,7 +113,7 @@ Requires:       cloud-regionsrv-client >= 6.0.0
 Guest registration plugin for images intended for Google Compute Engine
 
 %package plugin-ec2
-Version:        1.0.4
+Version:        1.0.5
```
schaefi, line 116: It's not ok to bump the version in a pull request. Version updates should be done via 

```
bumpversion major|minor|patch
```

at the time when we prepare for a new release such that all commits and fixes between the version and its tags can be simply detected and collected and also compared with the changes entry to be correct and complete. Please revert that part of the PR, it also doesn't seem to be done via bumpversion but rather manually which increases the risk to forget source paths that uses version information

schaefi (changes requested):
See my comment on the version change

Other than that I have not enough knowledge about the changed endpoint to verify that this works. From a code perspective it looks clean to me. I assume you have integration tested the change ?

cloud-regionsrv-client.spec
schaefi: Also see `.bumpversion.cfg` file in the sources for a better understanding what a version bump does

schaefi: If the plugin sub-packages should take different versions than the main package I'm ok with your change, if not we should make them bumpversion aware.

Thanks

schaefi: Looking closer to the spec I see that every sub-package has its own version, why is it done that way ? This makes maintenance more complicated in my eyes and also harms the concept of a major version for cloud-regionsrv-client unless the sub-packages are completely independent of the main package, which I believe is not the case

rjschwei: The sub packages (plugins) follow their own versioning. The reason behind this is that for the most part (99+%) we do not want to release new plugins when the main implementation changes. This separation reduces churn in the maintenance queue of SLE since code changes to the plugins are very rare.

schaefi: well maybe they are completely independent. If they are we can keep this. Please confirm ?

rjschwei: Yes they are independent. A plugin has to provide the `generateRegionSrvArgs()` function. That is the only connection point to the rest of the code base.

rjschwei: And a little more detail. The plugin is configured via `regionsrvclnt.cfg` as `instanceArgs` option of the `instance`, as such it is completely optional. The plugin is loaded via `__get_framework_plugin` in `registerutils.py`.

Further evidence that these are independent is that the gce plugin has not changed since it's initial implementation and remains at `1.0.0`. The Azure plugin has only changed once when IMDS support was added in addition to supporting the wireserver protocol. The EC2 plugin had some iterations because of the introduction of local zones and other changes to the data of the `availability-zone` end point. With eth switch to the `region` endpoint we are hopefully done touching that code.

schaefi: ok thanks for clarification

schaefi (approved):

jesusbv (approved):
LGTM

PR #218, opened 2024-10-24
Fixed ipv4/ipv6 address detection from string

schaefi:
The get_hosts() method classifies the IP addresses into ipv4 and ipv6. The way it was done was not quite correct. This commit fixes it.


schaefi:
I stumbled over this wile learning a bit more about IPv6 addresses and their representation

rjschwei (approved):

PR #219, opened 2024-10-24
Spec clean

rjschwei:
Addresses various issues related to RPM build for SLE 15 and SLE 12 code streams.

- Fixes a bug (bsc#1230615) with the sudo rule setup, there is a separate issue with conflicting directory ownership, both issues are fixed
- The patch to disable registry related setup did not longer apply in SLE 12, fixed
- Removed registry related files from the SLE 12 package, since we do not support registry setup in SLE 12 these should not be there
- Cleaned up some rpmlint warnings for description length

### DIFF
--- a/cloud-regionsrv-client.changes
+++ b/cloud-regionsrv-client.changes
@@ -1,7 +1,15 @@
 -------------------------------------------------------------------
 Upcoming 10.3.6
 
-- Update to 10.3.6 (jsc#PCT-471)
+- Update to 10.3.6 (jsc#PCT-471, bsc#1230615)
+  + Fix sudo setup
+    ~ permissions cloudguestregistryauth
+    ~ directory ownership /etc/sudoers.d
+  + spec file
+    ~ Remove traces of registry related entries on SLE 12
+  + Forward port
+    ~ fix-for-sles12-disable-registry.patch
+    ~ fix-for-sles12-no-trans_update.patch
   + Deregister non free extensions at registercloudguest --clean
   + Fix registry cleanup at registercloudguest --clean, don't remove files
   + Prevent duplicate search entries in registry setup
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -1,7 +1,7 @@
 #
 # spec file for package cloud-regionsrv-client
 #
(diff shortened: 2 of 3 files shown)


schaefi (approved):
Nice this looks good :+1: 

Thanks for getting rid of the .orig patch info.

I think it would be good to catch up quickly regarding the way how we add information to the .changes file. The way you do it at the moment is to have one entry prior the next release and put data under the "Update to ..." headline. We can do it that way but I usually do this differently. We have bumpversion in place which creates a commit and tag of the same format e.g. `Bump version: 10.3.4 → 10.3.5`. When we create PRs a number of new commits between the last version bump accumulates. If we go for a release we can very easily collect the changes, create a .changes entry that summarizes them, bump the version and submit the package.

I usually prefer this way because that ensures nothing important gets forgotten in the .changes information. It would be nice to have a short conversation about this also including @jesusbv such that we at least consistently do the same thing. Thanks

rjschwei:
> Nice this looks good 👍
> 
> Thanks for getting rid of the .orig patch info.
> 
> I think it would be good to catch up quickly regarding the way how we add information to the .changes file. The way you do it at the moment is to have one entry prior the next release and put data under the "Update to ..." headline. We can do it that way but I usually do this differently. We have bumpversion in place which creates a commit and tag of the same format e.g. `Bump version: 10.3.4 → 10.3.5`. When we create PRs a number of new commits between the last version bump accumulates. If we go for a release we can very easily collect the changes, create a .changes entry that summarizes them, bump the version and submit the package.
> 

Copied that loosely from what the SCC team does for RMT where it is the PRs author's responsibility to add the changelog entry, which I like because....

> I usually prefer this way because that ensures nothing important gets forgotten in the .changes information.

"importance" is in the eye of the beholder. For example an internal refactor (i.e. no effect on user visible behavior) is important but for the most part of no interest to the consumer of the package, as such can be left out of the changelog. Also if the changelog is created by the person doing the version bump then it becomes that person's responsibility to summarize the work of others for the changelog and decide what is "important" to mention and what is not. 

> It would be nice to have a short conversation about this also including @jesusbv such that we at least consistently do the same thing. Thanks

PR #220, opened 2024-11-04
Add patch validation per pull request

schaefi:
When we change code this can impact the ability to apply the custom patches that we carry with the package. This commit adds a github action to apply the patches to the changed source to make sure all custom patches still applies.


jesusbv (approved):
LGTM

rjschwei (approved):

PR #221, opened 2024-11-05
Fix LTSS registration

rjschwei:
LTSS as an add on product has always SLES-LTSS as the product identifier in the product triplet even on other SLE products such as SLES For SAP and SLE HPC. Update the code to reflect this naming oddity.

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -208,12 +208,10 @@ def setup_ltss_registration(
         print(message)
     else:
         base_product = utils.get_product_triplet(product)
-
-        # The name of the product to register is the LTSS variant for the
-        # registered base product. In theory other combinations are possible
-        # but we don't allow this for the moment
-        ltss_product_triplet = '{0}-LTSS/{1}/{2}'.format(
-            base_product.name, base_product.version, base_product.arch
+        # LTSS registration is always SLES even for variants such as
+        # SAP and HPC.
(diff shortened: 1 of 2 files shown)


usr/sbin/registercloudguest
```diff
@@ -208,12 +208,10 @@ def setup_ltss_registration(
...
     else:
         base_product = utils.get_product_triplet(product)
-
-        # The name of the product to register is the LTSS variant for the
-        # registered base product. In theory other combinations are possible
-        # but we don't allow this for the moment
-        ltss_product_triplet = '{0}-LTSS/{1}/{2}'.format(
-            base_product.name, base_product.version, base_product.arch
+        # LTSS registration is always SLES even for variants such as
+        # SAP and HPC.
```
schaefi, line 212: This is sooooo weird, who in Gods name invented this model ??? anyway, thanks for fixing it. We will be screwed though if they ever change the LTSS model regarding the base product name in the future

schaefi (approved):
it doesn't look good but it seems to be the only way

usr/sbin/registercloudguest
jesusbv: I guessed and was kinda wishing I was wrong the fix would be hardcoding `SLES`, as Marcus said, fingers crossed

jesusbv (approved):

PR #222, opened 2024-11-13
Make sure to require container toolchain

schaefi:
cloud-regionsrv-client modifies config files from the container toolchain, docker and podman. These files are provided by packages under certain conditions, e.g. permission and ownership. If the files are not present on the system when registercloudguest is called they are created under the umask() of the calling user which is not neccessarily correct. Because of that, this commit adds a requirement to the respective packages providing the config files such that any modifications retains the permission and ownership.


schaefi:
This needs testing prior merge. I will create a package with the change and check the build status

rjschwei (approved):
Please do not bump the version yet, one more change coming I'd like to include in a 10.3.8 release tat is related to changes in suseconnect

jesusbv (approved):

PR #223, opened 2024-11-14
Update pipe to check for error message

rjschwei:
suseconnect is changing to write error messages to STDERR instead of STDOUT, as such we need to check both pipes for the error message to avoid having a hard version dependency. Further the error message has changed and we need adjust for this condition as well.

Reference: https://github.com/SUSE/connect-ng/pull/272/files

### DIFF
--- a/tests/test_registercloudguest.py
+++ b/tests/test_registercloudguest.py
@@ -2370,6 +2370,43 @@ def test_register_modules(mock_register_product):
     )
 
 
+@patch('cloudregister.registerutils.register_product')
+def test_register_modules_failed_credentials(mock_register_product):
+    prod_reg_type = namedtuple(
+        'prod_reg_type', ['returncode', 'output', 'error']
+    )
+
+    mock_register_product.return_value = prod_reg_type(
+        returncode=67,
+        output='',
+        error='missing system credentials try again'
+    )
+    extensions = [
+        {
+            'id': 23,
+            'name': 'SUSE Linux Enterprise Server LTSS',
+            'identifier': 'SLES-LTSS',
+            'former_identifier': 'SLES-LTSS',
+            'version': '15.4',
+            'release_type': None,
+            'release_stage': 'released',
+            'arch': 'x86_64',
+            'friendly_name':
+            'SUSE Linux Enterprise Server LTSS 15 SP4 x86_64',
+            'product_class': 'SLES15-SP4-LTSS-X86',
+            'free': False,
+            'repositories': [],
+            'product_type': 'extension',
+            'extensions': [],
+            'recommended': False,
+            'available': True
+        }
+    ]
+    register_cloud_guest.register_modules(
+        extensions, ['SLES-LTSS/15.4/x86_64'], 'reg_target', 'path', [], []
+    )
+
+
 @patch('cloudregister.registerutils.is_registry_registered')
 @patch('cloudregister.registerutils.get_credentials_file')
 @patch('cloudregister.registerutils.get_credentials')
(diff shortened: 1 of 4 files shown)


schaefi:
@rjschwei yep code change makes complete sense to me and looks good. But you missed to add a test for the new condition. Can you please add one, then we are good to go from my side

usr/sbin/registercloudguest
```diff
@@ -86,11 +86,16 @@ def register_modules(
...
                 error_message = prod_reg.output
+                if not error_message:
+                    error_message = prod_reg.error
                 if (
                     registration_returncode == 67 and
-                    'registration code' in error_message.lower()
+                    (
+                        'registration code' in error_message.lower() or
+                        'system credentials' in error_message.lower()
+                    )
```
schaefi, line 98: I would change this code part to be more future prepared to something like this 

```python
error_indicators = [
    'registration code',
    'system credentials'
]

if (registration_returncode == 67 and error_message.lower() in error_indicators):
    ...
```

It's imho a nicer read, will help you with writing a test and is an easy change if there will be more to check from the error data

Just an idea, thus not relevant for my approval

schaefi:
An entry to the .changes file is missing as well

usr/sbin/registercloudguest
rjschwei: That would not work because the error message will be some longer string and thus it would never match, the `in` test must be from the keywords we are looking for to the error message we receive from the registration command.

rjschwei:
> An entry to the .changes file is missing as well

Didn't want to create a conflict with your changes from #222. If this lands first I'll add something to your PR.

usr/sbin/registercloudguest
schaefi: ok this makes sense. Sorry I did not took into account that the complete error message looks differently.

schaefi:
> > An entry to the .changes file is missing as well
> 
> Didn't want to create a conflict with your changes from #222. If this lands first I'll add something to your PR.

Thanks much appreciated and mine is merged now

schaefi (approved):

tests/test_registercloudguest.py
```diff
@@ -2370,6 +2370,43 @@ def test_register_modules(mock_register_product):
...
+@patch('cloudregister.registerutils.register_product')
+def test_register_modules_failed_credentials(mock_register_product):
+    prod_reg_type = namedtuple(
+        'prod_reg_type', ['returncode', 'output', 'error']
+    )
+
+    mock_register_product.return_value = prod_reg_type(
+        returncode=67,
+        output='',
+        error='missing system credentials try again'
```
schaefi, line 2382: This is just nitpick: It would be good to use the real full error message as it came from a real call of the command for the test. I might be wrong but the above seems like an artificial error message produced to match our code

usr/sbin/registercloudguest
schaefi, line 98: This is just nitpick: We could still change this code to be more pythonic as

```python
error_indicators = [
    'registration code',
    'system credentials'
]

...

if any(error_message.lower() in issue for issue in error_indicators):
    ...
```

tests/test_registercloudguest.py
rjschwei: Yes it's fabricated but we cannot track, at least I am not aware of a mechanism, when the real error message ( https://github.com/SUSE/connect-ng/pull/272/files#diff-5018489c1a47a658436f1cc583e09b0289c1c1416d0f80e77620da54c3742239R350) changes, as such I am not convinced there is an advantage of having the real message in the test.

usr/sbin/registercloudguest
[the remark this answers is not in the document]
rjschwei: If it gets any longer we should reconsider this.

PR #224, opened 2024-11-26
Update patch files to be clean again

schaefi:
Fix patch files such that they apply clean and without producing a .orig file. Anytime a patch gets applied unclean it leaves a .orig file and the package build fails because of a new and unpackaged file. This commit fixes the patch files to apply clean again and also makes them git compliant


rjschwei (approved):
Thanks, forgot the follow up after I did the last push to SLE 12.

jesusbv (approved):

PR #225, opened 2024-12-02
Add registration code when activating extensions

jesusbv:
When registering a BYOS SUMA instance, only SLE Micro repos are present and SUMA repos are not because of the lack of registration code

This Fixes bsc#1234050

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -57,7 +57,7 @@ registration_returncode = 0
 # ----------------------------------------------------------------------------
 def register_modules(
     extensions, products, registration_target, instance_data_filepath,
-    registered=[], failed=[]
+    regcode, registered=[], failed=[]
 ):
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -57,7 +57,7 @@ registration_returncode = 0
 # ----------------------------------------------------------------------------
 def register_modules(
     extensions, products, registration_target, instance_data_filepath,
-    registered=[], failed=[]
+    regcode, registered=[], failed=[]
```
schaefi, line 60: The default value for `regcode` in `register_product` is set to `regcode=''`. As this code just passes along the regcode to another method I recommend to use the same default value

```python
regcode=''
```

for consistency reasons

jesusbv: Done

rjschwei (approved):

PR #227, opened 2024-12-05
Fix build with applied patches

schaefi:
The %patch macro does not support to call patch such that it does not create .orig files. Under certain conditions patch creates them and this will break the build for files found but not packaged. From my perspective a successfully applied patch even with an offset drift is still acceptable and doesn't have to break the package build process. The way how we maintain the patches for SLE12 will always provide the offset drift as we are constantly working on other parts of the code. If the patch applies is checked via a github action per PR. Thus I believe the proposed patch should be ok.

Find an example build with this change applied here:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@
 .cache
 **/*/__pycache__/
 .pytest_cache
+
+dist
+*.tar.bz2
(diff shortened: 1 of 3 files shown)


jesusbv (approved):

schaefi:
we observed issues with incomplete submissions such that files were missing or outdated. To prevent this in the future I also added a commit such that preparing the package files for a submission can be done with

```
make package

cp dist/* /to/some/submission/branch
cd  /to/some/submission/branch
osc -ci
```

then

* make sure the package builds for ALL targets in /to/some/submission/branch
* only if it builds run osc sr to the needed targets

Thanks

PR #228, opened 2024-12-05
No need to for a versioned tarball

schaefi:
The way the source tarball is created produces a versioned source tree inside of the tarball. There is no need to duplicate this information in the file name of the tarfile. Providing a consistent source file name prevents issues with package maintenance in a way that on each version bump a new source file needs to be deleted/added and also prevents issues with multiple source files still available if the packager forgot to delete the old one. Also spec file inconsistencies if only a new source tarball gets committed but the spec and the old source was forgotten to be updated/removed will no longer occur.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ package: tar
 	git pull
 	mkdir -p dist
 	rm -f dist/*
+	mv cloud-regionsrv-client.tar.bz2 dist/
 	cp cloud-regionsrv-client* dist/
 	cp *.patch dist/
 	@echo "Find package files for submission below dist/"
@@ -25,7 +26,7 @@ tar:
 	find "$(nv)" -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete
 	find "$(nv)" -path "*/lib/cloudregister.egg-info/*" -delete
 	find "$(nv)" -type d -name "cloudregister.egg-info" -delete
-	tar -cjf "$(nv).tar.bz2" "$(nv)"
+	tar -cjf "cloud-regionsrv-client.tar.bz2" "$(nv)"
 	rm -rf "$(nv)"
 
 install:
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -30,7 +30,7 @@ Summary:        Cloud Environment Guest Registration
 License:        LGPL-3.0-only
 Group:          Productivity/Networking/Web/Servers
 URL:            http://www.github.com/SUSE-Enceladus/cloud-regionsrv-client
-Source0:        %{name}-%{version}.tar.bz2
+Source0:        %{name}.tar.bz2
 # PATCH-FIX-SLES12 bsc#1203382 fix-for-sles12-disable-ipv6.patch
 Patch0:         fix-for-sles12-disable-ipv6.patch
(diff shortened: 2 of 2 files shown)


schaefi:
Further enhancement on how to maintain the package without the risk for issues as we saw them. With this one in place the package data stays consistent and can be produced via

```
make package
```

The resulting data is complete and consistent in its names below `dist/` and can be taken as one blob of files to be placed into a submission branch. No further `osc rm`, `osc add` which are all source for trouble if not done correctly will happen anymore

All this is just inspired by recent issues that we had. We are all under high load and processes that requires us to think about a number of details are source for errors especially in stressful situations. Please let us fix them and agree on a clear and easy path to produce packaging data.

Thanks

schaefi:
For reference a package building with this patch applied can be found here:

rjschwei:
I very much prefer a versioned tarball as it requires deliberate action when the package gets updated in the build service, i.e. `add` and `remove` action when the package is getting updated and the version is immediately obvious when looking at the package directory rather than having to open the spec file.

schaefi:
> I very much prefer a versioned tarball as it requires deliberate action when the package gets updated in the build service, i.e. `add` and `remove` action when the package is getting updated and the version is immediately obvious when looking at the package directory rather than having to open the spec file.

well as I tried to explain this opens a bunch of possible inconsistencies and the risk to submit something broken. The action that you prefer to perform is a source for trouble and imho ok as long as you are the only one maintaining the package. However there are now at least three people who are in charge of maintenance and just recently we saw all the consequences for potential breakage. I'm sorry I strongly disagree with the judgement that opening a file to find the package version is such a burden compared to a prone to errors process that can easily lead to broken submissions as it happened just yesterday.

From my perspective there two solutions to this problem

1. Only one guy maintains the package and sets the stage, single point of failure, wrong solution in my eyes but fine with me
2. If there are more  people involved the package submission process should be as conflict free and straight forward as possible. Less commands to call prior `osc sr` are better than more of them

I'm working towards option 2. and have a hard time to understand your objections. Where is the difference between

```
ls -l cloud-regionsrv-client/*.tar.bz2
```

and

```
tar -tf cloud-regionsrv-client.tar.bz2
```

to spot the version ?

jgleissner:
FWIW, I too prefer versionized tarballs. You can immediately see the version in OBS without having to open a file. It's not a lot of hassle, sure, but still. Also, versionized tarballs is pretty much the standard. And yes, it's relatively easy to forget to run `osc ar` before `osc commit`, but osc shows you the changes in the commit comments and since the package will fail to build if you commit anyway, you should notice very quickly. If you open a submit request without even checking whether the package builds you're doing something wrong IMHO.

schaefi:
There is no standard how to name your source tarball and if you get moved into the git workflow, or look into other distributions source workflow you either see an automation for handling the sources which is good or you see a concept that does not offer a prone to errors process which then gets eventually catch by automated sanity and verification checks. As I said I find it very cumbersome to be forced to `osc rm` followed by `osc add` an information that is also duplicated throughout the source tarball itself. This is just cumbersome and error prone and could be easily avoided.

> If you open a submit request without even checking whether the package builds you're doing something wrong IMHO.

Correct and it happened to Jesus because it's also easy to overlook

Every possible error condition will happen sooner or later. This PR exists to eliminate just a few of them. I cannot understand why there is such a resistance against less work but there you go, it's not my package ;)

PR #230, opened 2024-12-31
Refactor register base product

jesusbv:
Move the base product registration to its own method

Easier to read, test and modify

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -243,6 +243,76 @@ def setup_ltss_registration(
 
 
 # ----------------------------------------------------------------------------
+def register_base_product(
+    registration_target, instance_data_filepath, args, region_smt_servers
+):
+    base_registered = False
+    failed_smts = []
+    while not base_registered:
+        utils.add_hosts_entry(registration_target)
+        prod_reg = utils.register_product(
+            registration_target=registration_target,
+            email=args.email,
+            regcode=args.reg_code,
+            instance_data_filepath=instance_data_filepath
+        )
+        if prod_reg.returncode:
+            registration_returncode = prod_reg.returncode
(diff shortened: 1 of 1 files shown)


usr/sbin/registercloudguest
```diff
@@ -242,6 +242,76 @@ def setup_ltss_registration(
...
+                        error_msg % (
+                            str((target_smt_ipv4, target_smt_ipv6)),
+                            str((new_smt_ipv4, new_smt_ipv6))
+                        )
+                    )
+                    utils.remove_registration_data()
+                    utils.clean_hosts_file(
+                        registration_target.get_domain_name()
+                    )
+                    registration_target = smt_srv
```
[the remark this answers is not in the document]
jesusbv: fixed in https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/230/commits/8e37550971917917a80bc5262baec1958dfca69c

usr/sbin/registercloudguest
```diff
@@ -454,7 +527,6 @@ argparse.add_argument(
 
 
 def main(args):
-    global registration_returncode
```
schaefi, line 457: you dropped the scope of registration_returncode to be global. Can we be sure this is not causing an issue elsewhere in the code ? In general I like this change but I can't foresee its after effects

jesusbv: I tracked the variable and I would like to remove the global one but for now, I kept it

the idea behind this refactor exercise is to isolate scopes and make them independent,
I do not want to introduce a regression, so let's keep it for this release and I will work in removing
it for the next

it is enough refactoring for one release

schaefi: sounds good, thanks

schaefi (approved):
looks good :+1:

rjschwei (approved):

PR #234, opened 2024-12-31
Refactor find registration target

jesusbv:
Move the finding of an alive registration target to its own method

Easier to read, test and modify


rjschwei (changes requested):
This change does not work and probably has not been tested or the failure mode should have been rather obvious.

In the existing code the variable `registration_smt`  gets reset to a new target server [1] if the current registration target is no reachable. The new target server is then used for other setup tasks[2].

In the proposed refactor `registration_smt`  becomes a variable that is local to the new function and as such the reassignment [3] is meaningless for the rest of the code after the new function has been called and returned.

[1] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L530
[2] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L554
[3] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/234/files#diff-64591b8c320243a33351830f696fb2145651800c0ece404826e022d9b55ba630R296

jesusbv:
> This change does not work and probably has not been tested or the failure mode should have been rather obvious.
> 
> In the existing code the variable `registration_smt` gets reset to a new target server [1] if the current registration target is no reachable. The new target server is then used for other setup tasks[2].
> 
> In the proposed refactor `registration_smt` becomes a variable that is local to the new function and as such the reassignment [3] is meaningless for the rest of the code after the new function has been called and returned.
> 
> [1] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L530 [2] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/usr/sbin/registercloudguest#L554 [3] https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/234/files#diff-64591b8c320243a33351830f696fb2145651800c0ece404826e022d9b55ba630R296

Good catch
Updated the returned values

The tests would also need an update but that would be after the refactor PRs are merged

rjschwei:
> Good catch Updated the returned values
> 
> The tests would also need an update but that would be after the refactor PRs are merged

Testing is not just unit tests. When PRs get submitted they are expected to have been tested before hand. This is why the integration test process was documented such that it is not left up to chance that all the test cases are being remembered. Following the documented integration test process would, I think, have shown this problem, because the integration test process has a section that tests failover. At the very least the integration test process should have provided the guideposts to provide ideas for improvement of the integration test.

usr/sbin/registercloudguest
```diff
@@ -577,63 +644,9 @@ def main(args):
...
-                registration_smt = new_target
-                registration_target_found = True
-                break
-            else:
-                msg = 'Configured update server is unresponsive. '
-                msg += 'Could not find a replacement update server '
-                msg += 'in this region. Possible network configuration issue'
-                logging.error(msg)
-                sys.exit(1)
+    registration_smt, reg_target_found = find_alive_registration_target(
```
schaefi, line 647: Can we keep the variable naming consistent and stick with `registration_target_found` instead of reg_target_found

it will also make the patch smaller because the subsequent `if registration_target_found` will then stay unchanged.

Thanks

schaefi (approved):
nice :+1:

PR #237, opened 2025-03-26
Remove azure-addon package

jesusbv:
The azure-addon package handles the AHB (license change for Azure), we are moving the license change handling to a new/different package outside cloud-regionsrv-client to a more generic solution, CSP agnostic

This change cleans up the azure-addon package from cloud-regionsrv-client


rjschwei:
This is way pre-mature and this change should come at the same time when we figure out how to switch the running service to the new service. I have some information about this, please don't start your own research, but have not gotten back to sorting it all out.

PR #238, opened 2025-03-28
Clean up when package gets removed

rjschwei:
At present when the package gets removed we leave repositories behind that point to the plugin that is supplied with this package. However with the removal of this package the plugin is also removed. This leaves the system in a state that is difficult to recover from and requires manual clean up. We clean up the registration when the package is removed to avoid this problem.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -210,6 +210,15 @@ gzip %{buildroot}/%{_mandir}/man1/*
 %pre addon-azure
 %service_add_pre regionsrv-enabler.timer
 
+%preun
+# Do not run during an upgrade process
+# When the package is removed we need to clean up or we will leave
+# repositories with "plugin://" behind while the plugin we supply is
+# being removed (bsc#1240310)
+if [ "$1" -eq 0 ]; then
+    %{_sbindir}/registercloudguest --clean
+fi
+
 %post
 # Scripts need access to the update infrastructure, do not execute them
 # in the build service.
--- a/cloud-regionsrv-client.changes
+++ b/cloud-regionsrv-client.changes
@@ -1,4 +1,12 @@
 -------------------------------------------------------------------
+NEXT version
+
+- Update version to
+  + Remove repositories when the package is being removed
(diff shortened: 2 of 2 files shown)


cloud-regionsrv-client.spec
```diff
@@ -210,6 +210,15 @@ gzip %{buildroot}/%{_mandir}/man1/*
 %pre addon-azure
 %service_add_pre regionsrv-enabler.timer
 
+%preun
+# Do not run during an upgrade process
+# WHen the package is removed we need to clean up or we will leave
```
schaefi, line 215: typo `WHen`

cloud-regionsrv-client.spec
```diff
@@ -210,6 +210,15 @@ gzip %{buildroot}/%{_mandir}/man1/*
...
+# repositories with "plugin://" behind while the plugin we supply is
+# being removed (bsc#1240310)
+if [ $1 -eq 0 ]; then
```
schaefi, line 218: -- SC2086 (info): Double quote to prevent globbing and word splitting.

```
if [ "$1" -eq 0 ]; then
```

cloud-regionsrv-client.spec
```diff
@@ -210,6 +210,15 @@ gzip %{buildroot}/%{_mandir}/man1/*
 %pre addon-azure
 %service_add_pre regionsrv-enabler.timer
 
+%preun
+# Do not run during an upgrade process
```
schaefi, line 214: The comment information that this script code does not run on a package upgrade process should be imho added directly in front of the if condition. It would be even better to give the $1 more meaning by storing the value in a more self explaining variable. In that case the comment can go away completely because reading the code will then be enough. But I leave this part to you.

schaefi (approved):
I'm fine with the concept idea, just some nitpick on the implementation. Feel free to adapt as you see fit

cloud-regionsrv-client.spec
rjschwei: Thanks fixed

bear454 (approved):

jesusbv (approved):

PR #239, opened 2025-03-28
Turn docker into an optional setup

schaefi:
The simple docker requirement in the spec file of the cloud-regionsrv-client package has turned into a big problem for customers who don't want docker on their systems. This commit makes docker a recommended package and refactors the registration setup code in a way that docker becomes an optional part of the setup procedure. With each call of registercloudguest a check is made to see if docker is installed. If it is installed the current setup is updated with a docker related configuration if not already done

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -145,7 +145,41 @@ def setup_registry(registration_target, clean='registry'):
         utils.get_credentials_file(registration_target)
     )
     registry_fqdn = registration_target.get_registry_FQDN()
+
+    with_docker = utils.is_docker_present()
+    if not with_docker:
+        logging.info(
+            'docker config file {} not found. '
+            'Registration for docker skipped. '
+            'Install docker to add docker support.'.format(
+                utils.DOCKER_CONFIG_PATH
+            )
+        )
+
     if utils.is_registry_registered(registry_fqdn):
+        if with_docker:
+            # Special handling for the docker case:
+            #
+            # When docker is present we perform the registry configuration
+            # for the docker container engine. This may cause calling the
+            # setup routine even though docker was setup before. However,
+            # the code will only effectively write a new registry setup
+            # if the required modifications are not included.
+            logging.info('Adding docker support')
+            docker_setup_ok = utils.set_registries_conf_docker(registry_fqdn)
+
+            # docker_setup_ok will take a True/False value on a real
+            # action that happened. In any other case a None value is
+            # present which indicates no action was performed.
+            if docker_setup_ok is False:
+                print(
+                    'Registration failed(docker), see {0} for details'.format(
+                        LOG_FILE
+                    ), file=sys.stderr
+                )
+                sys.exit(1)
+            elif docker_setup_ok is True:
+                print('Successfully added docker to the registry setup')
         logging.info(
             'Instance is setup to access container registry, nothing to do'
         )
@@ -154,7 +188,16 @@ def setup_registry(registration_target, clean='registry'):
         'Adding registry setup to instance, all existing shell '
         'sessions must be restarted to use the registry.'
     )
-    if not utils.setup_registry(registry_fqdn, user, password):
+    registry_setup_ok = False
+    if utils.prepare_registry_setup(registry_fqdn, user, password):
+        logging.info('Adding podman support')
+        registry_setup_ok = utils.set_registries_conf_podman(registry_fqdn)
+        if with_docker:
+            logging.info('Adding docker support')
+            registry_setup_ok = utils.set_registries_conf_docker(registry_fqdn)
+        if utils.is_suma_instance():
+            registry_setup_ok = utils.set_registry_fqdn_suma(registry_fqdn)
+    if not registry_setup_ok:
         if clean == 'all':
             cleanup()
         elif clean == 'registry':
--- a/cloud-regionsrv-client.changes
+++ b/cloud-regionsrv-client.changes
@@ -5,7 +5,9 @@ NEXT version
   + Remove repositories when the package is being removed
     We do not want to leave repositories behind refering to the plugin that
     is being removed when the package gets removed (bsc#1240310, bsc#1240311)
-
+  + Turn docker into an optional setup
+    Change the Requires into a Recommends and adapt the code accordingly
+    
 -------------------------------------------------------------------
 Tue Dec  3 17:21:52 UTC 2024 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
 
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -58,10 +58,10 @@ Requires:       %{pythons}-toml
 # podman related config files gets pulled in. We modify
 # /etc/containers/registries.conf
 Requires:       libcontainers-common
-# Add requirement for docker to make sure all docker related
-# config files gets pulled in. We modify
+# Add recommendation for docker to make sure all docker related
+# config files get pulled in. If present we modify
 # /etc/docker/daemon.json
-Requires:       docker
+Recommends:     docker
 %endif
 Requires:       regionsrv-certs
 Requires:       sudo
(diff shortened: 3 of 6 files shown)


schaefi:
This is a first shot of the ideas that we have discussed. In our current implementation we configured all container engines in one function because it was the requirement in the beginning. Now docker becomes an optional setup which of course lead to some refactoring on many places. I tried to code it as less intrusive as possible but as you can see it was not easy. Especially the unit testing is relatively hard to adapt. Please review carefully, thanks

schaefi:
@jesusbv can you please do me a favor and debug why in Gods name the coverage is not met on python 3.6. I have no clue and there is an explicit test the code path in question available at: `tests/test_registercloudguest.py::test_setup_registry_ok` Thanks much

usr/sbin/registercloudguest
```diff
@@ -143,8 +143,38 @@ def setup_registry(registration_target, clean='registry'):
     user, password = utils.get_credentials(
         utils.get_credentials_file(registration_target)
     )
+    with_docker = True
+    if not utils.is_docker_present():
+        with_docker = False
+        logging.info(
+            'Registration for docker registry skipped, '
+            'docker is not installed.'
```
jesusbv, line 151: do we know that ? we know the file is not present

can it be that docker is installed and the user removed that file ?

usr/sbin/registercloudguest
```diff
@@ -143,8 +143,38 @@ def setup_registry(registration_target, clean='registry'):
...
+            # Special handling for the docker case:
+            #
+            # When docker is present we perform the registry configuration
+            # for the docker container engine. This may cause calling the
+            # setup routine even though docker was setup before. However,
+            # the code will only effectively write a new registry setup
+            # if the required modifications are not included.
+            docker_setup_ok = utils.set_registries_conf_docker(registry_fqdn)
+
+            # registry_setup_ok will take a True/False value on a real
```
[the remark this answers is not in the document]
schaefi: yes, thanks and fixed

usr/sbin/registercloudguest
schaefi: The message will be the hardest part :) I tried to adapt it

schaefi:
@jesusbv I found the test coverage issue myself and fixed it

schaefi:
This change needs extensive testing to avoid regressions. Integration testing should cover

- [ ] podman and docker installed, both engines setup as before
- [ ] none of podman and docker installed, only a podman config is produced with our registry information. A followup fix to the code should be made to write the podman config information as extra file and not into the main config
- [ ] only podman is installed, expected podman setup and skip docker
- [ ] docker gets installed later and registercloudguest is called again, docker setup gets added
- [ ] cloud-regionsrv-client gets uninstalled with docker and podman installed, both configs must be cleared
- [ ] cloud-regionsrv-client gets uninstalled without any docker and podman present, no cleanup, no error
- [ ] cloud-regionsrv-client gets uninstalled with only podman installed, podman setup gets cleared
- [ ] cloud-regionsrv-client gets uninstalled with only docker installed, docker setup gets cleared

I'm surely missing a case, please help me in completing the list above. Thanks

jesusbv:
> This change needs extensive testing to avoid regressions. Integration testing should cover
> 
>     * [ ]  podman and docker installed, both engines setup as before
> 
>     * [ ]  none of podman and docker installed, only a podman config is produced with our registry information. A followup fix to the code should be made to write the podman config information as extra file and not into the main config
> 
>     * [ ]  only podman is installed, expected podman setup and skip docker
> 
>     * [ ]  docker gets installed later and registercloudguest is called again, docker setup gets added
> 
>     * [ ]  cloud-regionsrv-client gets uninstalled with docker and podman installed, both configs must be cleared
> 
>     * [ ]  cloud-regionsrv-client gets uninstalled without any docker and podman present, no cleanup, no error
> 
>     * [ ]  cloud-regionsrv-client gets uninstalled with only podman installed, podman setup gets cleared
> 
>     * [ ]  cloud-regionsrv-client gets uninstalled with only docker installed, docker setup gets cleared
> 
> 
> I'm surely missing a case, please help me in completing the list above. Thanks


Not sure a case we should cover but

*[ ] Docker is installed, cloud-regionsrv-client configures Docker (and Podman, not important) 
      Docker gets uninstalled, re run of cloud-regionsrv-client/cloud-regionsrv-client should remove

schaefi:
> *[ ] Docker is installed, cloud-regionsrv-client configures Docker (and Podman, not important) Docker gets uninstalled, re run of cloud-regionsrv-client/cloud-regionsrv-client should remove

I don't understand it, should remove ... what ? If docker is installed, podman is installed too (because we require it). That would be case Nr 1. If you uninstall docker and run the client again that would be case Nr 3

schaefi:
Regarding

> A followup fix to the code should be made to write the podman config information as extra file and not into the main config

This can't be done because the main config file is always read first. The search setup there would then take over precedence to any other custom setup. However, we want our pubcloud search entry to be the very first in the list (our code does that) and if we move to a custom file in /etc/containers/registries.conf.d/... we will never be first. Thus we have to stick with the solution we have

usr/sbin/registercloudguest
```diff
@@ -144,7 +144,43 @@ def setup_registry(registration_target, clean='registry'):
         utils.get_credentials_file(registration_target)
     )
     registry_fqdn = registration_target.get_registry_FQDN()
+
+    with_docker = True
+    if not utils.is_docker_present():
+        with_docker = False
+        utils.clean_registries_conf_docker(registry_fqdn)
```
rjschwei, line 151: We determined that the config file is not present, why call clean? That only triggers an additional message to the logfile """'Unable to read "{0}", cleanup not possible'.format(DOCKER_CONFIG_PATH)""".
If there's no docker then we only write the log message. Or am I missing something?

usr/sbin/registercloudguest
```diff
@@ -144,7 +144,43 @@ def setup_registry(registration_target, clean='registry'):
...
+            #
+            # When docker is present we perform the registry configuration
+            # for the docker container engine. This may cause calling the
+            # setup routine even though docker was setup before. However,
+            # the code will only effectively write a new registry setup
+            # if the required modifications are not included.
+            logging.info('Adding docker support')
+            docker_setup_ok = utils.set_registries_conf_docker(registry_fqdn)
+
+            # docker_setup_ok will take a True/False value on a real
```
rjschwei, line 172: The comment is confusing. Unless I missed something it appears that `set_registries_conf_docker` does not return `True` when the config was setup successfully, it returns `None`

usr/sbin/registercloudguest
schaefi: yes this doesn't make any sense, thanks for spotting it. I tricked myself here. The idea was to cleanup a left over config setup when docker was uninstalled. But we detect that state by checking for the config file. I also realized when you uninstall docker the config gets moved as .rpmsave. So yes the cleanup here is stupid. Sorry I'm going to drop this

schaefi: done

usr/sbin/registercloudguest
```diff
@@ -144,7 +144,43 @@ def setup_registry(registration_target, clean='registry'):
         utils.get_credentials_file(registration_target)
     )
     registry_fqdn = registration_target.get_registry_FQDN()
+
+    with_docker = True
+    if not utils.is_docker_present():
+        with_docker = False
```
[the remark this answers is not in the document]
schaefi: done

usr/sbin/registercloudguest
schaefi: set_registries_conf_docker has the following states

1. no docker config present: `None`
2. docker config present but failed to read: `False`
3. docker config present, read in and modified_by_us stays false: `None`
4. docker config present, read in and modified_by_us becomes true. This causes the call of `return write_registries_conf`. This method logs "File %s updated" and returns True: `True`
5. docker config present, read in and modified_by_us becomes true and  write_registries_conf raises: `False`

That's how I read the code

schaefi:
I started running the integration tests from the checkbox list I sent. I will check off the tests that went fine

usr/sbin/registercloudguest
rjschwei: I missed the `return` statement in front of the call to `write_registries_conf()`

rjschwei (approved):

schaefi:
I finished my tests, so far this looks good

PR #240, opened 2025-03-31
Package and compat fixes

schaefi:
This patch is two fold

**Fixed urllib3 import to be compatible**
    
Simply importing urllib3 is incompatible with the python version on SLE12. The same trick as in other parts of the source is applied which imports urllib3 from requests

**Fixed spec file %preun section**
    
Latest spec change broke the spec file because there was already a %preun section specified. Adding another one for the same package is not supported and failed on rpm parsing. This commit fixes the spec file and also fixes the invocation for registercloudguest to be done only in an environment  providing a client configuration which is mandatory for the  tool to run. If no client config exists the package build also fails because rpmbuild invokes the scripts at build  time.

### DIFF
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -35,7 +35,6 @@ import requests
 import sys
 import time
 import urllib.parse
-import urllib3
 import uuid
 
 import cloudregister.registerutils as utils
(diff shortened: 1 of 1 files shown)


schaefi:
Package building here:

jesusbv (approved):

PR #241, opened 2025-03-31
Generalize flavor transition

rjschwei:
We currently support Azure instances in a transition from PAYG to BYOS and vice-versa. This transition is going to be supported in GCE as well. Therefore create an implementation that is not framework specific and utilizes the flavor determination implemented with the instance-billing-flavor-check.

### DIFF
--- a/usr/sbin/cloudguest-lic-watcher
+++ b/usr/sbin/cloudguest-lic-watcher
@@ -1,0 +1,63 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2025, SUSE LLC, All rights reserved.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3.0 of the License, or (at your option) any later version.
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.
+
+import logging
+import os
+
+import instance_billing_flavor_check.utils as inst_flvr_utils
+import cloudregister.registerutils as utils
+
+CACHE_LICENSE_PATH = os.path.join(utils.get_state_dir(), 'cached_license')
+
+def update_license_cache(license_type):
+    """Update the cache to track the license type"""
+    with open(CACHE_LICENSE_PATH, 'w') as cached_license:
+        cached_license.write(license_type)
+
+def has_license_changed(license_type):
+    """Check license type changes against metadata info"""
+    if os.path.exists(CACHE_LICENSE_PATH):
+        with open(CACHE_LICENSE_PATH, 'r') as cached_license:
+            old_license = cached_license.read()
+            return license_type != old_license
+    else:
+        update_license_cache(license_type)
+
+
+utils.start_logging()
+service_name = 'guestregister.service'
+current_flavor = inst_flvr_utils.check_payg_byos()[0]
+if has_license_changed(current_flavor):
+    if current_flavor == 'BYOS' and not utils.uses_rmt_as_scc_proxy():
+        logging.info('Detected flavor change to BYOS, clean up registration')
+        utils.clean_all()
+        utils.exec_subprocess(['systemctl', 'disable', service_name])
+    if current_flavor == 'PAYG':
+        base_msg = 'Detected flavor change to PAYG, {status}'
+        current_target = utils.get_current_smt()
+        if (
+                not current_target or not
+                utils.is_registered(current_target.get_FQDN())
+        ):
+                logging.info(base_msg.format(status='registering'))
+                utils.exec_subprocess(['registercloudguest'])
+                utils.exec_subprocess(['systemctl', 'enable', service_name])
+        logging.info(base_msg.format(
+            status='already registered, nothing to do')
+        )
+    update_license_cache(current_flavor)
+        
+        
+    
--- a/usr/sbin/registercloudguest
+++ b/usr/sbin/registercloudguest
@@ -121,16 +121,7 @@ def register_modules(
 # ----------------------------------------------------------------------------
 def cleanup():
     """Remove any registration artifacts"""
-    try:
-        utils.clean_non_free_extensions()
-    except Exception as err:
-        print('Could not check the system product data: {}'.format(err))
-
-    utils.clean_registry_setup()
-    utils.remove_registration_data()
-    utils.clean_smt_cache()
-    utils.clear_new_registration_flag()
-    utils.clean_framework_identifier()
+    utils.clean_all()
 
 
 # ----------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+
+CACHE_LICENSE_PATH = os.path.join(utils.get_state_dir(), 'cached_license')
+
+def update_license_cache(license_type):
+    """Update the cache to track the license type"""
+    with open(CACHE_LICENSE_PATH, 'w+') as cached_license:
+        cached_license.write(license_type)
+
+def has_license_changed(license_type):
+    """Check license type changes against metadata info"""
```
jesusbv, line 30: should we update the docstring to include `create cache if cache is not present` ?

jesusbv (changes requested):
other than the changes that should be in a different PR

LGTM

usr/sbin/cloudguest-lic-watcher
rjschwei: Yes there is a side effect in this implementation in the exception clause. But the "creates file if necessary" would be better placed in the write method. I do not consider it necessary in this case.

usr/sbin/registercloudguest
```diff
@@ -120,16 +120,7 @@ def register_modules(
 # ----------------------------------------------------------------------------
 def cleanup():
     """Remove any registration artifacts"""
-    try:
```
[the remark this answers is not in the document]
rjschwei: It is a reasonably small change that is straight forward. As it is a simplification that reduces complexity in this  implementation I decided to keep it in one unit and I will not take this PR apart.

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+import os
+
+import instance_billing_flavor_check.utils as inst_flvr_utils
+import cloudregister.registerutils as utils
+
+CACHE_LICENSE_PATH = os.path.join(utils.get_state_dir(), 'cached_license')
+
+def update_license_cache(license_type):
+    """Update the cache to track the license type"""
+    with open(CACHE_LICENSE_PATH, 'w+') as cached_license:
```
schaefi, line 26: why "w+" open for writing and reading when you only write to the file ?

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+        cached_license.write(license_type)
+
+def has_license_changed(license_type):
```
schaefi, line 29: is this code expected to run on SLE12 ? If not and we know it will be python >= 3.6 I vote for adding type annotations 

```python
def has_license_changed(license_type: str) -> bool:
```

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+    """Check license type changes against metadata info"""
+    try:
+        with open(CACHE_LICENSE_PATH, 'r') as cached_license:
+            old_license = cached_license.read()
+            return license_type != old_license
+    except FileNotFoundError:
```
schaefi, line 35: try / except for FileNotFoundError. Any particular reason why you can't say `if os.path.isfile(...):`

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+def update_license_cache(license_type):
+    """Update the cache to track the license type"""
+    with open(CACHE_LICENSE_PATH, 'w+') as cached_license:
+        cached_license.write(license_type)
+
+def has_license_changed(license_type):
+    """Check license type changes against metadata info"""
+    try:
+        with open(CACHE_LICENSE_PATH, 'r') as cached_license:
+            old_license = cached_license.read()
```
schaefi, line 33: no lstrip() needed ? no CR or other space characters that could harm the subsequent comparison expected ?

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+            return license_type != old_license
+    except FileNotFoundError:
+        update_license_cache(license_type)
+
+
+utils.start_logging()
+service_name = 'guestregister.service'
+current_flavor = inst_flvr_utils.check_payg_byos()[0]
```
schaefi, line 41: is check_payg_byos() safe to return a List in any case ? If not this code is prone to a raise condition for index access of e.g None type

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+        utils.clean_all()
+        utils.exec_subprocess(['systemctl', 'disable', service_name])
+    if current_flavor == 'PAYG':
+        base_msg = 'Detected flavor change to PAYG, %s'
+        current_target = utils.get_current_smt()
+        if (
+                not current_target or not
+                utils.is_registered(current_target.get_FQDN())
+        ):
+                logging.info(base_msg %'registering')
```
schaefi, line 54: nested logging with format sequences should be avoided and can easily lead to "TypeError: not all arguments converted during string formatting" especially when not covered by unit tests. I suggest to change this to

```python
base_msg = 'Detected flavor change to PAYG, {status}'

logging.info(base_msg.format(status='registering'))
```

usr/sbin/cloudguest-lic-watcher
```diff
@@ -0,0 +1,61 @@
...
+                utils.exec_subprocess(['registercloudguest'])
+                utils.exec_subprocess(['systemctl', 'enable', service_name])
+        logging.info(base_msg %'already registered, nothing to do')
```
schaefi, line 57: dito

schaefi (changes requested):

usr/sbin/cloudguest-lic-watcher
rjschwei: Yep, fixed.

usr/sbin/cloudguest-lic-watcher
rjschwei: SLE 12 support is needed. No types.

usr/sbin/cloudguest-lic-watcher
rjschwei: Cut and paste laziness from existing code. Changed.

usr/sbin/cloudguest-lic-watcher
rjschwei: We control the write operation, I think we can do without normalizing what we read.

usr/sbin/cloudguest-lic-watcher
rjschwei: Yes, `check_payg_byos` returns the human readable flavor, i.e. "PAYG" or "BYOS" **and** a code indicting the confidence level of the determination.

usr/sbin/cloudguest-lic-watcher
rjschwei: Changed

usr/sbin/cloudguest-lic-watcher
rjschwei: Changed

jesusbv (approved):

PR #242, opened 2025-04-08
Switch license flavor handling in Azure

rjschwei:
With 198bdb7 a generic way to detect the instance billing flavor was introduced. This commit removes the Azure specific implementation of this feature in favor of the generic implementation. We drop the addon-azure package and the files this RPM package contained.


rjschwei:
Supersedes/Obsoletes #237 and #138

schaefi (approved):
This looks good to me :+1:

jesusbv (approved):

rjschwei:
> > Supersedes/Obsoletes #237 and #138
> 
> regarding #138 the issue just moved to a different place but it's still present

Please explain

rjschwei:
Agreed the issue attempted to be addressed by #138 is not solved. That said once this is merged, which I will do #138 is obsolete since it makes changes to removed code.

Expressing the issue that motivated #138 as an issue #243

PR #244, opened 2025-04-15
Handle state inconsistency

rjschwei:
It is possible that our state file disappears. In this case we need to still handle the state transition between billing flavors correctly. We do not want the user to have to mess withe registration even if our cache disappears.

### DIFF
--- a/usr/sbin/cloudguest-lic-watcher
+++ b/usr/sbin/cloudguest-lic-watcher
@@ -20,11 +20,13 @@ import instance_billing_flavor_check.utils as inst_flvr_utils
 import cloudregister.registerutils as utils
 
 CACHE_LICENSE_PATH = os.path.join(utils.get_state_dir(), 'cached_license')
+SERVICE_NAME = 'guestregister.service'
 
 def update_license_cache(license_type):
     """Update the cache to track the license type"""
     with open(CACHE_LICENSE_PATH, 'w') as cached_license:
         cached_license.write(license_type)
+
 
 def has_license_changed(license_type):
     """Check license type changes against metadata info"""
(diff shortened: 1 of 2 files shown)


rjschwei:
Test cases completed:
  - PAYG instance installed lic-watcher (no cache) present, this represents
    initial boot use case as well, run cloudguest-lic-watcher
    + Cache created
      ~ # cat /var/cache/cloudregister/cached_license
          PAYG
    + No changes to repos
      ~ zypper lr
        1 | Basesystem_Module_x86_64:
        .....
    + Result: Pass
  - Stop PAYG instance switch to BYOS, run cloudguest-lic-watcher
    + Cache updated
      ~ # cat /var/cache/cloudregister/cached_license
          BYOS
    + Repos removed
      ~ zypper lr
        1 | google-cloud
    + Result: Pass
  - Remove cache file, register with SCC, stop instance switch to PAYG, run cloudguest-lic-watcher
    + rm /var/cache/cloudregister/cached_license
    + SUSEConnect -r XXXXXXXXX
      Announcing system to https://scc.suse.com ...

      Activating SLES 15.6 x86_64 ...
      -> Adding service to system ...

    + zypper lr
      1 | Basesystem_Module_15_SP6_x86_64
      .....
    + Cache created
      ~ # cat /var/cache/cloudregister/cached_license
          PAYG
    + Repos switched to update infra
      ~ zypper lr
        1 | Basesystem_Module_x86_64:
      ~ grep susecloud 
        # grep susecloud /etc/zypp/services.d/*
/etc/zypp/services.d/Basesystem_Module_x86_64.service:url = plugin:/susecloud?c
    + Result: Pass
  - Remove cache file, stop instance switch to BYOS, run cloudguest-lic-watcher
    + rm /var/cache/cloudregister/cached_license
    + Cache created
      ~ # cat /var/cache/cloudregister/cached_license
          BYOS
    + Repos removed
      ~ zypper lr
        1 | google-cloud
    + Result: Pass
  - Stop instance switch to PAYG, run cloudguest-lic-watcher
    + Cache updated
      ~ # cat /var/cache/cloudregister/cached_license
          PAYG
    + Registered instance to the update infrastructure
      ~ zypper lr
        1 | Basesystem_Module_x86_64:
        .....
      ~ grep susecloud 
        # grep susecloud /etc/zypp/services.d/*
    + Result: Pass

usr/sbin/cloudguest-lic-watcher
```diff
@@ -34,29 +35,69 @@ def has_license_changed(license_type):
             return license_type != old_license
     else:
         update_license_cache(license_type)
+        maybe_drop_registration(license_type)
+        maybe_register_system(license_type)
```
jesusbv, line 39: those two lines are a bit weird, wouldn't it be better to have one method that decides 
whether to de-register or register depending on the condition than having two methods
"`may drop registration` hold on, `may register`" ?

something like
`update_registration_state(license_type)`
and call either of the two methods inside, depending on condition

also, it would be nice to rename the methods to `drop_registration_if_x` and `register_system_if_x`, for example

usr/sbin/cloudguest-lic-watcher
```diff
@@ -34,29 +35,69 @@ def has_license_changed(license_type):
...
-service_name = 'guestregister.service'
-current_flavor = inst_flvr_utils.check_payg_byos()[0]
-if has_license_changed(current_flavor):
-    if current_flavor == 'BYOS' and not utils.uses_rmt_as_scc_proxy():
-        logging.info('Detected flavor change to BYOS, clean up registration')
-        utils.clean_all()
-        utils.exec_subprocess(['systemctl', 'disable', service_name])
+def maybe_drop_registration(license_type):
+    """Clean up registration data if needed"""
+    if current_flavor == 'BYOS':
```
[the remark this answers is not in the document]
jesusbv: odd this error did not show up in the tests that passed

jesusbv (changes requested):
`current_flavor` does not seem to exist in that scope (inside `maybe_drop_registration` method), that would produce an error

usr/sbin/cloudguest-lic-watcher
```diff
@@ -34,29 +35,69 @@ def has_license_changed(license_type):
...
+            utils.exec_subprocess(['systemctl', 'disable', service_name])
+
+
+def maybe_register_system(license_type):
+    """Register the system if needed handle the folowing cases
+       - System is not registered at all
+       - System is registered as BYOS to update server
+       - System is registered to SCC
+    """
     if current_flavor == 'PAYG':
```
jesusbv, line 61: ditto

usr/sbin/cloudguest-lic-watcher
```diff
@@ -34,29 +35,69 @@ def has_license_changed(license_type):
...
+    """Clean up registration data if needed"""
+    if current_flavor == 'BYOS':
+        if not utils.get_current_smt():
+            # There is no target registration server, nothing to do
+            return
+        if not utils.uses_rmt_as_scc_proxy():
+            logging.info(
+                'Detected flavor change to BYOS, clean up registration')
+            utils.clean_all()
+            utils.exec_subprocess(['systemctl', 'disable', service_name])
```
jesusbv, line 52: same for `service_name`, it is set in line 96 but not in the call in line 38

usr/sbin/cloudguest-lic-watcher
```diff
@@ -34,29 +35,69 @@ def has_license_changed(license_type):
...
 
-utils.start_logging()
-service_name = 'guestregister.service'
-current_flavor = inst_flvr_utils.check_payg_byos()[0]
-if has_license_changed(current_flavor):
-    if current_flavor == 'BYOS' and not utils.uses_rmt_as_scc_proxy():
-        logging.info('Detected flavor change to BYOS, clean up registration')
-        utils.clean_all()
-        utils.exec_subprocess(['systemctl', 'disable', service_name])
+def maybe_drop_registration(license_type):
```
jesusbv, line 42:
```suggestion
def maybe_drop_registration(license_type, service_name='guestregister.service'):
```

usr/sbin/cloudguest-lic-watcher
rjschwei: The condition to drop or register is in the function implementation to avoid having to repeat the condition here and and in another part of the code. Since the outcome for the function is not deterministic, `maybe` is the appropriat ename.

usr/sbin/cloudguest-lic-watcher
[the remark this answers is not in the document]
rjschwei: The code was tested, unless I made a mistake when creating the package. Anyway, yes good catch this needs fixing. Thanks.

usr/sbin/cloudguest-lic-watcher
rjschwei: Created a constant

rjschwei:
Re-ran the tests from the original sequence. All worked as expected.

usr/sbin/cloudguest-lic-watcher
rjschwei: A function to call 2 other functions and do nothing else? A bit of an overkill to me. If it were a whole process yes, but for 2 I don't see the point.

jesusbv: https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L1097

we use that pattern with no issue before

rjschwei:
Also ran the tests on SLE 12 for sanity checking, all passed.

schaefi (approved):
LGTM

I approve only for the state of the code not for functionality as I have no idea how to properly integration test this part

PR #245, opened 2025-04-18
Properly handle the root argument for zypper

rjschwei:
When parsing the zypper command line to determine if we operate on a different root system we did not account for the space in the argument list to indicate the directory. This resulted in incorrect parsing results for packages with "-R" in the package name. Resulting in failed package updates.


jesusbv (approved):

schaefi (approved):
What a bug :facepalm:

PR #249, opened 2025-05-22
Update SUMA and MLM check

jesusbv:
Before MLM the SUMA product was named 'SUSE-Manager-Server.prod', 
now it is called 'Multi-Linux-Manager-Server.prod', we need to check both products

This fixes bsc#1243437

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -2314,12 +2314,12 @@ def is_suma_instance():
         'manager-server.prod',
         'sle-micro.prod'
     ]
-    num_prod_matches = [
+    prod_matches = [
         suma_prod for product in products
-        for suma_prod in suma_prods if suma_prod in Path(product.lower()).name
+        for suma_prod in suma_prods if suma_prod in product.lower()
     ]
 
-    return len(num_prod_matches) == len(suma_prods)
+    return len(prod_matches) == len(suma_prods)
 
 
 # ----------------------------------------------------------------------------
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -2308,13 +2308,13 @@ def is_suma_instance():
     """Return True if there is a match for SUMA and SLE-Micro products,
     otherwise False."""
     products = glob.glob('/etc/products.d/*.prod')
+    num_matches = 0
     # suse-manager-server changed to
     # multi-linux-manager-server, check both (bsc#1243437)
-    num_matches = 0
     suma_prods = [
-        '/etc/products.d/suse-manager-server.prod',
-        '/etc/products.d/multi-linux-manager-server.prod',  # SUMA 5.1
-        '/etc/products.d/sle-micro.prod',
+        '/etc/products.d/suse-manager-server.prod',  # SUMA 5.0
+        '/etc/products.d/multi-linux-manager-server.prod',  # SUMA 5.1 (MLM)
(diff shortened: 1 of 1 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -2308,16 +2308,18 @@ def is_suma_instance():
     """Return True if there is a match for SUMA and SLE-Micro products,
     otherwise False."""
     products = glob.glob('/etc/products.d/*.prod')
-    num_matches = 0
+    # suse-manager-server changed to
+    # multi-linux-manager-server, check both (bsc#1243437)
     suma_prods = [
-        '/etc/products.d/suse-manager-server.prod',
-        '/etc/products.d/sle-micro.prod'
+        'manager-server.prod',
```
jesusbv, line 2314: we could compare `manager-server` instead of `manager-server.prod` 
in case the name gets more text before the extension in the future

Thoughts ?

schaefi: I think product names are not changing that often such that we should be good with sticking to matching against a part of the name that will not accidentally match with another product. I would even go that far and match the strings against `endswith(...)` instead of `a in b` but with the `.prod` in the matching text that's solid enough

schaefi: Personally I would prefer a list of complete product names and no substring matches. In terms of product names the list will not grow endlessly

schaefi: The reason for this is, your code becomes explicit and you can spare out a comment that explains that `manager-server` actually matches two products. So I would just go and maintain 

```
suma_prods = [
    'suse-manager-server.prod',
    'multi-linux-manager-server.prod',
    'sle-micro.prod'
```

jesusbv: we can keep the minimum change in [fde052f](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/249/commits/fde052ffc4bfd4dfda02a3a4e9b8ada24357c033)

or keep with this change and the full name in [fc31964](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/249/commits/fc319640003e43077283b2097512276e7d05509c)

lib/cloudregister/registerutils.py
```diff
@@ -2308,16 +2308,18 @@ def is_suma_instance():
...
+        'sle-micro.prod'
+    ]
+    prod_matches = [
+        suma_prod for product in products
+        for suma_prod in suma_prods if suma_prod in product.lower()
```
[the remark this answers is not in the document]
jesusbv: kept the minimum change in [a04be76](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/249/commits/a04be768777543a03821eab93df50a32b5715170) as mentioned in the prev comment

lib/cloudregister/registerutils.py
```diff
@@ -2313,14 +2313,17 @@ def is_suma_instance():
...
+        '/etc/products.d/sl-micro.prod' # Micro 6.1 and 6.2
     ]
     for product in products:
         if product.lower() in suma_prods:
             num_matches += 1
 
-    return num_matches == len(suma_prods) - 1
+    # only one Micro + SUMA product possible at the same time for
+    # a SUMA instance
+    return num_matches == 2
```
jesusbv, line 2326: we are not checking if there is one Micro and one SUMA prod, I understand we do not need to do that here, that would be a different problem

PR #251, opened 2025-06-04
Add delay if registration is not completed during fail over

jesusbv:
When getting an update infrastructure server and there is a fail over and registration is not completed, we wait the regsharing time to be triggered plus 10 seconds to make sure the credentials are shared and sync'ed between the siblings

This is part of the changes to fix bsc#1242435

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -1902,7 +1902,7 @@ def is_registration_completed():
     """Indicate whether a registration is in process or completed based on the
        marker file. Note it is the responsibility of the process to properly
        manage the marker file"""
-    return os.path.exists(
+    return os.path.isfile(
         os.path.join(get_state_dir(), REGISTRATION_COMPLETED_MARKER)
     )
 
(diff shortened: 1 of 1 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -1888,6 +1897,16 @@ def is_new_registration():
     )
 
 
+# ----------------------------------------------------------------------------
+def is_registration_completed():
+    """Indicate whether a registration is in process or completed based on the
+       marker file. Note it is the responsibility of the process to properly
+       manage the marker file"""
+    return os.path.exists(
```
schaefi, line 1905: nitpick, could be changed to `os.path.isfile` because that's what we expect

jesusbv: fixed in [0d35dac](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/251/commits/0d35dacf5fc286bb2b5d4eaf848bd40e85998357)

jesusbv:
> But the flag needs to be set at some point, that should happen in the executable, i.e. `registercloudguest`

yes, two PRs, smaller changes, easier to review both
the other PR is [here](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/250)

rjschwei (approved):

PR #253, opened 2025-06-09
Do not wait for sleep on tests

jesusbv:
we should not wait for sleep on tests, as it makes the tests slower

make tests faster by mocking the call and assert time sleep calls


schaefi (approved):
yay that will speed up the test runs :smile:

PR #255, opened 2025-06-26
Use region server IP for checking

jesusbv:
In the case of a CSP blocking other CSPs IPs by firewall or not
allowed list, we should be able to verify IPv4 and IPv6 access

We check the list of region server IP addresses for connectivity 

This Fixes bsc#1245305


rjschwei (changes requested):
Let's switch to using the region server IPs for this test. Those can be obtained from the certs built into the images. That part is SUSE specific, and if there are no certs we fall back to 8.8.8.8.

And yes we can add a new configuration option `network_check_ip` that can be set to any IP anyone building their own images can setup to perform this test.

PR #256, opened 2025-07-11
Fix code state

rjschwei:
Code was not releasable due to untested changes being merged triggering an error when we were checking for access to the network for the IPv6 protocol. One error was that IPAdress objects were passed for socket creation instead of a string and the other error was that we did not pass an address at all.


schaefi (approved):
ok if you accept my additional commits

rjschwei:
I mad the minimal changes to pass the integration test suite. @schaefi thanks for fixing the unit tests, didn't get to that before I had to leave on Friday.

PR #258, opened 2025-07-23
Refactor state file setting

rjschwei:
We have a number of state files that get added and removed based on the state and progress of registration or based on billing model. Each of these had it's own implementation for creating and removing the state file. Create a utility functions to create and remove a state file and change the existing code to use the new functions.

### DIFF
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -174,7 +174,6 @@ def clean_framework_identifier():
 # ----------------------------------------------------------------------------
 def clean_smt_cache():
     """Clean the disk cache for SMT data"""
-
     smt_data = glob.glob(os.path.join(get_state_dir(), '*SMTInfo*'))
     for cache_entry in smt_data:
         os.unlink(cache_entry)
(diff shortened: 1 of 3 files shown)


lib/cloudregister/registerutils.py
```diff
@@ -2668,6 +2663,20 @@ def __remove_service(smt_server_name):
...
+    status = True
+    if os.path.exists(filepath):
+        status = False
+        try:
+            os.unlink(filepath)
+            status = True
+        except OSError as error:
+            logging.error(error)
+    return status
+
```
schaefi, line 2678: If another exception than an OSError appears, the method will raise and the application probably exits with a stacktrace. Is this behavior wanted/expected ? I'm asking because other state cleanups could be skipped in case another than an OSError exception is raised.

In case we want to capture all exceptions. This can be changed to

```python3
except Exception as error:
    logging.error('State cleanup for {0} failed with {1}'.format(filepath, error))
```
```

schaefi (approved):
Looks good to me

Just one question about exception handling

schaefi:
> Good refactor

my calendar says you are on vacation :smile:

lib/cloudregister/registerutils.py
rjschwei: Done, skipped the OS error treating all errors equally

PR #260, opened 2025-07-24
Appease the build service

rjschwei:
During install and removal test in the build service the order of dependent packages is non deterministic, this may lead to failures during build. Force the exit condition of preun to be 0 to avoid this non deterministic failure.

### DIFF
--- a/cloud-regionsrv-client.spec
+++ b/cloud-regionsrv-client.spec
@@ -233,6 +233,17 @@ fi
 if [ "$1" -eq 0 ] && [ -e "%{_sysconfdir}/regionserverclnt.cfg" ]; then
     %{_sbindir}/registercloudguest --clean
 fi
+# Avoid unpredictable errors in the build service. The build service check
+# that removes packages does not guarantee that the generic config package
+# gets removed during install/remove testing prior to removing the
+# cloud-regionsrv-client package. The generic example configuration
+# triggers errors during the clean operation. Therefore we force the exit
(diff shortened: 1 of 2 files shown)


rjschwei:
@schaefi this does the trick.

cloud-regionsrv-client.spec
```diff
@@ -233,6 +233,17 @@ fi
...
 fi
+# Avoid unpredictable errors in the build service. The build service check
+# that removes packages does not guarantee that the generic config package
+# gets removed during install/remove testing prior to removing the
+# cloud-regionsrv-client package. The generic example configuration
+# triggers errors during the clean operation. Therefore we force the exit
+# code to be 0 during build. If a user installs this package on a system with
+# and invalid config they have to uninstall the package with the "--noscripts"
+# option.
+if [ -e "/.build.log" ]; then
```
schaefi, line 244: Detecting if we are in a buildservice env or not was also something I had to implement in kiwi. Back in the days the BS team told me check for for `/.buildenv`. You are checking for `/.build.log` which is probably working as good as checking for the env file, but I believe it would be a bit more stable to check for the .buildenv as this file is going to be sourced by BS code and it must exist. It's also less likely that the name changes because this would also require code changes. For a log file that does not necessarily apply.

I'm fine if we leave it the way you implemented it though.

rjschwei: Switched to `.buildenv` thanks

PR #262, opened 2025-08-05
Store IP addresses as str

jesusbv:
When getting the IP addresses from the region server, we were storing `IPv4Address` and `IPv6Address` types

When we do the socket connection with that it fails

```
  File "/usr/lib64/python3.4/socket.py", line 498, in create_connection
    for res in getaddrinfo(host, port, 0, SOCK_STREAM):
  File "/usr/lib64/python3.4/socket.py", line 537, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
TypeError: getaddrinfo() argument 1 must be string or None
```

This Fixes bsc#1247466

### DIFF
--- a/integration_test-process.txt
+++ b/integration_test-process.txt
@@ -22,6 +22,7 @@ After installing the test package with "zypper in"
   + grep DOCKER_CONFIG /etc/profile.local has value
   + grep susecloud /etc/docker/daemon.json has value
   + grep susecloud /etc/containers/registries.conf has value
+  + instance-flavor-check return PAYG/BYOS accordingly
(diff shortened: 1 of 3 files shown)


jesusbv:
we could do the cast to `str` in the [call ](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/lib/cloudregister/registerutils.py#L1672)throwing the error

but I think it is safer to store them as strings from the method that gets them

schaefi (approved):
looks good :+1:

PR #263, opened 2025-08-08
Modify the credentials_files_are_equal interface

rjschwei:
Make the function less expensive on re-use. Previously the function encapsulated the extraction of the root directory from the zypper command line, meaning for every invocation we ended up calling the ps utility. A use case has arisen where we want to re-use the function to find orphaned credetials. In this case zypper is not running and trying to extract the root path becomes and expensive undertaking. With this change the function takes the rtarget root path as an additional argument, which by default is the system root, i.e. /. This pushed the responsibility of getting the target root to the caller. At this point only the urlresolver needs the target root functionality. Further, the urlresolver already has a feature to reduce the redundant calls when we are in the same zypper process leading to additional efficiency when extracting the target root.

Create 2 new constants to reduce repetition of the base credentials name used as a string and the zypper credentials path.

### DIFF
--- a/cloud-regionsrv-client.changes
+++ b/cloud-regionsrv-client.changes
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
-Fri Aug  8 14:46:13 UTC 2025 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
+$RELEASE_DATE
 
-- Update version to 10.5.2
+- Update version to NEXT
   + Fix the case where an instance can not get verified
     in any of the siblings, leaving credential files present
     under /etc/zypp/credentials.d
--- a/lib/cloudregister/registerutils.py
+++ b/lib/cloudregister/registerutils.py
@@ -41,6 +41,7 @@ from requests.auth import HTTPBasicAuth
 from cloudregister import smt
 
 AVAILABLE_SMT_SERVER_DATA_FILE_NAME = 'availableSMTInfo_%s.obj'
+BASE_CREDENTIALS_NAME = 'SCCcredentials'
 FRAMEWORK_IDENTIFIER = 'framework_info'
 HOSTSFILE_PATH = '/etc/hosts'
 NEW_REGISTRATION_MARKER = 'newregistration'
@@ -57,6 +58,7 @@ SUMA_REGISTRY_CONF_PATH = '/etc/uyuni/uyuni-tools.yaml'
 BASE_PRODUCT_PATH = '/etc/products.d/baseproduct'
 SUSE_REGISTRY = 'registry.suse.com'
 REGSHARING_SYNC_TIME = 30
+ZYPP_CREDENTIALS_PATH = '/etc/zypp/credentials.d/'
 
 requests.packages.urllib3.disable_warnings(
     requests.packages.urllib3.exceptions.InsecureRequestWarning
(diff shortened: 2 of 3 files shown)


cloud-regionsrv-client.changes
```diff
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------
-Fri Aug  8 14:46:13 UTC 2025 - Jesús Bermúdez Velázquez <jesus.bv@suse.com>
+$RELEASE_DATE
```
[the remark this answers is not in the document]
schaefi: Robert is using this concept as sort of an intermediate changelog state. The placeholders for the version and the release date will be filled out at the time someone submits the package. That guy has then to remember to mess with the changelog. In my projects I use the git history to produce a changelog, no need to do things twice and on release I use bumpversion followed by a tag. But this project does it differently and as long as nobody forget anything it will also work.

So long story short, this is by intention

lib/cloudregister/registerutils.py
```diff
@@ -313,18 +315,18 @@ def get_product_data(registration_target=None):
...
 
-    base_credentials_location = '/etc/zypp/credentials.d/'
-    target_root = get_zypper_target_root()
-    credentials_location = target_root + base_credentials_location
-    credentials_base = os.path.join(credentials_location, 'SCCcredentials')
-    credentials_header = os.path.join(credentials_location, repo_credentials)
+    credentials_location = os.path.join(target_root, ZYPP_CREDENTIALS_PATH)
+    credentials_base = os.path.join(credentials_location, BASE_CREDENTIALS_NAME)
+    credentials_header = os.path.join(
+        credentials_location, os.path.basename(repo_credentials)
```
schaefi, line 328: I'm not a big fan of os.path.join()  because it behaves weird in the following condition. Try this code

```python3
print(os.path.join('etc/zypp/credentials.d', '/foo'))

print(os.path.join('/etc/zypp/credentials.d', 'foo'))
```

Depending on the leading '/' the behavior is in my opinion not really a join. As you are using variables in your join parameters it is not guaranteed that the expected result comes out of this depending on if `credentials_location` and/or basename of repo_credentials comes up with a leading '/' or not. 

Because of this I avoid os.path.join and always use a combination of

```
os.path.normpath(os.sep.join([path_a, path_b]))
```

I leave it up to you but I think it's super risky to use os.path.join with dynamic variable content

lib/cloudregister/registerutils.py
```diff
@@ -2713,7 +2717,7 @@ def __replace_url_target(config_files, new_smt):
 def __generate_registry_auth_token(username=None, password=None):
     if not (username and password):
         username, password = get_credentials(
-            '/etc/zypp/credentials.d/SCCcredentials'
+            os.path.join(ZYPP_CREDENTIALS_PATH, BASE_CREDENTIALS_NAME)
```
schaefi, line 2720: same here, simple os.sep.join() would do it

schaefi (approved):
Looks ok, even though I would not use os.path.join(), see my comments and the example

lib/cloudregister/registerutils.py
rjschwei: Thanks for pointing this out, I was unaware of the unexpected behavior of `os.path.join()`. AT present everything is consistent w.r.t. to the separators `/`. And this change makes the code more consistent of using `join()` vs. `"a" + "b"` which also depended on "a" and "b" having the separator properly included. I'll merge this and will come back with a follow up PR that replaces all uses of `join()` with the new pattern.

PR #264, opened 2025-08-12
Orphaned credentials fix

jesusbv:


integration_test-process.txt
```diff
@@ -237,3 +237,14 @@ Extended testing
...
   + zypper up has no errors
+
+Orphaned credentials
+On SLE 15 PAYG
+- registercloudguest
+  + check all repos are there
+  + configure the instance/env so it can't access or verify on any sibling on the update infrastructure
+    + possible best approach is:
+      pointing the instance to test infra and edit test infra for this test
+  + rm /etc/products.d/$SOME_PRODUCT
```
rjschwei, line 248: This would not orphan credentials. The `__get_referenced_credentials()` function is driven based on repository setup, not based on installed products. A `.repo` will have a refernce to a credentais file in the URL, hence the name of the function. So the test should be to remove a `.repo` file for a specific product. No restriction to access the infra is needed as zypper is not executed in this path. An access restricted environment with `--clean` is yet another test case and for that why not use the `ip` command to simply drop packages, there's an example for this in the doc already in the fail over testing.

lib/cloudregister/registerutils.py
```diff
@@ -2609,14 +2609,29 @@ def __remove_credentials(smt_server_names):
     """Remove the server generated credentials"""
     logging.info('Deleting locally stored credentials')
     referenced_credentials = __get_referenced_credentials(smt_server_names)
```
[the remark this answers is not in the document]
rjschwei: The `__get_referenced_credentials`  does not necessarily collect credentials that match `SCCcredentials` that is a happy coincidence in most cases. But there can be, LTSS, is an example cases where credentials are referenced and have a different user name and password. And of course there is no guarantee that all modules, even from the same server will always have the same password.

lib/cloudregister/registerutils.py
```diff
@@ -2610,13 +2610,28 @@ def __remove_credentials(smt_server_names):
...
     # Special files that may exist but may not be referenced
-    referenced_credentials += [BASE_CREDENTIALS_NAME, 'NCCcredentials']
+    referenced_credentials += ['NCCcredentials']
     system_credentials = glob.glob(os.path.join(ZYPP_CREDENTIALS_PATH, '*'))
+    base_credentials_path = os.path.join(
```
rjschwei, line 2615: `os.sep.join`

schaefi:
With all this code here I wonder why we can run into the situation that the cleanup leaves credentials behind. To be honest I believe it would be better to refactor the cleanup code to be robust rather than adding more code to deal with a situation that a) should not happen and b) needs code to be detected for running the cleanup after the cleanup

I believe it would be better to invest time in a proper cleanup fix instead of adding more code because the entire registration code is already super hard to get a clear overview due to so many corner cases

I mark this as a no merge and suggest a conversation also with regards to merge conflicts

schaefi:
Something like

```
# init a git to all places where we do changes
git init ...

# some config file already exists
git add my_file
git commit -m origin_my_file
git tag my_file
# we change it
...

# on cleanup:
git reset --hard my_file --


# some config file is net new
# we create it
touch some_other_file...

# on cleanup
git ls-files --error-unmatch some_other_file || rm -f some_other_file
```

We have to keep track of a list of files that we create and/or modify. Shouldn't be that hard

The git info on that system can also be handy for debugging and I believe things becomes more stable and robust

Just ideas floating...

PR #265, opened 2025-08-12
Consist path merge

rjschwei:
-  update the uses of os.path.join to use either os.sep.join or
    os.path.normpath with os.sep.join to avoid oddities of th path.join
    implementation if the second argument starts with a /. While we have not
    had problems with this so far the previous use of path.join required
    careful setu of constants and processing when files are being used. Also
    discovered 1 incorrect use of path.join() and found another case where
    string concatenation was used.

- Add clarifying note to separate command that should be run on SLE 15 or later
    and commands that only apply to SLE 12. Also clean up after changes introduced
    with 1faefb5e66

### DIFF
--- a/tests/test_registercloudguest.py
+++ b/tests/test_registercloudguest.py
@@ -2,6 +2,7 @@ import inspect
 import json
 import os
 import requests
+import tempfile
 
 from io import StringIO
 from collections import namedtuple
@@ -182,6 +183,8 @@ def test_register_cloud_guest_force_reg_zypper_running(
     mock_is_zypper_running.return_value = True
     mock_get_available_smt_servers.return_value = ['some', 'smt', 'servers']
     mock_has_network_access.return_value = True
+    with tempfile.TemporaryDirectory(suffix='foo') as tdir:
+        mock_get_state_dir.return_value = tdir
     with raises(SystemExit) as sys_exit:
         register_cloud_guest.main(fake_args)
     assert sys_exit.value.code == 1
@@ -221,6 +224,8 @@ def test_register_cloud_guest_force_reg_zypper_runnning_write_config(
     mock_is_zypper_running.return_value = True
     mock_get_available_smt_servers.return_value = []
     mock_has_network_access.return_value = True
+    with tempfile.TemporaryDirectory(suffix='foo') as tdir:
+        mock_get_state_dir.return_value = tdir
     with raises(SystemExit) as sys_exit:
         register_cloud_guest.main(fake_args)
     assert sys_exit.value.code == 1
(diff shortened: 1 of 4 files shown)


schaefi:
@rjschwei you replaced all os.path.join with os.sep.join but I'm missing the normpath on the result. 

```python3
os.path.normpath(os.sep.join(['path_a', 'path_b', 'path_c']))
```

This might also ease the pain on the tests unless the tests mocked os.path.join which would be weird

schaefi:
@rjschwei I fixed the unit test, but what I found was a bit concerning as the mocking for the return of fetch_smt_data seems to be wrong. Actually that has nothing to do with your change and I have no idea why it got uncovered now. There are also other tests which still uses the same "wrong" mocking but they pass. It's very hard to read these tests so I'd like to ask @jesusbv to double check on the commit I made and maybe revise the tests in this regard. Thanks

tests/test_registercloudguest.py
```diff
@@ -272,7 +272,10 @@ def test_register_cloud_guest_force_reg_zypper_not_running_region_changed(
          SMTregistryName="registry-ec2.susecloud.net"
          region="antarctica-1"/>''')
 
-    smt_server = SMT(etree.fromstring(smt_data_ipv46))
+    child = etree.fromstring(smt_data_ipv46)
+
+    smt_server = SMT(child)
+
```
rjschwei, line 278: Duh, thanks again for picking up on this. The light for this should have really come on in my head but obviously it didn't. Much appreciate the help.

rjschwei:
ping @jesusbv @schaefi

rjschwei:
> @rjschwei you replaced all os.path.join with os.sep.join but I'm missing the normpath on the result.
> 
> ```python
> os.path.normpath(os.sep.join(['path_a', 'path_b', 'path_c']))
> ```
> 
> This might also ease the pain on the tests unless the tests mocked os.path.join which would be weird

@schaefi I don't quite see the point of using `normpath()` when what is being joined are defined constants. The constants do not change and cannot magically get additional `/`es.  So if `A` is defined as `/tmp` and `B` is defined as `foo.txt`  then `os.sep.join([A, B])` is the same as `os.path.normpath(os.sep.join([A, B]))` but we save the overhead of a function call if we just use `os.sep.join` for everything where the parts that get joined are dynamically generated I agree and I think I caught those occurrences.

As a safe guard we can add unit tests to make sure the constants don't change.

schaefi:
Just for the record; You consider an extra normpath() as overhead which is ok. However, I always norm my paths if they get created from variable content, just because I have seen so many code that cannot make the guarantees that you pointed out. Also code evolves and what is guaranteed today can be wrong tomorrow. The after effects of a wrong path can be really severe, the little overhead of a normpath() is imho worth the extra work. I'm just saying that because I will continue norming path joins for eventual code additions that I make to cloud-regionsrv-client in the future and I hope that this is then still ok. Thanks

jesusbv (approved):

PR #267, opened 2025-08-21
Add tests for constant settings

rjschwei:
During review of 7b1718c05d46 concerns were raised about ensuring that our defined constants that are represent either directory paths or file names may not have the expected definition w.r.t. path separators which would potentially lead to undesirable behavior. Add unit tests to ensure our constants remain as expected.


tests/test_registerutils.py
```diff
@@ -5793,6 +5821,20 @@ def get_test_config():
     return utils.get_config(data_path + '/regionserverclnt.cfg')
 
 
+def _check_dir_path(dir_name):
+    """A directory path is expected to start with a '/' and will _not_ have
+       a '/' at the end"""
+    assert (dir_name[0] == '/')
```
jesusbv, line 5827: we use `os.sep` in the code, we could use it in the tests

jesusbv (approved):
LGTM

schaefi (approved):
a wise idea, looks good :+1:

PR #268, opened 2025-09-04
Add python 3.13 to test matrix

schaefi:
SLE16 is based on python 3.13 and our code should
prove to work there. Thus this commit adds 3.13
into the game

### DIFF
--- a/tests/test_registerutils.py
+++ b/tests/test_registerutils.py
@@ -11,6 +11,8 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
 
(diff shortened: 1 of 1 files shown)


schaefi:
@rjschwei you are approving stuff that doesn't work :-)) As you can see, the tests are not prepared to work on python 3.13. We have luck that the actual python code seems to work as I have tested the registration on a SLE16 client yesterday. Nevertheless this PR needs further adaptions. So once the checkmark is green I will ask you for a review again ;)

rjschwei (approved):

PR #269, opened 2025-09-09
Move to poetry and make this project a real python project

schaefi:
I'm aware this is a big change and I was very careful in making these changes and also tested them. To allow you a proper review of the changes I'd like to ask you to take a look at the changes in the .spec file and please compare the diff tarballs that I will attach to this PR which compares the `rpm -qpl` output of the current package with the new package build that you can find here

The rest of the changes in this PR covers the following topics

* drop old and obsolete use of setup.py and setuptools except for the SLE12 target
* Fixed unit tests and flake8 issues
* Use proper entrypoint definitions for all provided bin tools. This required me to add an entry method which I consistently called `app`
* Update structure of code to match with current python standards, especially a clear setup of the `cloudregister` namespace
* Move entire development to `poetry`, drop tox completely
* Stay compatible with runtime code down to SLE12 (python 3.4)

This is really a big change but please support me in performing this action otherwise maintaining this python project will become even more cumbersome and troublesome in the future.

For development, just call

```
# run flake tests
make check

# run unit tests
make test

# to run a virtual env session
poetry shell

# to run some tool
poetry run ...

# to build the package
make package
cp dist/* ...
```


schaefi:
[SLE_15_cloud-regionsrv-client.package.patch.TGZ](https://github.com/user-attachments/files/22233660/SLE_15_cloud-regionsrv-client.package.patch.TGZ)

schaefi:
Also for review of the python structure it's handy to checkout the full branch `move_to_poetry` rather than review patch based

schaefi:
I tested client registration in SLE15 and SLFO successfully with the package build from the initial comment. I have not tested SLE12 registrations as I don't have a client build at hand

schaefi:
> I have not tested SLE12 registrations as I don't have a client build at hand

I have now also tested with SLE12, using the client build from here
* https://build.suse.de/package/show/home:mschaefer/RMTClient-sle12

registration is also working there

schaefi:
This completes testing results, plus diff of the package on all code bases. Please double check

Thanks

rjschwei:
Well that merge was a little faster than I had hoped. I had committed to get to a review this week. Anyway, I will review and make comments and if I find things we can address them in a follow up request.

Makefile
```diff
@@ -1,34 +1,37 @@
...
-files = Makefile README LICENSE setup.py cloud-regionsrv-client.spec
+buildroot = /
 
-nv = $(shell rpm -q --specfile --qf '%{NAME}-%{VERSION}|' *.spec | cut -d'|' -f1)
-verSpec = $(shell rpm -q --specfile --qf '%{VERSION}|' *.spec | cut -d'|' -f1)
-verSrc = $(shell cat lib/cloudregister/VERSION)
-ifneq "$(verSpec)" "$(verSrc)"
-    $(error "Version mismatch, will not take any action")
-endif
+python_version = 3
```
rjschwei, line 3: We are going to have problems wit this at some point. The intention is to drop the `python3` symlink at some point, we should see if we cannot figure out some other way to find the appropriate Python interpreter.

cloudregister/defaults.py
```diff
@@ -0,0 +1,16 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library.
+#
+#
+LOG_FILE = '/var/log/cloudregister'
+ZYPP_SERVICES = '/etc/zypp/services.d'
```
rjschwei, line 16: If we are skinning the cat this way then we should probably pull all the constants that are in the utils file into this file.

cloudregister/switchcloudguestservices.py
```diff
@@ -0,0 +1,133 @@
...
+            if service not in current_services:
+                os.unlink(service)
+        os.unlink(service_revert_file)
+        os.unlink("%s.lock" % service_revert_file)
+        # zypper is slow, wait loop
+        while utils.get_zypper_pid():
+            time.sleep(5)
+        utils.exec_subprocess(['zypper', '--non-interactive', 'refs', '-f'])
+    else:
+        utils.switch_services_to_plugin()
```
rjschwei, line 133: Since we are doing major surgery we could have dropped this script at this point. We've not handed out `https://` from our update infrastructure in 5 or 6 years. If anyone still has an instance that old and has not update cloud-regionsrv-client yet access to the update infrastructure is probably broken anyway.

package/cloud-regionsrv-client.changes
```diff
@@ -1,3 +1,29 @@
+-------------------------------------------------------------------
+Wed Sep 17 07:47:42 UTC 2025 - Marcus Schäfer <marcus.schaefer@suse.com>
+
+- Update to version 10.5.3
```
rjschwei, line 4: Due to the spec file issue this is actually not releasable. We still have the open conversation about how we want to handle version bumps in this project. I am generally not in favor that every change constitutes a version bump, but can certainly be convinced otherwise.

package/cloud-regionsrv-client.spec
```diff
@@ -88,10 +88,14 @@ BuildRequires:  %{pythons}-zypp-plugin
 %if 0%{?suse_version} >= 1600
 BuildRequires:  %{pythons}-pip
 BuildRequires:  %{pythons}-wheel
+BuildRequires:  %{pythons}-poetry-core >= 1.2.0
 %endif
 %if 0%{?suse_version} > 1315
 BuildRequires:  %{pythons}-toml
 %endif
+%if 0%{?suse_version} && 0%{?suse_version} < 1600 && 0%{?suse_version} > 1315
+BuildRequires:  python311-poetry-core >= 1.2.0
```
rjschwei, line 97: This breaks on SLE 16. Should use the `{pythons}` definition. SEL 16 switched to Python 3.13

Makefile
schaefi: > The intention is to drop the python3 symlink at some point

I doubt this will ever happen. But if it happens it's easy to change python_lookup_name to watch for `python` and `python3` as long as just `python` is considered to be python v2.x the above is the most robust way to do it. I use this concept in kiwi and it was approved by all distros (Fedora, RHEL, Debian, Ubuntu) to be safe. If SUSE really makes that move I'm happy to follow and add code, until then there is no need to mess with it

cloudregister/defaults.py
schaefi: yes I agree and this is also my plan for further PRs. The reason why I did it only partially was because this PR was already super hard to review and as such not everything was moved.

package/cloud-regionsrv-client.changes
schaefi: I bumped the patch version due to changes on the logging which results in a more verbose information of the log file. Thus this refactor had an impact for users which imho requires a patch level bump.

> I am generally not in favor that every change constitutes a version bump

I agree and I only bump versions on changes or changesets that impacts behavior or fix a bug. I don't see what should be wrong with that

package/cloud-regionsrv-client.spec
schaefi: This is not correct. The %if statement if you read it from the current master branch looks like this

```
%if 0%{?suse_version} && 0%{?suse_version} < 1600 && 0%{?suse_version} > 1315 && 0%{?sle_version} > 150300
BuildRequires:  python311-poetry-core >= 1.2.0
%endif
```

And this matches for anything > SLES15-SP3 and < SLE16. In this window poetry exists on SLES15 for python 3.11 only and for that reason exactly this version was selected. This is required because cloud-regionsrv-client on SLE15 builds as python 3.6 package which is what you enforce with your `pythons` macro, but for that version no poetry exists. So we fetch as a build requires the 3.11 stack for poetry but build a 3.6 target module which is not a problem.

Especially that part I have extensively tested. See here:

As you can see I build for all targets that are listed via `osc maintained cloud-regionsrv-client` such that there will be zero breakage when this package gets submitted to the SUSE namespace

I also wrote that I have tested all of this targets with a real integration test, meaning a client build against all the targets for which this tool is supported including running a registration

Thanks

rjschwei: Thanks for the additional details and sorry I missed the version condition and created noise.

PR #270, opened 2025-09-17
Fix use of logging facility

### LINKED ISSUE #188: Inconsistent use of messaging
I found several places in the code where ```logging.``` is mixed with ```print``` and the print statements are also mixed with printing to the default stdout channel or to the stderr channel. I don't see a good reason why the messaging of the tooling is mixed up that way.

I think we should change this to use the python logging facility exclusively and get rid of any extra print statements. If there is need to be to handle specific messages differently it can be done through the logging facility too

schaefi:
Use logging facility in the desired way throughout the entire code base. This includes the following changes and refactor:

* Add handler and formatter for the logfile containing more information about function and position in code for the message
* Add handler for stdout (INFO and WARNING)
* Add handler for stderr (ERROR).
* Implement Logger class providing the logging setup and methods
* Drop the start_logging() method.
* Fix and refactor all unit tests around the use of logging with a proper fixture and place all tests for registerutils into its own class TestRegisterUtils.
* Add --debug switch for registercloudguest. Allow to increase logfile information. All messages produced via log.debug(...) in code will be part of the logfile. Debug messages will not be shown on the console
* Update SLE12 patches due to logging refactor
* Use --debug flag in guestregister service

This Fixes #188

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ check: setup
 	# python flake tests
 	poetry run flake8 --statistics -j auto --count cloudregister
 	poetry run flake8 --statistics -j auto --count test/unit
+	poetry run flake8 --statistics -j auto --count usr/lib/zypp/plugins/urlresolver
 
 test: setup
 	# unit tests
--- a/cloudregister/cloudguest_lic_watcher.py
+++ b/cloudregister/cloudguest_lic_watcher.py
@@ -13,14 +13,19 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library.
 
-import logging
 import os
 
 import instance_billing_flavor_check.utils as inst_flvr_utils
 import cloudregister.registerutils as utils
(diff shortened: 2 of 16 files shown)


schaefi:
Going forward I think some messages are not so meaningful and we can change a lot on the level of what is good for a user compared to what is good for us from a debugging perspective.

schaefi:
This PR was tested in registrations for

* SLE12
* SLE15 SP4/5
* SLE16

Please find a package build here:

smarlowucf (approved):

rjschwei:
> Going forward I think some messages are not so meaningful and we can change a lot on the level of what is good for a user compared to what is good for us from a debugging perspective.

We have to be careful with this. We do not want to get into a situation where we have to go back to the user and say "now turn on this switch". When we get a supportconfig from a system with issues that should contain all the information we need to make a determination of what is going on.

PR #273, opened 2025-09-18
Consolidate command execution

schaefi:
Make use of the central exec_subprocess method. In addition the exec calls are logged as debug message and are allowed to appear in the logfile.


cloudregister/cloudguest_repo_service.py
```diff
@@ -144,10 +143,6 @@ def app():
...
         if has_plugin_format and not trigger_set:
             trigger_set = True
-            # Python has no sane way to create a detached process
-            subprocess.Popen(
-                ['nohup', '/usr/sbin/switchcloudguestservices'],
-                stdin=subprocess.DEVNULL,
-                stdout=subprocess.DEVNULL,
-                stderr=subprocess.DEVNULL
+            utils.exec_subprocess(
+                ['/usr/sbin/switchcloudguestservices'], disconnect=True
```
rjschwei, line 147: But this is disappearing ;) in #274 so let's get that merged first than this just goes away.

cloudregister/registerutils.py
```diff
@@ -346,30 +346,51 @@ def enable_repository(repo_name):
...
 
 
 # ----------------------------------------------------------------------------
-def exec_subprocess(cmd, return_output=False):
-    """Execute the given command as a subprocess (blocking)
-       Returns one off:
-           - exit code of the command
-           - stdout, stderr and exit code
-           - -1 indicates an exception"""
+def exec_subprocess(cmd, tolog=True, disconnect=False):
```
rjschwei, line 357: We must obfuscate the registration code. With this change an invocation of `registercloudguest -r 123456` would end up logging `SUSEConnect --url....... -r 123456` and that is not an option. So we eithre have to disable logging, i.e. set `tolog` false when we get a registration code from the user or we need to put a filter in this code that turns the reg code into `XXXXXX` or some other meaningless collection of characters.

cloudregister/registerutils.py
```diff
@@ -346,30 +346,51 @@ def enable_repository(repo_name):
...
+    """
+    Execute the given command as a subprocess (blocking)
+    Returns a tuple list:
+        (binary_stdout, binary_stderr, exitcode)
+        A -1 exitcode indicates an OSError at command invocation
+    """
+    stdout = b''
+    stderr = b''
+    ecode = -1
```
rjschwei, line 366: `ecode` may imply "error code" that in turns implies that we expect an error. But we shouldn't so maybe this should be called `rcode`

rjschwei (changes requested):

cloudregister/cloudguest_repo_service.py
schaefi: yep done, also dropped the feature for daemon like process calls, as it was used only once in this code that got dropped

cloudregister/registerutils.py
schaefi: done

cloudregister/registerutils.py
schaefi: I added a commit to prevent the exec logging from the exec_subprocess in register_product(). If you look into this method you also see that we create a special obfuscated log information for the call. Thus there is no need to let exec_subprocess log it again. That should do the trick

schaefi:
updating my package build with the latest changes and give it a try. Will also test LTSS registration to see that we don't expose any sensitive information. I'll come back with a new logfile...

schaefi:
Testing looks good to me so far. When running 

```
registercloudguest -r ...
```

you get among others the following information 

```
2025-09-19 12:46:57,136: INFO: Running LTSS registration...
2025-09-19 12:46:57,141: INFO: Registration: /usr/sbin/SUSEConnect --url https://localhost:44300 --product SLES-LTSS/15.4/x86_64 --instance-data /var/cache/cloudregister/c537d815-7388-476b-96e2-7205ae9cf7f9 --regcode XXXX
```

so the regcode is obfuscated.

In the logfile we see

```
2025-09-19 12:46:57,136: registercloudguest: 237 setup_ltss_registration: INFO: Running LTSS registration...
2025-09-19 12:46:57,141: registerutils: 464 register_product: INFO: Registration: /usr/sbin/SUSEConnect --url https://localhost:44300 --product SLES-LTSS/15.4/x86_64 --instance-data /var/cache/cloudregister/c537d815-7388-476b-96e2-7205ae9cf7f9 --regcode XXXX
```

No exec_subprocess logging.

I also did a search for the used regcode in the full logfile and got no matches

PR #274, opened 2025-09-18
Drop obsolete switchcloudguestservices.py

### LINKED ISSUE #185: Remove switch_services_to_plugin
At the end of 2024 it will be roughly 5 years since we switched the update server technology from SMT to RMT. During this switch we also switched from handing out http based URLs for the repositories to use the zypper url resolver mechanism. The switch_services_to_plugin code handles the transition from http based repo urls to resolver based urls. In order for this to be useful an instance would have to have been stopped for 5 years which means it is very unlikely that the user can get an updated cloud-regionsrv-client package from the update infrastructure to begin with.

Therefore the code can be removed. Removal after December 31st 2024.

schaefi:
The service to handle the transition from http based repo urls to resolver based urls can be dropped because the transition period is over. This Fixes #185


schaefi:
@aosthof this is related to a bug we discussed today and I can't remember the number. Can you please help me as long as your mind is still fresh :) Thanks much

aosthof:
The related bug is https://bugzilla.suse.com/show_bug.cgi?id=1245370.

PR #279, opened 2025-10-01
Refactor cleanup code

schaefi:
This commit refactors the cleanup code into single responsibility methods for each type of cleanup. The registration client has three areas for the cleanup

1. API delete requests to the SCC/RMT API
2. Cleanup of registration client specific cache data
3. Cleanup of configuration data created and or modified in /etc

The code changes refactors the existing cleanup methods to not intermix the different cleanup types and provides methods to implement cleanup methods properly to the type of cleanup.

### DIFF
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -39,7 +39,10 @@ import uuid
 from cloudregister.logger import Logger
 from cloudregister.defaults import (
     LOG_FILE,
-    ZYPP_SERVICES
+    ZYPP_SERVICES,
+    DOCKER_CONFIG_PATH,
+    REGISTERED_SMT_SERVER_DATA_FILE_NAME,
+    AVAILABLE_SMT_SERVER_DATA_FILE_NAME
 )
 import cloudregister.registerutils as utils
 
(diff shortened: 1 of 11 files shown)


cloudregister/registercloudguest.py
```diff
@@ -809,7 +818,10 @@ def main(args):
         os.unlink(instance_data_filepath)
 
     if registration_returncode:
-        cleanup()
+        utils.deregister_non_free_extensions()
+        utils.deregister_from_update_infrastructure()
+        utils.deregister_from_SCC()
+        utils.clean_cache()
```
rjschwei, line 824: misses `/etc/hosts` cleanup

cloudregister/registercloudguest.py
```diff
@@ -317,7 +318,10 @@ def register_base_product(
                 # 67: Server responded with error: see log output
                 log.error('Baseproduct registration failed')
                 log.error(error_message)
-                cleanup()
+                utils.deregister_non_free_extensions()
+                utils.deregister_from_update_infrastructure()
+                utils.deregister_from_SCC()
+                utils.clean_cache()
```
rjschwei, line 324: Misses `/etc/hosts` clean up

rjschwei (changes requested):

cloudregister/registercloudguest.py
schaefi: When I refactored this the hosts cleanup was not part of it which is the reason why it's not there. Please take a closer look to the method

`def register_base_product`

This method runs in a loop `while not base_registered:` trying several times. Each time it fails it does a full cleanup. You will see that also utils.clean_hosts_file() is part of it. At the end when all attempts failed it also does the above cleanup when it already has done the cleanup for a failed attempt. As such the hosts cleanup was called in this sequence of failed attempts. That's why I did not add it here. Please verify if this is correct. In this refactor I tried to not add code logic changes

Thanks

PR #280, opened 2025-10-01
Add new Git class for content management

### LINKED ISSUE #271: Idea on refactor config management
So far we implement our own way to manage (cleanup/revert/create) config files in the system that gets registered. However that cleanup code has already lead to inconsistent system states and we have no opportunity to really see what we change. My idea here is to make use of `git` to manage and also cleanup data in the following way

* init a git at places where we do changes prior anything else
```
git init /etc
```

* some config file already exists
```
git add my_file
git commit -m origin_my_file
git tag my_file
# we change it
...

# on cleanup:
git reset --hard my_file --
```

* some config file is net new
```
# we create it
touch some_other_file...

# on cleanup
git ls-files --error-unmatch some_other_file || rm -f some_other_file
```

We have to keep track of a list of files that we create and/or modify. Shouldn't be that hard. So far the list of files I found that we change or create is as follows

* /etc/zypp/services.d
* /etc/hosts
* /etc/containers/config.json
* /etc/profile.local
* /etc/containers/registries.conf
* /etc/docker/daemon.json
* /etc/uyuni/uyuni-tools.yaml

So just `/etc` git managed

schaefi:
As of today all files created or modified by the client registration in the systems /etc path are more or less unknown. There is no way to differentiate between
modifications done by the registration client and changes done by a user. As we are touching files that are likely to be changed by users e.g. etc/hosts it is handy to put all those under git control. This allows the client to perform the following actions:

* git based restore to the state prior registration on cleanup
* possible detection of file changes done by a user between registration and cleanup
* proper rollback in case of exceptions or interrupts during the registration via python context manager

This feature can be enabled via the --git option
and requires the user to install git beforehand as we cannot force users to install git.
This Fixes #271

### DIFF
--- a/doc/man/man1/registercloudguest.1
+++ b/doc/man/man1/registercloudguest.1
@@ -51,6 +51,18 @@ removed from SCC.
 Delay the process by the given number of seconds.
 .IP "--debug"
 Debug mode, increase verbosity of logfile information.
+.IP "--git"
+Use git for content management. In this mode a git repo is initialized
+for the /etc directory of the system. All files modified and/or produced
+by the registration client will be managed via the git content manager
+and the registration process runs in a context manager to allow for
+seamless cleanup in case of errors. As files are known to git in this
+mode. Also the user initiated cleanup or force registration will make
+use of the content manager to restore data to is original state. As a
+user there is also the opportunity to watch the changes done by the
+registration client via the git utility. As we can't force git to
+become part of the system, this mode is fully optional and requires
+the installation of the git package prior use.
 .IP "-e --email"
 The e-mail address associated with a subscription in SCC. Only use this option
 for BYOS instances and in conjunction with "--regcode" option.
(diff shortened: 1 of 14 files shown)


doc/man/man1/registercloudguest.1
```diff
@@ -51,6 +51,18 @@ removed from SCC.
...
 .IP "--debug"
 Debug mode, increase verbosity of logfile information.
+.IP "--git"
+Use git for content management. In this mode a git repo is initialized
+for the /etc directory of the system. All files modified and/or produced
+by the registration client will be managed via the git content manager
+and the registration process runs in a context manager to allow for
+seamless cleanup in case of errors. As files are known to git in this
+mode, also the user initiated cleanup or force registration will make
+use of the content manager to restore data to is origin state. As a
```
smarlowucf: s/to is origin/to it's original

Issue #283, opened 2025-12-19
Info messages are sent to the terminal/console


rjschwei:
Since ef53273953682 and because of https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/cloudregister/logger.py#L33 info message are sent to the terminal. This makes as standard registration performed by a user via the `registercloudguest` command very noisy.

What used to be

```
registercloudguest
Instance registry setup done, sessions must be restarted !
Registration succeeded
```

is now

```
# registercloudguest 
2025-12-19 21:29:53,322: INFO: Region server arguments: ?regionHint=us-east-1
2025-12-19 21:29:53,324: INFO: Using API: regionInfo
2025-12-19 21:29:53,329: INFO: Region server arguments: ?regionHint=us-east-1
2025-12-19 21:29:53,329: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Skipping IPv6 protocol version, no network configuration
2025-12-19 21:29:53,330: INFO: Getting update server information, attempt 1
2025-12-19 21:29:53,330: INFO: 	Using region server: 52.79.82.165
.............MANY MANY MORE LINES.............
```
This is information we want in the log file but not in the terminal when the user runs `registercloudguest`

rjschwei:
This needs fixing before the next release.

PR #284, opened 2025-12-19
Update message regarding registry setup

rjschwei:
The previous message was considered to contain insufficient information and potentially leading to confusion, bsc#1254960. Update the message to be more explicit about the meaning of "session".

### DIFF
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -178,17 +178,17 @@ def setup_registry(registration_target):
                         LOG_FILE
                     )
                 )
+                # TODO, we probably do not want to exit here, we do not have
+                # clear documented decision tree.
                 sys.exit(1)
             elif docker_setup_ok is True:
                 log.info('Successfully added docker to the registry setup')
         log.info(
             'Instance is setup to access container registry, nothing to do'
         )
+        # Podman registry data is already set up, nothing to do
         return True
-    log.info(
-        'Adding registry setup to instance, all existing shell '
-        'sessions must be restarted to use the registry.'
-    )
+    log.info('Adding registry setup to instance')
     registry_setup_ok = False
     if utils.prepare_registry_setup(registry_fqdn, user, password):
         log.info('Adding podman support')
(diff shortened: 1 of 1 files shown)


cloudregister/registercloudguest.py
```diff
@@ -205,7 +202,9 @@ def setup_registry(registration_target):
             )
         )
         return False
-    log.warning('Instance registry setup done, sessions must be restarted !')
+    # This message should be on stdout, see issue#238
+    log.info('Instance registry setup completed, '
+             'all existing shell sessions must be restarted !')
```
smarlowucf: Nitpick: last time I read the SUSE docs docs there was a mention to avoid words like must. Not sure if that's still the case. Another idea "all existing shell sessions require a restart".

schaefi (changes requested):
I agree the messaging of the client needs to be improved. When I cleaned up the messaging in the first place I made sure that everything is properly handled through the logging facility. It's now time to properly classify the messages which I'm working on in another PR for #283

cloudregister/registercloudguest.py
```diff
@@ -178,17 +178,14 @@ def setup_registry(registration_target):
                         LOG_FILE
                     )
                 )
-                sys.exit(1)
```
schaefi: You dropped this sys.exit() call. This will lead to a `return True` of `setup_registry()` and will consequently prevent the right thing to happen at 

```
# 1. check/setup the container registry for this target
        if not setup_registry(registration_smt):
            utils.clean_registry_setup()
            sys.exit(1)
```

and

```
# Setup container registry
    if not setup_registry(registration_target):
        cleanup()
        sys.exit(1)
```

As such and if you don't have a specific reason to change this behavior I recommend to keep it or change it to a `return False` 

 I do not understand why it is subject of a log change though ?

schaefi: and... has this change been tested ?

rjschwei: It's a latent bug, I think. but yes I missed something and with what I missed it definitely becomes a bigger topic. Added the exit condition back.

PR #287, opened 2026-01-06
Properly set the registration return value indicator on registration success

rjschwei:
In the situation where the initial base product registration against a specific update server failed but a registration of he base product to another update server succeeded we did not set the global registration code indicator to 0 and we retained the return code of the initial failed registration. Improperly indicating that registration had failed.


schaefi (approved):
good catch, thanks

PR #288, opened 2026-01-07
Add a wait loop if zypper is busy

rjschwei:
- Users my attempt to perform package management operations before registration is complete, ssh is generaly available prior to registration completion. The package management operations may hit a window where package management is not locked by the registercloudguest process. In such a condition it is possible that the registercloudguest process encounters a locked package management system and the registration would fail. This change adds a waitloop when package management is locked. We wait up to 80 seconds. If the package management system is still locked after 80 seconds the operation will fail in the same way as previously. This change is related to bsc#1254303.
- The logic for error handling during registration was not correct only treating the suseconnect error code 67 as a registration failure. However, other codes also lead to module registration failure. Update the logic to properly handle the conditions, including return codes from zypper that are passed through by suseconnect.


schaefi (approved):
Looks good

PR #289, opened 2026-01-07
Process lock

rjschwei:
ssh is generally available prior to registration completion. This allows a user or automation to start another registercloudguest process and as such interfere with any possibly running process. Lock the process to ensure only 1 registraion runs.

### DIFF
--- a/cloudregister/registerutils.py
+++ b/cloudregister/registerutils.py
@@ -49,6 +49,7 @@ from cloudregister.defaults import (
     REGISTRATION_COMPLETED_MARKER,
     REGISTRATION_DATA_DIR,
     REGISTERED_SMT_SERVER_DATA_FILE_NAME,
+    REGISTRATION_PROC_LOCK_FILE_NAME,
     RMT_AS_SCC_PROXY_MARKER,
     REGISTRY_CREDENTIALS_PATH,
     PROFILE_LOCAL_PATH,
@@ -143,6 +144,7 @@ def clean_cache():
     clean_framework_identifier()
     clear_registration_completed_flag()
     clean_registered_smt_data_file()
+    unlock_registration()
 
 
 # ----------------------------------------------------------------------------
@@ -469,6 +471,22 @@ def get_register_cmd():
                 break
 
     return register_cmd
+
+
+# ----------------------------------------------------------------------------
+def is_registration_running():
+    """Check if registercloudguest is already running"""
+    return os.path.exists(
+        os.sep.join([get_state_dir(), REGISTRATION_PROC_LOCK_FILE_NAME])
+    )
+
+
+# ----------------------------------------------------------------------------
+def lock_registration():
+    """Set the process lock for registercloudguest"""
+    _set_state_file(
+        os.sep.join([get_state_dir(), REGISTRATION_PROC_LOCK_FILE_NAME])
+    )
 
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


schaefi:
Hmm, while I agree that the locking makes sense I would not implement it in the way it is done here. From my perspective this should be done by calling the existing `main()` method in a try/finally block which acquires and releases a lock. That way the implementation becomes more readable and robust and does not require to release a lock multiple times with the risk to forget that at some code point. Here is how I would do this

```python
import fcntl

lock_file = '/home/ms/foo.lock'

def lock(lock_file):
    fd = open(lock_file, 'a')
    fcntl.flock(fd.fileno(), fcntl.LOCK_EX)
    return fd

def release(fd):
    fcntl.flock(fd.fileno(), fcntl.LOCK_UN)

def main():
    # The main method already exists. I added it here to demonstrate what I mean
    # my main method just prints "hello" and waits for input. This allows you to test
    # this little example in an easy way
    input('hello')



try:
    fd = lock(lock_file)
    main()
finally:
    release(fd)
```

Thoughts ?

schaefi:
This also survives any 'Ctrl-C' or other interruption type which your implementation would not cover imho

rjschwei:
> Hmm, while I agree that the locking makes sense I would not implement it in the way it is done here. From my perspective this should be done by calling the existing `main()` method in a try/finally block which acquires and releases a lock. That way the implementation becomes more readable and robust and does not require to release a lock multiple times with the risk to forget that at some code point. Here is how I would do this
> 
> ```python
> import fcntl
> 
> lock_file = '/home/ms/foo.lock'
> 
> def lock(lock_file):
>     fd = open(lock_file, 'a')
>     fcntl.flock(fd.fileno(), fcntl.LOCK_EX)
>     return fd
> 
> def release(fd):
>     fcntl.flock(fd.fileno(), fcntl.LOCK_UN)
> 
> def main():
>     # The main method already exists. I added it here to demonstrate what I mean
>     # my main method just prints "hello" and waits for input. This allows you to test
>     # this little example in an easy way
>     input('hello')
> 
> 
> 
> try:
>     fd = lock(lock_file)
>     main()
> finally:
>     release(fd)
> ```
> 
> Thoughts ?

Well learning something new here, that the `finally` path is executed even after a `sys.exit()`. As such yes that is more robust than the current "splatter unlock calls everywhere approach". That said, I'd stay with using the lock, unlock and is_running functions rather than tying up a file handle for the whole execution of the process.

cloudregister/registerutils.py
```diff
@@ -471,6 +473,22 @@ def get_register_cmd():
     return register_cmd
 
 
+# ----------------------------------------------------------------------------
+def is_registration_running():
+    """Check if registercloudguest is already running"""
+    return os.path.exists(
+        os.sep.join([get_state_dir(), REGISTRATION_PROC_LOCK_FILE_NAME])
+    )
+
```
schaefi: To be honest, I don't like this. It's not robust. The simple presence of the file is not really an indicator of something is running. However a main function only called with an open descriptor that has a lock set by libc fcntl is not something a user in the system can easily abuse. Setting the lock on an open descriptor is also considered more of a standard way of doing it.

If you insist on using os.path.exists I still strongly vote for writing a class instead of adding more lines of code to the ~3000 lines of registerutils.py

Sorry really not so much convinced about this solution

schaefi:
Closing on behalf of #290

PR #290, opened 2026-01-08
Implement process locking

schaefi:
Allow only one registercloudguest process to run at a time

### DIFF
--- a/cloudregister/lock.py
+++ b/cloudregister/lock.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2026, SUSE LLC, All rights reserved.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 3.0 of the License, or (at your option) any later version.
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


schaefi:
Testing finished, worked for me. It could be implemented even nicer when we put that into its own class instead of the already too bug utils code. I'm willing to do this if my solution would be accepted

cloudregister/lock.py
```diff
@@ -0,0 +1,43 @@
...
+    def lock(self):
+        """
+        Set file lock to the process lock file
+        This call blocks if the file is already locked
+        """
+        fd = open(self.lock_file, 'a')
+        try:
+            fcntl.flock(fd.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except IOError:
+            logging.warning('registercloudguest is locked by another instance')
```
smarlowucf: Should we add in the warning comment that the thread will wait for release of lock? Otherwise it might be assumed that the thread should exit.

smarlowucf (approved):

cloudregister/lock.py
schaefi: @smarlowucf yes the message can be more explicit. Would you mind to provide an example that you see fit ?

Other than that I still haven't convinced @rjschwei that using an open filehandle and libc fcntl attributes is better than os.path.exists :) 

As such we are still negotiating which locking PR is the one we will merge

schaefi:
@rjschwei This implements the locking behavior as we have discussed it

PR #293, opened 2026-01-08
Msg classification

rjschwei:
- Quiet the output the stdout.  After switching to the new logging setup to output to stdout when invoked from the command line was still to verbose.
-  Switch the SMT class to use the new logging infrastructure.
-  Make sure we get all the messages we want in the log file so we can debug any issues without having to go back to the user and asking for more information.

### DIFF
--- a/cloudregister/logger.py
+++ b/cloudregister/logger.py
@@ -40,7 +40,7 @@ class Logger:
         error.setFormatter(formatter)
         logging.getLogger().addHandler(error)
 
-    def set_logfile(self, logfile, debug=False):
+    def set_logfile(self, logfile, debug=True):
         # Add file handler, with level DEBUG (all)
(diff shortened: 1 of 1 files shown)


schaefi:
In this course of explanation you can also say that the user doesn't need to be informed about anything and can just expect access to repos and registries.

I'm ok if this is how we want it to behave

schaefi (approved):
So let's do it that way... and maybe become a bit more chatty in the future. Tools that doesn't say what they do are usually suspicious :)

rjschwei:
The idea is to convey the basic information when the command is called directly, which in general should not really happen except for BYOS instances. And if the user cares about more details they can look in the log.

Having the messages about the registry setup makes the "restart you shell session" more meaningful, I think.

Thanks

PR #295, opened 2026-01-09
Switch to new logging infrastructure

rjschwei:
Switch the plugin for GCE to the new logging setup to ensure we get all messages in the log file.

### DIFF
--- a/cloudregister/googlece.py
+++ b/cloudregister/googlece.py
@@ -29,8 +29,11 @@ def generateRegionSrvArgs():
     try:
         zoneResp = requests.get(metaDataUrl + zoneInfo, headers=headers)
     except requests.exceptions.RequestException:
-        msg = 'Unable to determine zone information from "%s"'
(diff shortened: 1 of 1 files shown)


cloudregister/googlece.py
```diff
@@ -27,21 +30,21 @@ def generateRegionSrvArgs():
...
+            log.debug(
                 'Unable to form region string from text: %s' % zoneResp.text
             )
             return
     else:
-        logging.warning('Unable to get zone metadata')
-        logging.warning('\tReturn code: %d' % zoneResp.status_code)
-        logging.warning('\tMessage: %s' % zoneResp.text)
+        log.debug('Unable to get zone metadata')
+        log.debug('\tReturn code: %d' % zoneResp.status_code)
```
jesusbv: shouldn't we use `format` ?

```python
log.debug('\tReturn code: {}'.format(zoneResp.status_code))
```

PR #297, opened 2026-01-09
More coverage for code formating

rjschwei:
Also apply code formatting via black to our test code and the urlresolver.


schaefi (approved):
Thanks, looks good

PR #322, opened 2026-05-21
Fix cache instance data

jesusbv:
## Description

do not clean instance data cache file

when a registration attempt to an update infrastructure server fails,
the clean_cache method is invoked, cleaning the whole cache directory
causing the subsequent registration attempt by a sibling server to
fail as the file is not present

This Fixes [bsc#1265930](https://bugzilla.suse.com/show_bug.cgi?id=1265930)

## How to test

on a regular instance, i.e. `suse-sles-15-sp6-byos-v20260519-hvm-ssd-x86_64` run the following commands:

```bash 
sudo registercloudguest -r XXX               # version 10.5.2   
sudo zypper patch && zypper update
sudo zypper ref && zypper up
sudo registercloudguest -r XXX --force-new   # version 11.0.2
```

when registration fails on the first attempt it produces the following output
```bash
2026-05-20 16:43:44,700: ERROR: Registration with ('52.28.214.37', '2a05:d014:9a2:8402:fa7d:8cad:1b49:aa6a') failed. Trying ('54.93.130.182', '2a05:d014:9a2:8400:5dbc:7482:ef44:826f')
2026-05-20 16:43:45,130: ERROR: Registration with ('54.93.130.182', '2a05:d014:9a2:8400:5dbc:7482:ef44:826f') failed. Trying ('54.93.131.24', '2a05:d014:9a2:8401:29e0:3e5f:84a:1f30')
2026-05-20 16:43:45,405: ERROR: Baseproduct registration failed
2026-05-20 16:43:45,406: ERROR: Registering system to registration proxy https://smt-ec2.susecloud.net
SUSEConnect error: open /var/cache/cloudregister/dbf21ceb-cd55-4627-a9c2-9132aaac000b: no such file or directory
```

## Change Type

- [X] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)


cloudregister/registerutils.py
```diff
@@ -136,9 +136,10 @@ def clean_all_standard():
...
 # ----------------------------------------------------------------------------
-def clean_cache():
+def clean_cache(except_files=[]):
     if os.path.isdir(REGISTRATION_DATA_DIR):
-        shutil.rmtree(REGISTRATION_DATA_DIR)
+        registration_files = glob.glob(os.sep.join([REGISTRATION_DATA_DIR, '*']))
+        [os.unlink(reg_f) for reg_f in registration_files if reg_f not in except_files]
         # Python 3.4 compatibility does not have "exist_ok"
         try:
             Path(REGISTRATION_DATA_DIR).mkdir(parents=True)
```
jesusbv: this would raise an exception, no matter the Python version, as we are not deleting the entire directory tree, unless I'm mistaken

jesusbv:
we could keep `clean_cache` method as it is, and re-generate the instance data file in each sibling attempt,
IMO, that is suboptimal as we know the file is going to be used again

cloudregister/registerutils.py
```diff
@@ -136,9 +135,16 @@ def clean_all_standard():
 
 
 # ----------------------------------------------------------------------------
-def clean_cache():
+def clean_cache(except_files=[]):
```
jesusbv: maybe a rename s/except_files/keep_files is better ?

cloudregister/registercloudguest.py
```diff
@@ -389,7 +389,7 @@ def register_base_product(
                     utils.clean_hosts_file(
                         registration_target.get_domain_name()
                     )
-                    utils.clean_cache()
+                    utils.clean_cache([instance_data_filepath])
```
schaefi: can we make this more readable as 

```python3
utils.clean_cache(except_files=[instance_data_filepath])
```

I'm suggesting this because then it's clear that instance_data_filepath is the exception to the cleanup, meaning it will stay

cloudregister/registerutils.py
schaefi: I have no strong opinion as long as the call makes it clear :)

cloudregister/registerutils.py
```diff
@@ -136,9 +135,16 @@ def clean_all_standard():
...
     if os.path.isdir(REGISTRATION_DATA_DIR):
-        shutil.rmtree(REGISTRATION_DATA_DIR)
+        registration_files = glob.glob(
+            os.sep.join([REGISTRATION_DATA_DIR, '*'])
+        )
```
schaefi: Do we know that the directory is a flat tree ? I mean no sub-directories with files in it ? I'm asking because glob will not travel recursively

cloudregister/registerutils.py
schaefi: correct, this raises a FileExistsError. So with the changed code we have to check if the target dir exists prior creating it

schaefi: or a try block ignoring FileExistsError. This might be better because testing for exists and then create it introduces a small but possible race condition.

schaefi: oh but looking at the code, this is what we do already

schaefi: So I think we are good here.

cloudregister/registerutils.py
jesusbv: good point, I will modify it to handle sub-directories as well

jesusbv: done in [ff7b559](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/322/commits/ff7b5595758ffd612e536e3361e1f360fb49ccff)

test/unit/registercloudguest_test.py
```diff
@@ -3048,7 +3048,7 @@ def test_setup_ltss_registration_registration_failed_error_message(
 
     @patch('cloudregister.registerutils.register_product')
     @patch('cloudregister.registerutils.add_hosts_entry')
-    @patch('cloudregister.registercloudguest.cleanup')
+    @patch('cloudregister.registerutils.clean_cache')
```
jesusbv, line 3051: had to change this leftovers for the tests to pass

cloudregister/registercloudguest.py
jesusbv: done in [903e05d](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/322/commits/903e05d721aba03807ea6ba8199afdad682f0b82)

cloudregister/registerutils.py
```diff
@@ -2895,3 +2901,37 @@ def _matches_susecloud(elements):
         if 'registry-' in element and 'susecloud.net' in element:
             return element
     return ''
+
+
+# ----------------------------------------------------------------------------
+def _move_files_to_tmp(preserve_files=[]):
+    """Move file in preserve_files list to tmp/PRESERVE_FILES."""
+    parent_tmp_dir = os.path.join(os.sep, 'tmp', PRESERVE_FILES)
```
schaefi, line 2909: I would do this differently. If possible from a python version point of view I suggest to use 

```python3
from tempfile import TemporaryDirectory

PRESERVE_FILES = TemporaryDirectory(dir='/var/tmp', prefix='cloudregister_')
```
* I suggest to use `var/tmp` and not `tmp` for preserving data because `tmp` is a tmpfs and we don't know how much data we preserve here. If we know it never will be big you can use `/tmp`
* TemporaryDirectory has the big advantage that it cleans up automatically when the scope is gone. You can then also initialize PRESERVE_FILES with a None value

TemporaryDirectory exists in python 3.4 which is the oldest python we have to support
* https://docs.python.org/3.4/library/tempfile.html

cloudregister/registerutils.py
```diff
@@ -2895,3 +2901,37 @@ def _matches_susecloud(elements):
...
+    and remove tmp/PRESERVE_FILES/target_dir."""
+    parent_tmp_dir = os.path.join(
+        os.sep, 'tmp', PRESERVE_FILES, target_dir.removeprefix(os.sep)
+    )
+    # move the preserved data back
+    shutil.move(parent_tmp_dir, target_dir)
+    tmp_dir = os.path.join(
+        os.sep, 'tmp', PRESERVE_FILES, Path(target_dir).parts[1]
+    )
+    shutil.rmtree(tmp_dir)
```
schaefi, line 2937: Given you use `TemporaryDirectory` all of this should be in a block saying 

```python3
if PRESERVE_FILES and os.path.isdir(PRESERVE_FILES):
    ...
```

Because then you know in the scope of this runtime there is data to preserve

rjschwei (changes requested):
It should be easy to recreate the data that is being attempted to be preserved, why do we need this complicated arrangement of moving stuff around?

https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/cloudregister/registercloudguest.py#L753 to line 762 has the necessary code. Stick it into utils as `write_instance_data()` and call the new function from here and from line 753. Sounds a lot simpler to me than shuffling data around.

schaefi:
> It should be easy to recreate the data that is being attempted to be preserved

I was looking at the code from the perspective of cache data that must not be lost. That the information can be easily re-created was not subject to the issue so far. So you are saying the background of all this is just the instance data file path presented as UUID filename in /var/cache/cloudregister/ ?

If so I agree this file could be regenerated any time which is a simpler solution than preserving it

However, I have another question with regards to cached data. Is there any condition when cached data must be preserved for some reason in order to run a re-registration ? Actually by design of a cache information it shouldn't.
Thanks

@jesusbv ... this might change everything ;)

jesusbv:
> > It should be easy to recreate the data that is being attempted to be preserved
> 
> I was looking at the code from the perspective of cache data that must not be lost. That the information can be easily re-created was not subject to the issue so far. So you are saying the background of all this is just the instance data file path presented as UUID filename in /var/cache/cloudregister/ ?
> 
> If so I agree this file could be regenerated any time which is a simpler solution than preserving it
> 
> However, I have another question with regards to cached data. Is there any condition when cached data must be preserved for some reason in order to run a re-registration ? Actually by design of a cache information it shouldn't. Thanks
> 
> @jesusbv ... this might change everything ;)

I took this path because I thought it was preferred to used information that was already generated that query the imds of the CSP, like "why are we querying the IMDS if we have that information in the cache already ?"

if we are happy with querying for new instance data, fine by me and the code becomes simpler

I will push that change, unless there is any objection on querying for instance data per update server sibling attempt

jesusbv:
closing in favor of #323

PR #323, opened 2026-06-02
write instance data cache file

jesusbv:
## Description

write instance data cache file

when a registration attempt to an update infrastructure server fails,
the clean_cache method is invoked, cleaning the whole cache directory
causing the subsequent registration attempt by a sibling server to
fail as the file is not present

write the instance data if it's not in the cache directory

This Fixes [bsc#1265930](https://bugzilla.suse.com/show_bug.cgi?id=1265930)

## How to test

on a regular instance, i.e. `suse-sles-15-sp6-byos-v20260519-hvm-ssd-x86_64` run the following commands:

```bash 
sudo registercloudguest -r XXX               # version 10.5.2   
sudo zypper patch && zypper update
sudo zypper ref && zypper up
sudo registercloudguest -r XXX --force-new   # version 11.0.2
```

when registration fails on the first attempt it produces the following output
```bash
2026-05-20 16:43:44,700: ERROR: Registration with ('52.28.214.37', '2a05:d014:9a2:8402:fa7d:8cad:1b49:aa6a') failed. Trying ('54.93.130.182', '2a05:d014:9a2:8400:5dbc:7482:ef44:826f')
2026-05-20 16:43:45,130: ERROR: Registration with ('54.93.130.182', '2a05:d014:9a2:8400:5dbc:7482:ef44:826f') failed. Trying ('54.93.131.24', '2a05:d014:9a2:8401:29e0:3e5f:84a:1f30')
2026-05-20 16:43:45,405: ERROR: Baseproduct registration failed
2026-05-20 16:43:45,406: ERROR: Registering system to registration proxy https://smt-ec2.susecloud.net
SUSEConnect error: open /var/cache/cloudregister/dbf21ceb-cd55-4627-a9c2-9132aaac000b: no such file or directory
```

## Change Type

- [X] **Bug Fix** (a non-breaking change which fixes an issue)
- [ ] **New Feature** (a non-breaking change which adds new functionality)
- [ ] **Documentation Update** (a change which only updates documentation)

### DIFF
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -34,7 +34,7 @@ import requests
 import sys
 import time
 import urllib.parse
-import uuid
+
 
 from cloudregister.lock import Lock
 from cloudregister.logger import Logger
@@ -390,6 +390,9 @@ def register_base_product(
                         registration_target.get_domain_name()
                     )
                     utils.clean_cache()
+                    # instance data is passed to the registration utility,
+                    # make sure it persists when attempting a retry
+                    instance_data_filepath = utils.write_instance_data()
                     registration_target = smt_srv
                     break
         else:
@@ -750,17 +753,7 @@ def main(args):
         find_alive_registration_target(registration_smt, region_smt_servers)
     )
 
-    # Check if we need to send along any instance data
-    instance_data_filepath = ''
-    instance_data = utils.get_instance_data(cfg)
-    if instance_data:
-        instance_data_filepath = os.path.join(
-            utils.get_state_dir(), str(uuid.uuid4())
-        )
-        inst_data_out = open(instance_data_filepath, 'w')
-        inst_data_out.write(instance_data)
-        inst_data_out.close()
-
+    instance_data_filepath = utils.write_instance_data(cfg)
     if registration_target_found:
         # 1. check/setup the container registry for this target
         if not setup_registry(registration_smt):
--- a/cloudregister/registerutils.py
+++ b/cloudregister/registerutils.py
@@ -32,6 +32,7 @@ import sys
 import time
 import toml
 import yaml
+import uuid
 
 from collections import namedtuple
 from lxml import etree
@@ -2863,6 +2864,24 @@ def set_registry_fqdn_suma(private_registry_fqdn):
 
     # the registry value inside the file has the update server registry FQDN
     return True
+
+
+# ----------------------------------------------------------------------------
+def write_instance_data(cfg=None):
+    if not cfg:
+        cfg = get_config()
+
+    # Check if we need to send along any instance data
+    instance_data_filepath = ''
+    instance_data = get_instance_data(cfg)
+    if instance_data:
+        instance_data_filepath = os.path.join(
+            get_state_dir(), str(uuid.uuid4())
+        )
+        with open(instance_data_filepath, 'w') as inst_data_out:
+            inst_data_out.write(instance_data)
+
+    return instance_data_filepath
 
 
 # ----------------------------------------------------------------------------
--- a/package/fix-for-sles12-disable-ipv6.patch
+++ b/package/fix-for-sles12-disable-ipv6.patch
@@ -1,11 +1,12 @@
 diff --git a/cloudregister/smt.py b/cloudregister/smt.py
-index 62c461a..19c7ebd 100644
+index 55f189f..59cb0e4 100644
 --- a/cloudregister/smt.py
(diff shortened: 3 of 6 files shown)


schaefi:
@jesusbv @rjschwei Just to put me right.

The current code change applies to the loop of `register_base_product()` when there are still entries in `failed_smts`. For each failed attempt a cleanup is performed but we have to keep the instance metadata around 
for the next `utils.register_product(...)`  call, which is now achieved by creating this file after cleanup.

Am I right with this logic ? If yes this looks good to me. It would be nice if the commit message or some inline code comment points to a reason why the instance metadata gets created after cleanup.

Other than that I believe we are good once the patches for packaging (SLE12 variant) applies again

Thanks

jesusbv:
> @jesusbv @rjschwei Just to put me right.
> 
> The current code change applies to the loop of `register_base_product()` when there are still entries in `failed_smts`. For each failed attempt a cleanup is performed but we have to keep the instance metadata around for the next `utils.register_product(...)` call, which is now achieved by creating this file after cleanup.
> 
> Am I right with this logic ? 

correct

> If yes this looks good to me. It would be nice if the commit message or some inline code comment points to a reason why the instance metadata gets created after cleanup.

it is not clear from the current commit message ?

> Other than that I believe we are good once the patches for packaging (SLE12 variant) applies again

I will fix that

schaefi:
> it is not clear from the current commit message ?

@jesusbv It is, which lead to the comment I wrote. However, by just reading the code without this information here it is imho not clear unless the full retry logic is understood. I think it would be nice to have a one liner comment why we clean cache data and then create parts of it directly after.

cloudregister/registercloudguest.py
```diff
@@ -390,6 +390,7 @@ def register_base_product(
                         registration_target.get_domain_name()
                     )
                     utils.clean_cache()
+                    instance_data_filepath = utils.write_instance_data()
```
schaefi, line 395: some comment like this would be helpful imho

```python3
# instance data is passed to the registration utility, make sure it persists when attempting a retry
```

cloudregister/registercloudguest.py
```diff
@@ -750,17 +751,7 @@ def main(args):
...
-    instance_data = utils.get_instance_data(cfg)
-    if instance_data:
-        instance_data_filepath = os.path.join(
-            utils.get_state_dir(), str(uuid.uuid4())
-        )
-        inst_data_out = open(instance_data_filepath, 'w')
-        inst_data_out.write(instance_data)
-        inst_data_out.close()
-
+    instance_data_filepath = utils.write_instance_data(cfg)
```
schaefi, line 756: As we are on it, the comment on the first line saying that we check if we need to send any instance data is pretty misleading. We always sent instance when there is any.

rjschwei: Not quite, we send instance data when it is configured. There is a path where an instance may have instance metadata but it does not get sent. Keep in mind the code is not only for our operations.

cloudregister/registercloudguest.py
rjschwei: Yes a comment would be good.

rjschwei (approved):
Please also fix the typos in the commit message before merge

jesusbv:
> Please also fix the typos in the commit message before merge

I can squash the commits into one and fix the typos, but I do not see any typo ?
I re structured/wrote the commit message

cloudregister/registercloudguest.py
schaefi: ok but the above code always creates instance metadata when `utils.get_instance_data` returns something. The comment on top of that code path is imho misleading. If there is a check that we send that data only on a condition, shouldn't that comment then appear at this place where this conditional check really happens ?

schaefi: Looking closer there are more inconsistencies. So the actual check if we need instance data is done in `get_instance_data` itself. So the method creates instance data if there is a dataProvider configured. As such can we please change the comment to

```
Create instance data if a dataProvider is configured
```

The actual sending of that data happens later and there is actually no check for sending, just a check for
creating.

In addition the initialization of 

```
instance_data_filepath = ''
```

is wrong because in `get_instance_data` it`s initialized as 

```
instance_data = b''
```

so why is it initialized in the first place, when `get_instance_data` always returns with a value ?

Sorry this is a bit off topic and I can create an extra PR for this one

schaefi: ok there is a `instance_data.decode()` 

forget the last comment

PR #324, opened 2026-06-08
create the state directory when populating the cache

jesusbv:
when populating the cache, the method writes to the state directory and fails because the state directory does not exist

add a method to create the state directory

This Fixes bsc#1267739


schaefi (approved):
Looks good 👍

PR #325, opened 2026-06-08
use the method get_state_dir to get the registration data directory

jesusbv:


schaefi:
@jesusbv Code looks good but the title and/or the description looks incomplete

PR #327, opened 2026-06-09
update changelog

jesusbv:


rjschwei (approved):

schaefi:
I'm sorry this is not how it works. A version bump and tag must not be done from a pull request. We agreed to create a discrete PR for only the update of the changelog. The respective version bump, tag and push is then an action done from master followed by a submission of the package. Otherwise you would create merge commits for version change which is confusing and inconsistent to how it was done so far.

Long story short please drop the commit that changes the version here. Thanks

schaefi (changes requested):
no version bump in a merge request. Thanks

schaefi (approved):
perfect 👍 Thanks

PR #330, opened 2026-06-26
Add timeout and DNS validation to dataProvider execution

pablo-herranz:
## Problem
`registercloudguest` hangs indefinitely when the domain specified in the `dataProvider` configuration does not resolve (NXDOMAIN). This causes:

1. **Indefinite hang**: The process blocks waiting for DNS resolution or metadata service response
2. **Zombie processes**: SSH timeouts kill the connection but leave `registercloudguest` and child processes running
3. **Registration failures**: Subsequent registration attempts fail with "already running" errors
4. **No clear error message**: The hang provides no feedback about the root cause

This was observed on GCE BYOS instances where `/etc/regionserverclnt.cfg` contains:
```
dataProvider = /usr/bin/gcemetadata --query instance --identity http://smt-gce.susecloud.net
```
but the domain `smt-gce.susecloud.net` does not exist in DNS.

## Solution
This PR implements a two-layer defense mechanism to prevent indefinite hangs:

### 1. DNS Validation (Fail Fast)
Before executing the `dataProvider` command, validate that any domain in the URL resolves:
- Extract domain from URLs in the `dataProvider` configuration
- Attempt DNS resolution with `socket.gethostbyname()`
- If resolution fails (NXDOMAIN), skip execution and return early with empty instance data
- Log clear warning message for debugging

### 2. Execution Timeout (Safety Net)
If DNS validation passes but the command still hangs (slow metadata service, network issues):
- Execute `dataProvider` command with a 30-second timeout
- Kill the process if timeout is reached
- Return `-2` exit code to indicate timeout occurred
- Log timeout event for debugging

### Modified `exec_subprocess()` function
- Added optional `timeout` parameter
- Added `subprocess.TimeoutExpired` exception handling
- Process is killed if timeout is reached
- Returns `-2` exit code on timeout

### Modified `get_instance_data()` function
- Added DNS validation before executing `dataProvider` command
- Early return with empty instance data if domain does not resolve
- Execute `dataProvider` with 30-second timeout
- Handle timeout return code (`-2`)

## Testing
Tested on GCE BYOS instance (SLE 15 SP7) with misconfigured `dataProvider`:

**Before patch:**
- Process hangs indefinitely waiting for DNS/metadata
- SSH timeout creates zombie processes
- No clear error message
- Subsequent retries fail with "already running"

**After patch:**
```
2026-06-26 09:20:24,211: WARNING: Domain "smt-gce.susecloud.net" in dataProvider does not resolve. Skipping instance data collection.
2026-06-26 09:20:24,486: ERROR: Registration with ('35.204.109.102', '2600:1900:4060:b89:0:1:0:8') failed. Trying ('35.204.122.117', '2600:1900:4060:b89:0:2:0:8')
2026-06-26 09:20:24,646: WARNING: Domain "smt-gce.susecloud.net" in dataProvider does not resolve. Skipping instance data collection.
2026-06-26 09:20:24,887: ERROR: Baseproduct registration failed
2026-06-26 09:20:24,087: ERROR: Registering system to registration proxy https://smt-gce.susecloud.net
Error: Registration server returned 'Malformed instance data' (422)
```

**Results:**
✅ No hang - process completes in ~1 second
✅ No zombie processes
✅ Clear warning message identifying the issue
✅ Fast failure instead of indefinite hang
✅ Subsequent retries work immediately (no "already running" errors)

The final `422 Malformed instance data` error is expected and correct - the SMT server rejects registration without valid instance metadata. This is an infrastructure issue (missing DNS entry), not a code issue.

## Related Issues
- https://progress.opensuse.org/issues/202485
- https://bugzilla.suse.com/show_bug.cgi?id=1261908

## Backward Compatibility
- The `timeout` parameter in `exec_subprocess()` is optional and defaults to `None` (no timeout)
- Existing callers continue to work without changes
- DNS validation only applies to `dataProvider` execution
- Empty instance data is a valid scenario already handled by the registration flow

### DIFF
--- a/cloudregister/registerutils.py
+++ b/cloudregister/registerutils.py
@@ -399,12 +399,13 @@ def enable_repository(repo_name):
 
 
 # ----------------------------------------------------------------------------
-def exec_subprocess(cmd, tolog=True):
+def exec_subprocess(cmd, tolog=True, timeout=None):
     """
     Execute the given command as a subprocess (blocking)
     Returns a tuple list:
         (binary_stdout, binary_stderr, exitcode)
         A -1 exitcode indicates an OSError at command invocation
+        A -2 exitcode indicates a timeout occurred
     """
     stdout = b''
     stderr = b''
@@ -415,8 +416,13 @@ def exec_subprocess(cmd, tolog=True):
         proc = subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
         )
-        stdout, stderr = proc.communicate()
+        stdout, stderr = proc.communicate(timeout=timeout)
         rcode = proc.returncode
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        stdout, stderr = proc.communicate()
+        rcode = -2
+        log.warning('EXEC timeout after {} seconds: {}'.format(timeout, cmd))
     except OSError as issue:
         log.debug('EXEC failed with: {}'.format(issue))
     return (stdout, stderr, rcode)
@@ -1454,9 +1460,30 @@ def get_instance_data(config):
                     errMsg = 'Could not find configured dataProvider: %s' % cmd
                     log.error(errMsg)
             if os.access(cmd, os.X_OK):
+                # Validate DNS resolution for URLs in the command
+                url_match = re.search(r'https?://([^\s/:]+)', instance_data_cmd)
+                if url_match:
+                    domain = url_match.group(1)
+                    try:
+                        socket.gethostbyname(domain)
+                        log.debug('DNS validation passed for domain: %s' % domain)
+                    except socket.gaierror:
+                        warn_msg = 'Domain "%s" in dataProvider does not resolve. ' % domain
+                        warn_msg += 'Skipping instance data collection.'
+                        log.warning(warn_msg)
+                        # Skip execution and return early with empty instance data
+                        return '<repoformat>plugin:susecloud</repoformat>\n'
+
+                # Execute with 30 second timeout to prevent indefinite hangs
                 instance_data, errors, returncode = exec_subprocess(
-                    instance_data_cmd.split()
+                    instance_data_cmd.split(),
+                    timeout=30
                 )
+                if returncode == -2:
+                    # Timeout occurred
+                    warn_msg = 'dataProvider command timed out after 30 seconds. '
+                    warn_msg += 'Instance data collection failed.'
+                    log.warning(warn_msg)
                 if errors:
                     errMsg = 'Data collected from stderr for instance '
                     errMsg += 'data collection "%s"' % errors.decode()


cloudregister/registerutils.py
```diff
@@ -1454,9 +1460,32 @@ def get_instance_data(config):
...
+                        socket.gethostbyname(domain)
+                        log.debug('DNS validation passed for domain: %s' % domain)
+                    except socket.gaierror:
+                        warn_msg = 'Domain "%s" in dataProvider does not resolve. ' % domain
+                        warn_msg += 'Skipping instance data collection.'
+                        log.warning(warn_msg)
+                        # Skip execution and return early with empty data
+                        inst_data = ''
+                        repo_format = '<repoformat>plugin:susecloud</repoformat>\n'
+                        return inst_data + repo_format
```
m-dati: IMHO, if you return here and `inst_data` is empty, no need to add that empty string 🙂 
```suggestion
                        return repo_format
```

mloviska: I guess it should be rather populated by `instance_data.decode()`

jesusbv: but at point you know it is empty ?

```python3
    inst_data = instance_data.decode()
    repo_format = '<repoformat>plugin:susecloud</repoformat>\n'
    return inst_data + repo_format
```
could be 
```python3
    return '<repoformat>plugin:susecloud</repoformat>\n'
```

pablo-herranz: Seems legit, let's simplify as @jesusbv suggests, thanks :)

cloudregister/registerutils.py
```diff
@@ -1454,9 +1460,30 @@ def get_instance_data(config):
...
+                        # Skip execution and return early with empty instance data
+                        return '<repoformat>plugin:susecloud</repoformat>\n'
+
+                # Execute with 30 second timeout to prevent indefinite hangs
                 instance_data, errors, returncode = exec_subprocess(
-                    instance_data_cmd.split()
+                    instance_data_cmd.split(),
+                    timeout=30
                 )
+                if returncode == -2:
```
jesusbv, line 1482: not sure we need this block

if the `exec_subprocess` raised a `TimeoutExpired` exception, wouldn't `instance_data` be empty and `errors` may also be populated ? making the error  going into either or both blocks ?

[error block](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/cloudregister/registerutils.py#L1460) 

[no instance block](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/blob/master/cloudregister/registerutils.py#L1464)

cloudregister/registerutils.py
```diff
@@ -1454,9 +1460,30 @@ def get_instance_data(config):
                     errMsg = 'Could not find configured dataProvider: %s' % cmd
                     log.error(errMsg)
             if os.access(cmd, os.X_OK):
+                # Validate DNS resolution for URLs in the command
+                url_match = re.search(r'https?://([^\s/:]+)', instance_data_cmd)
```
ricardobranco777, line 1464: I suggest using `urlparse` from `urllib.parse` to both validate and extract the domain like this:

```
>>> from urllib.parse import urlparse
>>> url = urlparse("https://suse.com")
>>> url.netloc
'suse.com'
```

jesusbv: I would suggest to avoid using `re.search` for the URL,
it should be the `--identity` option value (for all CSPs) so 

```python3
    foo = dataProvider.split()
    if foo.index('--identity') != -1:
        url_index = foo.index('--identity') + 1
        url = foo[url_index]
```
or
```python3
    foo = dataProvider.split('--identity ')
    if len(foo) > 1:
        url = foo[1].split()[0]
```

cloudregister/registerutils.py
jesusbv: the code would log the error of the `TimeoutExpired` exception in the right method, meaning that info is in the log, then `get_instance_data` would log that the metadata is empty or there was an issue, as that method does not care why it failed, it cares that it failed

with that block, log would have something like:
```bash
EXEC timeout after 30 seconds: /usr/bin/gcemetadata --query instance --identity http://smt-gce.susecloud.net
dataProvider command timed out after 30 seconds. Instance data collection failed.
Possible issue accessing the metadata service. Metadata is empty, may result in registration failure.
```
vs removing that block
```bash
EXEC timeout after 30 seconds: /usr/bin/gcemetadata --query instance --identity http://smt-gce.susecloud.net
Possible issue accessing the metadata service. Metadata is empty, may result in registration failure.
```

Thoughts ?

rjschwei:
> This was observed on GCE BYOS instances where /etc/regionserverclnt.cfg contains:
dataProvider = /usr/bin/gcemetadata --query instance --identity http://smt-gce.susecloud.net
but the domain smt-gce.susecloud.net does not exist in DNS.

And it doesn't need to exist because the identity argument is used for generation of the JWT and not used for accessing anything at the time gcemetadata is executed. Could also be `this_domain_does_not_have_to_exists.fobar`

cloudregister/registerutils.py
rjschwei: The `--identity` option does not come into play here. The premise of the initial problem statement is incorrect. `--identity` is an argument to the configured command to get certain information from the metadata server, it is not used to resolve anything at this point. While the idea of timeouts in the proper places is a good thing basing it on an incorrect assumption that the domain presented with the `--identity` option to a command which is completely independent of this process is not the proper approach.

If the token generation process on GCE IMDS should have changed and the domain needs to now be reachable than that would be a fix in the metadata tool and not in cloud-regionsrv-client.

cloudregister/registerutils.py
pablo-herranz: Yup, I'll remove the block so that the output is cleaner, thank you :)

rjschwei (changes requested):
Lets not start with a problem statement the is fundamentally incorrect.

rjschwei:
> If resolution fails (NXDOMAIN), skip execution and return early with empty instance data

This will make registration fail for the wrong reason.

rjschwei:
And to show that the initial analysis is completely incorrect, and therefore the conclusion drawn for this PR are incorrect do the following:

launch a Debian instance on GCE then ssh to the instance and execute the following commands:

```
-> sudo -i
-> apt install git bind9-dnsutils make python3-setuptools
-> git clone https://github.com/SUSE-Enceladus/gcemetadata.git
-> nslookup smt-gce.susecloud.net
-> cd gcemetadata
-> python3 setup.py install
-> /usr/local/bin/gcemetadata --query instance --identity http://smt-gce.susecloud.net
```

And the last command will produce a JWT token as expected where the produced token will have

```
"aud": "http://smt-gce.susecloud.net"
```

i.e. exactly what we expect when using the `--identity` argument with the `gcemetadata` command.

I am going to close this since

a.) we are starting from the wrong origin
b.) we cannot make a decision in cloud regionsrv client as to what arguments to accept for the externally provided commands

As stated earlier, timeouts in the proper places are generally still a good idea but this implementation is not correct and we should start over.

PR #331, opened 2026-07-28
fix _sitelibdir macro

jesusbv:
RPM cannot parse recursive/nested macros using the %{%{macro}_suffix} syntax natively in this format

we could keep _sitelibdir and re define it as

  %global _sitelibdir %{expand:%%{%{pythons}_sitelib}}

but we are importing python-rpm-macros with a BuildRequires, so we can use python_sitelib directly

### DIFF
--- a/package/cloud-regionsrv-client.spec
+++ b/package/cloud-regionsrv-client.spec
@@ -24,7 +24,6 @@
 %define pythons python3
 %endif
 %endif
-%global _sitelibdir %{%{pythons}_sitelib}
 
 %define eflag /run/azuretimer-was-enabled
 %define aflag /run/azuretimer-was-running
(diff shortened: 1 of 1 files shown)


package/cloud-regionsrv-client.spec
```diff
@@ -24,7 +24,6 @@
 %define pythons python3
 %endif
 %endif
-%global _sitelibdir %{%{pythons}_sitelib}
```
jesusbv, line 27: this was [working](https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/209/changes) but it seems it is no longer working

`amazonec2.py` and `cloudguest_lic_watcher.py` are missing in the directory for `3.11` for example

jesusbv:
the issue was not to call the right `zypper up` command, as `cloud-regionsrv-client` can not have a dependency on the plugins

PR #332, opened 2026-07-28
Make the client package require a plugin package

schaefi:
Apply the same concept as used with the config package and make the client to require a plugin package. For third party vendors for which no plugin package exists a generic package is provided to allow seamless installation of the client. Requiring a plugin package solves the potential inconsistency that exists when only the client package gets updated without having a plugin requirement that must be updated as well.

### DIFF
--- a/package/cloud-regionsrv-client.spec
+++ b/package/cloud-regionsrv-client.spec
@@ -48,6 +48,7 @@ Requires:       sed
 Requires:       SUSEConnect > 0.3.31
 Requires:       ca-certificates
 Requires:       cloud-regionsrv-client-config
+Requires:       cloud-regionsrv-client-plugin = %{version}
 %ifarch %ix86 x86_64
 Requires:       dmidecode
 %endif
@@ -112,8 +113,6 @@ Obtain cloud SMT server information from the region server configured in
 /etc/regionserverclnt.cfg
 
 %package generic-config
-Version:        1.0.0
-Release:        0
 Summary:        Cloud Environment Guest Registration Configuration
 Group:          Productivity/Networking/Web/Servers
 Provides:       cloud-regionsrv-client-config
@@ -124,46 +123,55 @@ Conflicts:      otherproviders(cloud-regionsrv-client-config)
 Generic configuration for the registration client. The configuration needs
 to be adapted for the specific cloud framework after installation.
 
+%package plugin
+Summary:        Cloud Environment Guest Registration Plugin, Generic
+Group:          Productivity/Networking/Web/Servers
+Provides:       cloud-regionsrv-client-plugin = %{version}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}
+
+%description plugin
+Generic Guest registration plugin providing no framework
+specific implementation
+
 %package plugin-gce
-Version:        2.0.0
-Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for GCE
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}
 
 %description plugin-gce
 Guest registration plugin for images intended for Google Compute Engine
 providing information to get the appropriate data form the region server.
 
 %package plugin-ec2
-Version:        2.0.0
-Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Amazon EC2
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}
 
 %description plugin-ec2
 Guest registration plugin for images intended for Amazon EC2 providing
 information to get the appropriate data form the region server.
 
 %package plugin-azure
-Version:        3.0.0
-Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Microsoft Azure
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
 Requires:       %{pythons}-dnspython
+Requires:       cloud-regionsrv-client = %{version}
 
 %description plugin-azure
 Guest registration plugin for images intended for Microsoft Azure providing
 information to get the appropriate data form the region server.
 
 %package license-watcher
-Version:	1.0.0
-Release:	0
-Summary:	Enable/Disable Guest Registration for a running instance
-Group:		Productivity/Networking/Web/Servers
-Requires:	cloud-regionsrv-client >= 10.5.4
+Summary:        Enable/Disable Guest Registration for a running instance
+Group:          Productivity/Networking/Web/Servers
+Requires:       cloud-regionsrv-client = %{version}
 Requires:       python-instance-billing-flavor-check >= 1.0.0
 Provides:       cloud-regionsrv-client-addon-azure = 1.0.6
 Obsoletes:      cloud-regionsrv-client-addon-azure <= 1.0.5
@@ -240,9 +248,9 @@ sed -i s/python3/python3.11/ %{buildroot}/usr/lib/zypp/plugins/urlresolver/susec
 # new service.
 if [ $1 -ge 1 ]; then \
     if [ x$(systemctl is-enabled regionsrv-enabler-azure.timer 2>/dev/null ||:) = "xenabled" ]; then
-		touch %eflag
-	fi
-	systemctl is-active regionsrv-enabler-azure.timer &>/dev/null && touch %aflag ||:
+        touch %eflag
+    fi
+    systemctl is-active regionsrv-enabler-azure.timer &>/dev/null && touch %aflag ||:
 fi
 
 %preun
@@ -363,6 +371,9 @@ fi
 %dir /usr/lib/regionService/certs
 %config %{_sysconfdir}/regionserverclnt.cfg
 %config %{_sysconfdir}/logrotate.d/cloudregionsrvclient
+
+%files plugin
+%defattr(-,root,root,-)
 
 %files plugin-gce
 %defattr(-,root,root,-)


schaefi:
This is a start, let me know if it matches with the conversation we had on slack

package/cloud-regionsrv-client.spec
```diff
@@ -124,12 +125,27 @@ Conflicts:      otherproviders(cloud-regionsrv-client-config)
 Generic configuration for the registration client. The configuration needs
 to be adapted for the specific cloud framework after installation.
 
+%package plugin
+Version:        1.0.0
+Release:        0
+Summary:        Cloud Environment Guest Registration Plugin, Generic
+Group:          Productivity/Networking/Web/Servers
+Provides:       cloud-regionsrv-client-plugin
+Conflicts:      otherproviders(cloud-regionsrv-client)
```
jesusbv: shouldn't be `cloud-regionsrv-client-plugin` here ?

package/cloud-regionsrv-client.spec
```diff
@@ -140,7 +156,9 @@ Version:        2.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Amazon EC2
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin
+Conflicts:      otherproviders(cloud-regionsrv-client)
```
jesusbv: ditto

package/cloud-regionsrv-client.spec
schaefi: yes my bad

package/cloud-regionsrv-client.spec
```diff
@@ -124,12 +125,27 @@ Conflicts:      otherproviders(cloud-regionsrv-client-config)
...
+specific implementation
+
 %package plugin-gce
 Version:        2.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for GCE
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin
+Conflicts:      otherproviders(cloud-regionsrv-client)
```
jesusbv: ditto

package/cloud-regionsrv-client.spec
```diff
@@ -151,19 +169,21 @@ Version:        3.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Microsoft Azure
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin
+Conflicts:      otherproviders(cloud-regionsrv-client)
```
jesusbv: ditto

package/cloud-regionsrv-client.spec
jesusbv: copypasta has these things

package/cloud-regionsrv-client.spec
```diff
@@ -48,6 +48,7 @@ Requires:       sed
 Requires:       SUSEConnect > 0.3.31
 Requires:       ca-certificates
 Requires:       cloud-regionsrv-client-config
+Requires:       cloud-regionsrv-client-plugin = %{version}-%{release}
```
rjschwei: The use `%{version}-%{release}` will not work here as these macros will be expanded in the context of the main package, i.e. at the moment would resolve to 11.0.3 and none of the plugins meet that requirement as they have their own versioning scheme. I would rather not change that, meaning IMHO we should keep the version separation between the plugins and the main package. If we do want a version here then the best we could do is a `>= 1.0.0` where 1.0.0 is the version for the generic plugin and all the other plugins have higher versions already.

Also we should only use the versions everywhere and not the release since in general we do not set the release and let the build service handle that.

package/cloud-regionsrv-client.spec
```diff
@@ -124,12 +125,27 @@ Conflicts:      otherproviders(cloud-regionsrv-client-config)
...
 to be adapted for the specific cloud framework after installation.
 
+%package plugin
+Version:        1.0.0
+Release:        0
+Summary:        Cloud Environment Guest Registration Plugin, Generic
+Group:          Productivity/Networking/Web/Servers
+Provides:       cloud-regionsrv-client-plugin = %{version}-%{release}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}-%{release}
```
rjschwei: Will not work, version will be interpreted by RPM in the context of this package, currently this states `Requires:       cloud-regionsrv-client = 1.0.0` and that will fail.

The plugin interface has not changed since the inception of cloud-regionsrv-client and I don't expect many changes to that interface as such this Requires can have a hard coded version on the main package. Since we switched the Python interpreter in 15 SP4 and later the hard coded version should be the version when we switched the interpreter, i.e. `Requires:       cloud-regionsrv-client >= 11.0.0`.

However, this creates a circular dependency, i.e. the `cloud-regionsrv-client` package depends on the `cloud-regionsrv-client-plugin` package and vice-versa. That's not good. There is really no harm in just installing the plugin package. In the update situation without the circle one can still run into the issue where a `zypper up cloud-regionsrv-client-plug` would install a newer package that is built with a different interpreter and then end up with the same problem that triggered these changes. Anyway, we can probably get away without the circular dependency. Or at least double check with the solver team to better understand the harm of having such a circular dependency.

And release should not be used.

package/cloud-regionsrv-client.spec
```diff
@@ -124,12 +125,27 @@ Conflicts:      otherproviders(cloud-regionsrv-client-config)
...
+
 %package plugin-gce
 Version:        2.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for GCE
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}-%{release}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}-%{release}
```
rjschwei: ditto

package/cloud-regionsrv-client.spec
```diff
@@ -140,7 +156,9 @@ Version:        2.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Amazon EC2
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}-%{release}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
+Requires:       cloud-regionsrv-client = %{version}-%{release}
```
rjschwei: ditto

package/cloud-regionsrv-client.spec
```diff
@@ -151,19 +169,21 @@ Version:        3.0.0
 Release:        0
 Summary:        Cloud Environment Guest Registration Plugin for Microsoft Azure
 Group:          Productivity/Networking/Web/Servers
-Requires:       cloud-regionsrv-client >= 10.5.4
+Provides:       cloud-regionsrv-client-plugin = %{version}-%{release}
+Conflicts:      otherproviders(cloud-regionsrv-client-plugin)
 Requires:       %{pythons}-dnspython
+Requires:       cloud-regionsrv-client = %{version}-%{release}
```
rjschwei: ditto

package/cloud-regionsrv-client.spec
```diff
@@ -151,19 +169,21 @@ Version:        3.0.0
...
-Version:	1.0.0
-Release:	0
-Summary:	Enable/Disable Guest Registration for a running instance
-Group:		Productivity/Networking/Web/Servers
-Requires:	cloud-regionsrv-client >= 10.5.4
+Version:        1.0.0
+Release:        0
+Summary:        Enable/Disable Guest Registration for a running instance
+Group:          Productivity/Networking/Web/Servers
+Requires:       cloud-regionsrv-client = %{version}-%{release}
```
rjschwei: ditto

rjschwei (changes requested):

package/cloud-regionsrv-client.spec
schaefi: > Will not work, version will be interpreted by RPM in the context of this package

@rjschwei Hmm, yes too short thinking regarding the macro expansion

So I had this question for a long time. Do we have a specific reason why the sub-packages of the cloud-regionsrv-client package comes with different and sub-package specific version information ?

As we plan to couple plugin with client it would be imho more consistent to maintain only one version.
Since the version of the client is 11.x.x it would not cause an issue to drop `Version` and `Release:` tags from the sub-packages.

> And release should not be used.

according to the packaging policy a versioned provides tag should be set as "%{version}-%{release}" to make sure the provides tag stays versioned even if the main version of the package did not change. Something like this would then apply if you e.g move python interpreters causing python path changes.
We use this concept in the kiwi packages too

But all of this would only work reliable if there is only one version and release information

schaefi: I added a commit that deals with one version only. I also tested some (not all) update scenarios

schaefi: @rjschwei @jesusbv please find package builds for testing here

This results in packages of the form 

```
[   55s] RPMS/noarch/cloud-regionsrv-client-plugin-ec2-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-generic-config-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-plugin-azure-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-plugin-gce-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-plugin-11.0.3-150700.138.1.noarch.rpm
[   55s] RPMS/noarch/cloud-regionsrv-client-license-watcher-11.0.3-150700.138.1.noarch.rpm
[   55s] SRPMS/cloud-regionsrv-client-11.0.3-150700.138.1.src.rpm
```

In an image description that does not explicitly request a plugin package you have a choice now

```
<error>unresolvable: have choice for cloud-regionsrv-client-plugin = 11.0.3 needed by cloud-regionsrv-client: cloud-regionsrv-client-plugin cloud-regionsrv-client-plugin-azure cloud-regionsrv-client-plugin-ec2 cloud-regionsrv-client-plugin-gce</error>
```

In my client builds for a testing infrastructure not connected to a specific cloud I added

```
<package name="cloud-regionsrv-client-plugin"/>
```

which is the newly introduced "generic" plugin package.

All standard image builds of course will just continue to work as they add a certain plugin package in the package list like `cloud-regionsrv-client-plugin-azure`

In a scenario of an already installed system with an older client package, `zypper up cloud-regionsrv-client` is now automatically pulling the client and its versioned plugin package

In a scenario when we update the `cloud-regionsrv-client` package and increase the version by this update which is always the case, all sub-packages will also have this version. This would also update e.g. a plugin package even if the code for this plugin hasn't changed. I don't see an issue with that.

Please play around with the packages built

Thanks

package/cloud-regionsrv-client.spec
schaefi: done

package/cloud-regionsrv-client.spec
schaefi: from my perspective I have done the requested changes. Did anybody had a chance to give the example packages I built a test. I'm using them in my testing environment but this surely does not cover all install/upgrade procedures

Thanks

jesusbv: I haven't done a proper integration test with the newest changes, will do that tomorrow

rjschwei: Sorry I lost track of this. I personally am not a big fan of having a single version to cover all sub-packages, which is why the sub-packages were setup with their own version numbers to begin with. Having only 1 version for everything means that if I want to know if for example the plugin for gce has changed in version 11.2.3 (random version picked) I have to fish through the changelog. And that makes the assumption that the committer did not forget to mention that the plugin changed. When the plugins or other sub-packages have a separate version it is pretty evident that the plugin changed without having to fish through the changelog.

rjschwei (approved):
I am not a fan of making all the sub-packages have the same version, see my other comment for reasoning. The implementation from a technical perspective LGTM, albeit it violates the schematic versioning we follow as there were no changes to the code but we are bumping the versions. This is a pure packaging change for RPM and should not have a version number effect. That said its a inconsistency that we can live with even if I don't like it.

package/cloud-regionsrv-client.spec
schaefi: Hmm, it's all build from the same spec and with the client requiring a plugin we actually want that main code and plugin code to stay consistent. The drift in version that you described should actually not happen and with enforcing the client to require a plugin by the spec we imho should have only one version because multiple of them seems to be very confusing when everything comes from the same project and code base. I would fully agree with you when client work and plugin work would be separated projects but they are not in this case. Everything comes from cloud-regionsrv-client and pyproject.toml defines one version for all that code including what we later package as plugins. As such the version information for the plugins seemed somewhat artificial to me as they do not appear anywhere in code. Dealing with one version makes this way more consistent, no ?

schaefi:
> I am not a fan of making all the sub-packages have the same version, see my other comment for reasoning. The implementation from a technical perspective LGTM, albeit it violates the schematic versioning we follow as there were no changes to the code but we are bumping the versions. This is a pure packaging change for RPM and should not have a version number effect. That said its a inconsistency that we can live with even if I don't like it.

yes this is correct and I checked if that version bump would cause any harm, e.g. bumping to something lower than before. But the bumping is clean and I only did it for consistency reasons because of the new requirement situation that we didn't had before. I thought this is then ok with regards to the mentioned violation.

Please also take into account, if we stick to the old version concept per plugin we still have to bump the plugin versions such that we gain an effect. Remember it started with cloud-regionsrv-client to be updated and moving
its python version from 3.6 to 3.11. The associated plugin package stayed in the system at python 3.6 and nothing worked anymore. If we change the packaging as I suggested here and stick with per-plugin versions we still have to bump the version such that a package update really installs a new (python 3.11) version of the same plugin.

So to my understanding a plugin version bump is inevitable even though no plugin code has changed.

After all @rjschwei you are the master of this project and sometimes solutions are technically ok but disliked. In the end all this needs your blessing 

Thanks

smarlowucf:
If the code is all coming from the same source repo I would give a 👍🏼 to making the versions all the same. Having the versions different is very confusing and it goes against the RPM default behavior when no version is provided for a subpackage. By default it inherits the version from the parent package.

smarlowucf (approved):

PR #333, opened 2026-08-25
Fix handling of registration returncode

schaefi:
This change is three fold

Prior fixing the actual return code handling I saw the need for two refactoring steps

**Refactor use of global variable**
    
Drop using a global registration_returncode variable whose current value is hard to manage across the process of registering the system.

**Refactor register_base_product returncode handling**
    
Return from the function including a return code.

These two changes were done with help from Claude:claude-opus-4-6

The actual fix for the return code handling in a way that module registrations are not fatal was done in the last commit

**Fix registration status for module registrations**
    
Make sure module registration errors are only logged but do not fail the entire registration. Only base registration errors are handled as fatal.

This Fixes bsc#1275998

Please review carefully since some changes were needed to fix this in a good way

### DIFF
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -81,6 +81,7 @@ def register_modules(
     regcode='',
     registered=[],
     failed=[],
+    failed_undetermined=[],
     returncode=0,
 ):
     """Register modules obeying dependencies
@@ -105,6 +106,7 @@ def register_modules(
                 regcode,
                 registered,
                 failed,
+                failed_undetermined,
                 returncode,
             )
             continue
@@ -158,6 +160,7 @@ def register_modules(
                             triplet, returncode
                         )
                     )
+                    failed_undetermined.append(triplet)
             else:
                 log.info('Registration of {} succeeded'.format(triplet))
                 registered.append(triplet)
@@ -170,6 +173,7 @@ def register_modules(
             regcode,
             registered,
             failed,
+            failed_undetermined,
             returncode,
         )
 
@@ -837,13 +841,15 @@ def main(args):
 
     extensions = get_extensions(registration_target)
     failed_extensions = []
-    modules_returncode = register_modules(
+    failed_extensions_undetermined = []
+    register_modules(
         extensions,
         products,
         registration_target,
         instance_data_filepath,
         args.reg_code,
         failed=failed_extensions,
+        failed_undetermined=failed_extensions_undetermined,
         returncode=0,
     )
     if os.path.exists(instance_data_filepath):
@@ -858,27 +864,34 @@ def main(args):
 
     log.info('Registration succeeded')
 
-    if modules_returncode:
+    if failed_extensions:
         # Module registration issues do not invalidate the registration
         # of the system. The affected modules can be registered at a
         # later point in time
-        log.warning(
-            'Module registration failed(repository), see {0} for '
-            'details'.format(LOG_FILE)
-        )
-
-    if failed_extensions:
         msg = 'There are products that were not registered because they need '
         msg += 'an additional registration code, to register them please run '
-        msg += 'the following command(s): '
+        msg += 'the following command(s):\n'
 
-        activate_prod_cmd = 'SUSEConnect -p {} -r ADDITIONAL REGCODE'
+        activate_prod_cmd = 'SUSEConnect -p {} -r ADDITIONAL REGCODE\n'
         if utils.is_transactional_system():
             activate_prod_cmd = 'transactional-update register -p {} '
-            activate_prod_cmd += '-r ADDITIONAL REGCODE'
+            activate_prod_cmd += '-r ADDITIONAL REGCODE\n'
 
         for failed_extension in failed_extensions:
             msg += activate_prod_cmd.format(failed_extension)
+
+        log.warning(msg)
+
+    if failed_extensions_undetermined:
+        # There are module registrations with an undetermined error code
+        # Report this as a warning to the user.
+        msg = 'Following module registration(s) failed '
+        msg += 'with an undetermined error code, see {0} for details:\n'.format(
+            LOG_FILE
+        )
+
+        for failed_extension in failed_extensions_undetermined:
+            msg += '{}\n'.format(failed_extension)
 
         log.warning(msg)
 
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -146,7 +146,7 @@ def register_modules(
                     'registration code' in error_message.lower()
                     or 'system credentials' in error_message.lower()
                 ):
-                    log.debug(
+                    log.warning(
                         '\tModule registration unsuccesful: {} code {}'.format(
                             error_message, returncode
                         )
@@ -155,7 +155,7 @@ def register_modules(
                 else:
                     # Zypper sets codes that do not indicate registration
                     # failure but the registration is not clean
-                    log.debug(
+                    log.warning(
                         '\tModule registration status for {} undetermined: code {}'.format(
                             triplet, returncode
                         )
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -868,17 +868,19 @@ def main(args):
         )
 
     if failed_extensions:
-        log.warning(
-            'There are products that were not registered because they need '
-            'an additional registration code, to register them please run '
-            'the following command:'
-        )
+        msg = 'There are products that were not registered because they need '
+        msg += 'an additional registration code, to register them please run '
+        msg += 'the following command(s): '
+
         activate_prod_cmd = 'SUSEConnect -p {} -r ADDITIONAL REGCODE'
         if utils.is_transactional_system():
             activate_prod_cmd = 'transactional-update register -p {} '
             activate_prod_cmd += '-r ADDITIONAL REGCODE'
+
         for failed_extension in failed_extensions:
-            log.error(activate_prod_cmd.format(failed_extension))
+            msg += activate_prod_cmd.format(failed_extension)
+
+        log.warning(msg)
 
     # Enable Nvidia repo if repo(s) are configured
     # and destination can be reached
--- a/cloudregister/registercloudguest.py
+++ b/cloudregister/registercloudguest.py
@@ -70,7 +70,6 @@ log = Logger.get_logger()
 requests.packages.urllib3.disable_warnings(
     requests.packages.urllib3.exceptions.InsecureRequestWarning
 )
-registration_returncode = 0
 
 
 # ----------------------------------------------------------------------------
@@ -82,14 +81,19 @@ def register_modules(
     regcode='',
     registered=[],
     failed=[],
+    returncode=0,
 ):
-    """Register modules obeying dependencies"""
-    global registration_returncode
+    """Register modules obeying dependencies
+
+    The given returncode is the registration status of the modules
+    registered so far. It is updated by the registrations this call
+    performs and returned to the caller
+    """
     for extension in extensions:
         # If the extension is recommended it gets installed with the
         # baseproduct registration. No need to run another registration
         if extension.get('recommended'):
-            register_modules(
+            returncode = register_modules(
                 extension.get('extensions'),
                 products,
                 registration_target,
@@ -97,6 +101,7 @@ def register_modules(
                 regcode,
                 registered,
                 failed,
+                returncode,
             )
             continue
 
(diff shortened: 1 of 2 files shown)


schaefi:
On part still missing.... stay tuned

schaefi:
ready now

rjschwei (approved):

gubasso:
The fix is right: a module outcome no longer reaches `sys.exit`. One small thing on the new summary line, not on the fix itself.

## Minor: `returncode` is written then cleared, so the summary can be lost

In `register_modules`, the nonzero code is stored first and cleared afterwards:

```python
if prod_reg.returncode:
    returncode = prod_reg.returncode          # stored
    ...
    if returncode in reg_fail_codes and (
        'registration code' in error_message.lower()
        or 'system credentials' in error_message.lower()
    ):
        failed.append(triplet)
        returncode = 0                        # cleared
    else:
        log.warning('... undetermined')       # left set
```

`returncode` is one slot, not an accumulator, so the `returncode = 0` also wipes what an **earlier** module put there:

```text
module                         result                              returncode
-----------------------------------------------------------------------------
sle-module-web-scripting       ok                                       0
sle-module-confidential-...    ok                                       0
sle-module-public-cloud        rc=67, 'some other problem'             67   <- undetermined, WARNING logged
sle-module-desktop-appl...     ok                                      67
sle-module-development-tools   rc=64, 'invalid system credentials'      0   <- cleared, public-cloud's status gone
sle-module-containers          ok                                       0
                                                              return -> 0
```

`main` sees `0`, so `Module registration failed(repository)` is never logged, even though `sle-module-public-cloud` really was left undetermined.

Swap the two failing modules and the same outcomes produce the line:

```text
module                         result                              returncode
-----------------------------------------------------------------------------
sle-module-web-scripting       ok                                       0
sle-module-confidential-...    ok                                       0
sle-module-development-tools   rc=64, 'invalid system credentials'      0   <- cleared, nothing to clear yet
sle-module-desktop-appl...     ok                                       0
sle-module-public-cloud        rc=67, 'some other problem'             67   <- undetermined, survives
sle-module-containers          ok                                      67
                                                             return -> 67
```

Same system state, different summary, decided only by position.

Same for the two recursive calls: the child's cleared value replaces the parent's.

Registration order is the RMT `/connect/systems/products` array order, which is stable in our runs, so this is reproducible rather than flaky.

Impact is only the summary line. The per-module `undetermined` warning is still always emitted, and the exit code is unaffected.

## Suggestion: don't store what the next branch is going to clear

Set `returncode` only in the branch that means "status unknown":

```python
if prod_reg.returncode:
    # Older versions of SUSEConnect wrote error messages to stdout
    error_message = prod_reg.output
    if not error_message:
        error_message = prod_reg.error
    if prod_reg.returncode in reg_fail_codes and (
        'registration code' in error_message.lower()
        or 'system credentials' in error_message.lower()
    ):
        log.error(
            '\tModule registration unsuccesful: {}'.format(error_message)
        )
        # Module registration is a best effort approach. We do not fail
        # the system registration if we end up missing a specific
        # module, it can always be added later. The module is reported
        # through failed[], it is not a status we have to hand back.
        failed.append(triplet)
    else:
        # Zypper sets codes that do not indicate registration
        # failure but the registration is not clean
        warn_msg = '\tModule registration status for '
        warn_msg += '{} undetermined'.format(triplet)
        log.warning(warn_msg)
        returncode = prod_reg.returncode
```

The same run then keeps the status:

```text
module                         result                              returncode
-----------------------------------------------------------------------------
sle-module-web-scripting       ok                                       0
sle-module-confidential-...    ok                                       0
sle-module-public-cloud        rc=67, 'some other problem'             67   <- undetermined, WARNING logged
sle-module-desktop-appl...     ok                                      67
sle-module-development-tools   rc=64, 'invalid system credentials'     67   <- failed[], returncode untouched
sle-module-containers          ok                                      67
                                                             return -> 67
```

`returncode` only ever moves `0 → nonzero`, so no module and no recursive call can clear another one's status, and the two outcomes stay cleanly separated: modules needing a code are reported via `failed[]`, modules of unknown status via `returncode`.

schaefi:
@gubasso great catch, yes this makes perfect sense. I have added a commit that implements a fix similar to your suggestion. Can you give it a try ? It should operate as follows

```
module                         result                              returncode
-----------------------------------------------------------------------------
sle-module-web-scripting       ok                                       0
sle-module-confidential-...    ok                                       0
sle-module-public-cloud        rc=67, 'some other problem'             67   <- undetermined, WARNING logged plus code
sle-module-desktop-appl...     ok                                      67
sle-module-development-tools   rc=64, 'invalid system credentials'     64   <- failed[], ERROR logger plus code
sle-module-containers          ok                                      64
                                                             return -> 64
```

jesusbv (approved):
LGTM

gubasso (commented):
The fix is correct, and the two refactors carrying it check out.

Four things below, none of them blocking.

## 1. Base-product failure changes its exit code from 1 to the SUSEConnect code

The cleanup moved into `main`, and with it `sys.exit(1)` became `sys.exit(base_returncode)`. A BYOS instance registering with a bad registration code takes the `SERVER_GENERAL_ERROR and 'registration code' and args.reg_code` break:

```text
$ registercloudguest --regcode <invalid-code>; echo "exit=$?"

before   Baseproduct registration failed
         exit=1

after    Baseproduct registration failed
         Registration failed(baseproduct), see /var/log/cloudregister for details
         exit=67
```

Nothing inside the repo branches on it: `guestregister.service` is `Type=oneshot` and only cares about zero vs nonzero, and `cloudguest_lic_watcher` discards the result of `exec_subprocess(['registercloudguest'])`. But it is visible to anything outside that scripts on the exit status, so it may deserve a changelog line.

## 2. The regcode branch now sets the code too

Take the ordinary case this affects: a healthy system whose only nonzero is a module it is not entitled to. 67 is the code the repo's own `test_register_modules` uses for that case.

```text
module                         result                              returncode
-----------------------------------------------------------------------------
sle-module-web-scripting       ok                                       0
sle-module-public-cloud        ok                                       0
SLES-LTSS                      rc=67, 'registration code'              67   <- failed[], user gets the SUSEConnect hint
sle-module-containers          ok                                      67
                                                             return -> 67
```

Same system, same modules, before and after the commit:

```text
before   INFO:    Registration succeeded
         WARNING: There are products that were not registered because they need an
                  additional registration code, ...
         ERROR:   SUSEConnect -p SLES-LTSS/15.7/x86_64 -r ADDITIONAL REGCODE

after    ERROR:   Module registration failed(repository), see /var/log/cloudregister for details
         INFO:    Registration succeeded
         WARNING: There are products that were not registered because they need an
                  additional registration code, ...
         ERROR:   SUSEConnect -p SLES-LTSS/15.7/x86_64 -r ADDITIONAL REGCODE
```

Nothing failed in that run. The system just is not entitled to that module, which is why the lines under it say `Registration succeeded` and hand the user the command. `failed(repository)` reads as a repository problem, and it now fires on a run that is clean and self-service, so anything grepping the log for `ERROR` on a successful registration gains a new hit.

Setting the code only in the `else` branch returns `0` for that run and drops the line, while `failed[]` still reports the module a few lines further down. If you would rather keep both outcomes in `returncode`, the summary line needs to tell them apart. Either works for me; right now that one message carries two meanings.

## 3. Nit: the classification tests the accumulator, not the module

The branch that decides `failed[]` versus `undetermined` reads `returncode`, which is now also the value carried across the whole loop:

```python
returncode = prod_reg.returncode          # accumulator, set from this module
error_message = prod_reg.output
if not error_message:
    error_message = prod_reg.error
if returncode in reg_fail_codes and (     # reads the accumulator, not this module
```

It is correct today, but only because the assignment happens four lines earlier. Nothing states that dependency, so it is easy to break while tidying the branch. Two modules make the difference visible:

```text
module                    result                        today         assignment moved below the check
--------------------------------------------------------------------------------------------------------
sle-module-public-cloud   rc=67, 'some other problem'   undetermined  undetermined
sle-module-desktop-appl   rc=2,  'registration code'    undetermined  failed[]   <- wrong
```

In the right-hand column the second module is judged against the first module's leftover `67`. Its own code is `2`, which is not in `reg_fail_codes`, but `67` is, and its message does contain `registration code`, so both halves of the condition pass and it lands in `failed[]`. The user is then told to run `SUSEConnect -p <triplet> -r ADDITIONAL REGCODE` for a module that failed for a reason a registration code will not fix.

One word fixes it for good:

```python
if prod_reg.returncode in reg_fail_codes and (
```

Same for the two messages that now interpolate `returncode`.

## 4. Pre-existing, but this PR is the moment it becomes worth guarding

`register_base_product` does not terminate if `region_smt_servers` is empty. Each pass appends exactly one entry to `failed_smts`, so the `len(failed_smts) == len(region_smt_servers)` break is reached for any positive length, but never for zero, and the `for smt_srv in region_smt_servers` loop that could switch targets has nothing to iterate over either:

```text
region_smt_servers = [A, B]          region_smt_servers = []
pass  failed_smts  len check         pass  failed_smts  len check
1     [A]          1 != 2, try B     1     [A]          1 != 0, no sibling to try
2     [A, B]       2 == 2, break     2     [A, A]       2 != 0, no sibling to try
                                     3     [A, A, A]    3 != 0, ...
```

The left column also holds when the list has duplicate IPv4s or when the current target is not a member, since the counter grows by one per pass and lands on any positive length exactly.

`main` cannot reach the right column: `get_responding_update_server` runs first, and on an empty list its loop body never executes, so it logs `No response from: []` and exits 1 before `register_base_product` is ever called. This is latent, not live, and it predates the PR.

I raise it because this PR changes the function's contract. It used to exit on its own; now it returns a code and leaves the decision to the caller, which makes it a plain function a future caller could reasonably invoke directly, as the unit test already does. A `if not region_smt_servers: return registration_target, <code>` guard at the top is cheap insurance now that reaching the end of the function is a normal outcome rather than an impossible one. Entirely your call whether it belongs in this PR.

Also unrelated: `registered=[]` and `failed=[]` in the signature are shared mutable defaults. Harmless in `main`, which calls once per process, but `test_register_modules` and `test_register_modules_failed_credentials` both leave `failed` unpassed and append into the same list. Nothing asserts on it today, so nothing fails, just worth knowing before someone adds an assertion on `failed`.

schaefi:
@gubasso 
> Base-product failure changes its exit code from 1 to the SUSEConnect code

Thanks I mentioned this behavior change in the commit message now

>  The regcode branch now sets the code too

This error reporting of the client is a mess. I added a commit that collects the known errors as well as the undetermined errors and prints an error message for each of this conditions when applicable

> Nit: the classification tests the accumulator, not the module

fixed

schaefi:
@gubasso 

> Pre-existing, but this PR is the moment it becomes worth guarding

This might be a valid point but I don't want to confuse this PR with out of scope changes. Could you add this finding as an issue here on the project page or on bugzilla. Thanks

gubasso:
@schaefi 
> This might be a valid point but I don't want to confuse this PR with out of scope changes. Could you add this finding as an issue here on the project page or on bugzilla. Thanks

Sure, I will!

---

Everything else looks correctly addressed, with 2 residuals worth raising...

## Residual 1: the undetermined warning points at a detail the log file may not have

```
  log.debug( # (was log.warning)
      '\tModule registration status for {} undetermined: code {}'.format(triplet, returncode)
  )
```

This is only written to the log file when the debug switch is on... while the warning is always emitted, even though the detail it points to only reaches that log file when the debug switch is on.

```
msg += 'with an undetermined error code, see {0} for details:\n'.format(LOG_FILE)
log.warning(msg)
```

- "log.debug" is written there only when `debug` flag is on
- `log.warning(msg)` will emit the message at warn level and that error code and details will be written there only when debug is on.

A run without the `debug` flag will emit that warning but the code will not be saved, even though the message says otherwise.

## Residual 2: returncode in register_modules is now unused in production

`returncode` in register_modules is now dead weight in production... accumulated, threaded through both recursive calls, returned, and read by nobody except the unit tests.

The docstring  still describes it as the function's meaningful result. It could go, along with the parameter.

(written by a human! :smile:)

schaefi:
> returncode in register_modules is now unused in production

correct and I don't see an issue by that. The method returns an overall returncode. 0 indicates all good, not zero indicates an error. The current code does not evaluate it because module failure are not fatal to the registration and according to your last review we found that for a detailed and better error message the individual failed messages collected by the recursion are more of interest. That doesn't mean the method should not return an overall error code. So the implementation is still good and prepared for eventual changes in the future, e.g. module registrations must succeed or something like this

> This is only written to the log file when the debug switch is on

Yes this should be changed back to a warning

gubasso (approved):
LGTM!

PR #335, opened 2026-09-17
Check access no params

jesusbv:
use activation endpoint for RMT check

instead of migrations endpoint which requires parameters in the body
or it will produce a 422 log in the update server otherwise, use
activations endpoint (no parameters, same check/goal) and no update
server noise logging 422 that were generated on purpose

Changed Unauthorized string check to code

This Fixes [bsc#1279237](https://bugzilla.suse.com/show_bug.cgi?id=1279237)


schaefi (approved):
Looks good to me.

I haven't checked on the real consequences of the new API endpoint though

jesusbv:
> Looks good to me.
> 
> I haven't checked on the real consequences of the new API endpoint though

same behaviour from the client perspective:

- an endpoint to check access
- better info (either success or unauth, before it could return `422` and we treated anything not unauth as success)

better behaviour from the update server perspective:

- no error log because the post has no params
- no change on the server

cloudregister/registerutils.py
```diff
@@ -1884,10 +1884,10 @@ def has_smt_access(update_server_fqdn, user, password):
     """Check if the given update server can be accessed with the provided
     credentials."""
     auth_creds = HTTPBasicAuth(user, password)
-    api = 'https://%s/connect/systems/products/migrations'
-    res = requests.post(api % update_server_fqdn, auth=auth_creds, json={})
+    api = 'https://%s/connect/systems/activations'
+    res = requests.get(api % update_server_fqdn, auth=auth_creds)
```
rjschwei, line 1888: But this was done on purpose to avoid getting anything back. We do not want to request data to be sent over the wire that we just throw away.

If there is any data we have to get back to avoid generating messages people don't like server side then at least we need to use an authenticated endpoint that returns as little data as possible. In SLE 15 activations is a very long list because we registered all the modules.

rjschwei (changes requested):

cloudregister/registerutils.py
schaefi: well so this last comment then conflicts with the initial message of the pull request which seems to have the intention to use an endpoint that does not require any authentication. The thumbs up sent by Jesus, I interpret as an agreement to the comment by Robert. This in turns means people want to use an authenticated endpoint but this PR changes into the opposite.

That's pretty confusing and I don't understand how this should move forward now. Please either of you explain the next steps or close the PR for a new one. Thanks much

